PR #14, opened 2021-11-23
Exclusively use awkward v2

douglasdavis:
- Moves the existing `from_json` function to use Dask's `read_bytes` and Awkward's `from_iter` (which is v2 compatible).
- Use the v2 high level `Array` object from core `awkward`


tests/helpers.py
```diff
@@ -0,0 +1,16 @@
+import pathlib
+
+import awkward as ak
+
+from dask_awkward.io import from_json
+
+this_path = pathlib.Path(__file__).parent.resolve()
```
martindurant, line 7: Since I'm old fashioned, I wonder whether this gets you anything over
```
this_path = os.path.dirname(__fle__)
records_file = os.path.join(this_path, "data", "records.json")
```

src/dask_awkward/io.py
```diff
@@ -55,3 +52,36 @@ def from_parquet(source: Any, **kwargs: Any) -> DaskAwkwardArray:
...
+
+def _chunked_to_json_v2(chunks):
+    return from_iter_v2(json.loads(ch) for ch in chunks.split(b"\n") if ch)
+
+
+def _chunked_to_json_v1(chunks):
+    return from_iter_v1(json.loads(ch) for ch in chunks.split(b"\n") if ch)
+
+
+def _from_json(
```
martindurant, line 65: This is implicitly (line-)delimited JSON, i.e., not a huge single JSON object. Not sure if we need to bother considering that latter possibility.

src/dask_awkward/io.py
```diff
@@ -55,3 +52,36 @@ def from_parquet(source: Any, **kwargs: Any) -> DaskAwkwardArray:
...
+    name = f"from-json-{token}"
+    _, chunks = read_bytes(
+        source,
+        delimiter=delimiter,
+        blocksize=blocksize,
+        sample=None,
+    )
+    chunks = list(flatten(chunks))
+    dsk = {(name, i): (concrete, d.key) for i, d in enumerate(chunks)}
+    hlg = HighLevelGraph.from_collections(name, dsk, dependencies=chunks)
```
martindurant, line 81: Once again, this looks awfully like a `Layer` that doesn't need to be materialised.

src/dask_awkward/core.py
```diff
@@ -591,8 +610,16 @@ def calculate_known_divisions(array: DaskAwkwardArray) -> tuple[int, ...]:
         Locations (indices) of division boundaries.
 
     """
+    # if divisions are known, quick return
     if array.known_divisions:
         return array.divisions  # type: ignore
+    # handle the case where we have 1 partition
+    if array.npartitions == 1:
```
martindurant, line 617: Do we need this special case? Note that this branch returns tuples rather than a list.

src/dask_awkward/core.py
```diff
@@ -1,14 +1,17 @@
...
 from functools import partial
 from numbers import Number
 from typing import TYPE_CHECKING, Any, Callable
 
 import awkward as ak
 import numpy as np
 from awkward._v2._connect.numpy import NDArrayOperatorsMixin
+from awkward._v2.highlevel import Array as Array_v2
 from awkward._v2.tmp_for_testing import v1_to_v2
+from awkward.highlevel import Array as Array_v1
```
martindurant, line 14: (to be deprecated, but not too soon - this will become the same as v2.Array)

tests/helpers.py
agoose77: I personally find the `pathlib` variant more readable :)

src/dask_awkward/core.py
```diff
@@ -30,10 +33,16 @@
 
 
 def _finalize_daskawkwardarray(results: Any) -> Any:
-    if all(isinstance(r, ak.Array) for r in results):
+    if len(results) == 1:
+        return results[0]
+    elif all(isinstance(r, Array_v1) for r in results):
```
martindurant, line 38: Check v2 first, to be ready for the v1->v2 transition

src/dask_awkward/io.py
douglasdavis: Agreed!

src/dask_awkward/core.py
douglasdavis: Without the special case `np.cumsum` is failing because the `array.map_partitions(ak.num, axis=0)` returns an int instead of an array. I guess we can force the computation to be an array if it's not already?

src/dask_awkward/io.py
douglasdavis: now handling multiple cases (line delim (single and multi file), single file, multi file)

src/dask_awkward/core.py
douglasdavis: Always converting results to arrays now

douglasdavis:
Merging with Windows tests failing; to be addressed later.

PR #15, opened 2021-12-02
Fix windows encoding issue

martindurant:
The problem happened because on windows, a linending is two bytes, not one, and so the unfortunate choice of default bytes size for a chunk gave an extra empty partition (no bytes). The change here fixes it for now, but we should attempt to prune such am empty partition or at least enforce a schema (which we already know) to all partitions.


martindurant:
Also, I moves tests inside the code tree, because that's typical for dask projects, and makes importing the code used as part of the tests easier.

douglasdavis:
Glad it was a simple fix!

martindurant:
OK to merge with the tests having moved? 
Do put "empty partitions" onto the list of things to fix in the future. We could also upstream a fix to dask so that read_bytes should generate roughly equal sized chunks rather than absolute offsets.

douglasdavis:
Yea I think it's a good idea to follow the standard Dask pattern

martindurant:
We probably need both :)

douglasdavis:
Added both
- handle empty partitions gracefully
- potential uptream dask.bytes patch to read equal sized chunks

to the running list

PR #17, opened 2021-12-03
parquet support

martindurant:
closes #13


douglasdavis:
With #29 in, after a rebase the tests should hopefully pass

douglasdavis:
https://github.com/ContinuumIO/dask-awkward/commit/8478f903e8041d8011cf441d6acb78aff8cfa95c should fix the only error I'm seeing in the tests

martindurant:
(actually, needs a fsspec release to pass, because this was the first github repo I came across with sub-repos)

jpivarski:
Just to let you know, I've started work on a non-Dask version of to/from_parquet: https://github.com/scikit-hep/awkward-1.0/pull/1294

The first step will be a complete Arrow schema → Awkward Form converter. It should follow nearly the same path as the Arrow schema + Arrow array → Awkward Array conversion and always result in the same Form.

There's a corner case in which the Form (and Type) of an Awkward Array can depend on values in the Arrow array, data not found in the schema: if an Arrow array node is not in a Field and has no metadata from Awkward (as an ExtensionType), then we only know if that node is nullable if its mask is null. That prevented round-trip conversions before we added metadata through an ExtensionType, and it's still possible we'll get a file with no metadata (didn't come from Awkward, or it came from an old version, or the user opted out). To cover that case, `ak.from_arrow` will have a `conservative_optiontype` argument. If `True`, then it will assume optiontype and create blank BitMaskedArrays just so the Form can be 100% predicted.

martindurant:
Unfortunately, parquet very rarely has required columns these days, even when they have no nulls... This prevents decompress-into-buffer in some cases.

douglasdavis:
@martindurant I know parquet stuff is in flux and a rewrite is quite possible, but just for fun I rewrote the graph-construction part of `to_parquet` to use be a HLG blockwise layer (using `map_partitions` and `BlockIndex`)

douglasdavis:
@martindurant should we rename `read_parquet` to `from_parquet` to [mirror the awkward API](https://github.com/scikit-hep/awkward/blob/352b0dead74846ad2a56d385be4694ec87072a08/src/awkward/_v2/operations/ak_from_parquet.py#L6)?

martindurant:
Bahh, too slow!
```
 ! [remote rejected] parquet2 -> parquet2 (cannot lock ref 'refs/heads/parquet2': is at 94bc6988b63610363e0199b56a742ead783e43f6 but expected 3766bda16552d065c5202b020696ede1a3984eaf)
error: failed to push some refs to 'https://github.com/martindurant/dask-awkward'
```

douglasdavis:
😛

src/dask_awkward/io/parquet.py
```diff
@@ -0,0 +1,519 @@
...
+                forms.NumpyForm("uint8"),
+                parameters={"__array__": "bytestring"},
+                form_key="",
+            )
+
+        elif isinstance(arrow_type, pyarrow.DataType):
+            if arrow_type == pyarrow.null():
+                return forms.EmptyForm(form_key="")
+            else:
+                dtype = np.dtype(arrow_type.to_pandas_dtype())
```
douglasdavis, line 154: @martindurant when trying to load the chicago taxi dataset I'm getting an exception here of the form:
```
> /Users/ddavis/software/repos/dask-awkward/src/dask_awkward/io/parquet.py(161)recurse()
-> dtype = np.dtype(arrow_type.to_pandas_dtype())
(Pdb) np.dtype(arrow_type.to_pandas_dtype())
*** NotImplementedError: dictionary<values=string, indices=int32, ordered=0>
(Pdb) arrow_type
DictionaryType(dictionary<values=string, indices=int32, ordered=0>)
```
where the `arrow_type` can't use `to_pandas_dtype()`.

martindurant: That is clearly *not* a numpy type, but it is something that awkward can handle (indexed array?).

`_parquet_scheme_to_form` was written some time ago by @jpivarski . We should probably switch to using ak.metadata_from_parquet , after checking that it copes with this input.

jpivarski: Arrow DictionaryType is equivalent to Awkward IndexedArray. `ak._v2.from_arrow` makes that translation and puts the word "categorical" in the IndexedArray's parameters; `ak._v2.to_arrow` only makes that translation if it finds the word "categorical". The reason is that if you build an Arrow Dictionary with non-unique values, it segfaults, and in general, Awkward IndexedArrays don't forbid non-unique values. The "categorical" parameter is a promise that they're unique.

The alternative to preserving a DictionaryType or IndexedArray is to evaluate it (`project` in IndexedArray). The data in this structure is a lazy array-slice:

```python
dictionary_values[dictionary_index]
```

where both are arrays. So if the Arrow Dictionary's values are a NumPy dtype, applying this procedure would give you a NumPy dtype. The type of the Dictionary is the type of its values.

Isn't `ak.metadata_from_parquet` just a high-level wrapper of `_parquet_scheme_to_form`? When the dust settles, we should have just one implementation of Arrow Schema ↔ Awkward Form. (I think that's already the case: I don't remember seeing two implementations of it.)

martindurant: > Isn't ak.metadata_from_parquet just a high-level wrapper of _parquet_scheme_to_form

No, it uses `ak._v2._connect.pyarrow.form_handle_arrow`. `_parquet_schema_to_form` only exists in dask-awkward.

martindurant:
The only thing missing is passing through the fsspec.parquet parameters, which are not hard coded in _file_to_partition, and calling ak's private function _load. I am OK with these for now.

PR #20, opened 2021-12-13
Terminology page

douglasdavis:
After some discussion in Slack I thought it would be a good idea to put something in writing. Obviously we can expand on this as needed. I quickly jotted these thoughts down, happy to get some more input! @jpivarski @martindurant


martindurant:
Good idea, and it gives us a clear place to define things

PR #28, opened 2022-01-20
Continued development

douglasdavis:
- **getitem**: Working on dedicated paths for single key vs tuple key inputs to `getitem` (see [discussion below](https://github.com/ContinuumIO/dask-awkward/pull/28#discussion_r789717041)).
  - Followup PR(s) to continue this
- **packaging**: improving setup.cfg
- **misc**: typetracer detection and utilities for grabbing typetracers from heterogenous mixed sequences of objects

### DIFF
--- a/src/dask_awkward/core.py
+++ b/src/dask_awkward/core.py
@@ -8,6 +8,7 @@ from numbers import Number
 from typing import TYPE_CHECKING, Any, Callable, Mapping, Sequence
 
 import awkward._v2 as ak
+import awkward._v2._typetracer as aktt
 import numpy as np
 from awkward._v2._connect.numpy import NDArrayOperatorsMixin
 from awkward._v2.highlevel import _dir_pattern
@@ -101,6 +102,14 @@ class Scalar(DaskMethodsMixin):
     def meta(self) -> Any | None:
         return self._meta
 
+    @meta.setter
+    def meta(self, m: Any | None) -> None:
+        if m is not None and not isinstance(
+            m, (aktt.MaybeNone, aktt.UnknownScalar, aktt.OneOf)
+        ):
+            raise TypeError(f"meta must be a typetracer module object, not a {type(m)}")
+        self._meta = m
+
     def __repr__(self) -> str:
         return self.__str__()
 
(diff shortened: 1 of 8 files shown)


src/dask_awkward/core.py
```diff
@@ -398,29 +398,32 @@ def partitions(self) -> IndexCallable:
...
-            lambda x, gikey: operator.getitem(x, gikey), name, self, gikey=key
+            lambda x, gikey: operator.getitem(x, gikey), name, self, gikey=where
         )
         hlg = HighLevelGraph.from_collections(name, graphlayer, dependencies=[self])
-        meta = self.meta[key] if self.meta is not None else None
+        meta = self.meta[where] if self.meta is not None else None
         return new_array_object(hlg, name, meta=meta, divisions=self.divisions)
 
-    def _getitem_single_int(self, key: int) -> Any:
+    def _getitem_single_int(self, where: int) -> Any:
```
agoose77, line 411: Hey @douglasdavis, I'm not too up to speed with everything happening here, so stop me if I'm off on a wild tangent. Should the `where` annotations now be unified to `tuple[Any, ...]`?

douglasdavis: Yes thanks for pointing this out! This still has the old typehint

jpivarski: If the function name is meaningful, `where` should still be an `int`, right?

If this is mirroring what happens in the Awkward codebase, there's a "gateway" function that takes any type of slice and dispatches it to a different function for each distinct type. That way, if some other function wants to internally call this with a known type, it doesn't have to go through the dispatch. Presumably, there's another function somewhere that handles tuples, rather than single ints.

(Something to note about how Awkward handles these, after you've passed it down from the Dask layer: Awkward dispatches to a different set of functions for tuple slices vs non-tuple slices. So even `array[slice]` is different from `array[slice,]`. The tuple case is more computationally intensive: it creates an extra RegularArray layer, performs the slice with copy semantics, and then removes the RegularArray layer from the result. Without the tuple, some slices can get away with view semantics, rather than copy semantics. We could "fix" that by putting in an extra check in our `__getitem__` gateway to see if the tuple has length 1 and choose a different path. However, there's a whole suite of tests relying on this behavior to test the tuple-pathway—if we "fix" that, those tests would become redundant and some aspects of the tuple-pathway wouldn't be tested. That's not a _good_ reason—we should just think of better tests. Yeah, maybe we should do that...)

jpivarski: Okay, I see this was fixed in f60946dacc781306482ae04719f3b77be3d036ae:

```diff
-    def _getitem_single_int(self, where: int) -> Any:
+    def _getitem_single_int(self, where: tuple[Any, ...]) -> Any:
```

The function name is probably wrong then, but I realize this is still cooking.

douglasdavis: @jpivarski  Thanks for that explanation. Indeed a dispatch pattern is going on here; the decision is made based on the 0th element of the tuple. After reading your explanation of how and why Awkward implements its dispatch pattern I think what I have here can be improved both performance-wise and consistency-with-awkward-wise by at least having dedicated non-tuple and tuple paths. Thanks again for the input, I'll keep playing with this.

PR #34, opened 2022-02-15
More slicing; starting back to work on reducers/structure; misc.

douglasdavis:
- support empty slices on comma-separated getitem calls
- adding a centralized "this is not supported" message with a link to open an issue for feature requests
- fixed some dtype checks
- other misc. things


src/dask_awkward/core.py
```diff
@@ -29,8 +29,9 @@
 
 _NOT_SUPPORTED_MSG = """
 
-If think this unsupported call should be supported by dask-awkward
-please open an issue at https://github.com/ContinuumIO/dask-awkward."""
+If you would like this unsupported call should be supported by
```
agoose77, line 32:
```suggestion
If you would like this unsupported call to be supported by
```

douglasdavis: ah thanks for the catch!

PR #37, opened 2022-03-09
cull in optimizations; better first partition compute

douglasdavis:
- Bring in `cull` to `__dask_optimize__` methods.
- Improve function used to compute the first partition. This function is used when an array collection is created with an unknown typetracer (meta).

### DIFF
--- a/src/dask_awkward/core.py
+++ b/src/dask_awkward/core.py
@@ -12,9 +12,12 @@ import awkward._v2 as ak
 import awkward._v2._typetracer as aktt
 import numpy as np
 from awkward._v2.highlevel import _dir_pattern
-from dask.base import DaskMethodsMixin, dont_optimize, is_dask_collection, tokenize
+from dask.base import DaskMethodsMixin
+from dask.base import compute as dask_compute
+from dask.base import is_dask_collection, tokenize
 from dask.blockwise import blockwise as upstream_blockwise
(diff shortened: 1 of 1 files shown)


src/dask_awkward/core.py
```diff
@@ -808,8 +822,11 @@ def _first_partition(array: Array) -> ak.Array:
         awkward array.
 
     """
-    computed = array.__dask_scheduler__(
-        array.__dask_graph__(), array.__dask_keys__()[0]
+    (computed,) = dask_compute(
```
douglasdavis, line 825: @martindurant thought you might be interested in this.

I ran into a problem with the original implementation we thought up:

```python
parts = array.__dask_graph__()[list(array.__dask_keys__())[0]]
f = parts[0]
args = parts[1:]
return f(*args)
```
because `args` can end up containing uncomputed keys (especially in the case of `from_json` where `read_bytes` is used.

I realized with the addition of culling we can just use `array.partitions[0]` and get the "speedup" we observed from improving this function... but that created infinite recursion because `new_array_object` will always try to compute a meta if it's passed None (and calling `array.partitions[i]` always eventually calls `new_array_object`.. hence the infinite recursion. Anyway... adding the ability to ignore meta (a bit of an "internal" kwarg) fixes everything.

PR #41, opened 2022-03-15
introduce more flexibility in json meta determination.

douglasdavis:


src/dask_awkward/io.py
```diff
@@ -61,15 +65,109 @@ def _from_json_bytes(source) -> ak.Array:
...
+    # for compressed data (or if explicitly asked for with
+    # force_by_lines set to True) we read the first `sample_rows`
+    # number of rows after opening the compressed file.
+    with fsspec.open(source, mode="rt", compression=compression) as f:
+        lines = []
+        for i, line in enumerate(f):
+            lines.append(json.loads(line))
+            if i >= sample_rows:
+                break
+        return ak.Array(ak.from_iter(lines).layout.typetracer.forget_length())
```
douglasdavis, line 109: @martindurant I'm curious what your thoughts are on this `derive_json_meta` function (called in the case of `from_json(..., meta=None)` (which is default)

PR #44, opened 2022-03-17
from_root (via uproot) proof-of-concept

douglasdavis:
PoC for simple ROOT to dask-awkward arrays


douglasdavis:
@nsmith- @jpivarski I whipped this together today to; it's very rough and just a PoC that isn't tested yet
- `UprootReadWrapper` is the internal callable that is used on each node of a graph by both
- `read_uproot` which tries to break up reading a single file based on `start` and `stop`
  - `a = read_uproot("/path/to/file.root", tree_name="tree", npartitions=5)`
- `read_uproot_files` which is one file per partition.
  - `a = read_uproot_files(["f1.root", "f2.root"], tree_name="tree")`

douglasdavis:
Closing this since uproot has taken ownership of ROOT -> dask-awkward

PR #46, opened 2022-03-23
Use awkward-datasets in tests

douglasdavis:


martindurant:
@jpivarski , I haven't looked at awkward-datasets, but might I suggest that an Intake driver and catalog might be a good idea for all awkward loading.

jpivarski:
I agree; there's a danger of ending up with too many of this style of package. For Uproot (mainly), there's already https://github.com/scikit-hep/scikit-hep-testdata, which would be nice to swap out for something more standard. (I'm not familiar enough with Intake drivers & catalogs, but something has to be better than putting small files in a PyPI package and pushing a new release before a PR can be CI-tested against new data.)

The files in scikit-hep-testdata are not exclusively ROOT files (there are a few LHEs), and Parquet files would be welcome. Moving all of these files to a place for "irregularly shaped data for CI tests" would put like with like. The ROOT files are definitely Awkward data.

douglasdavis:
Martin and I just discussed using intake, I'll keep this for now and rip it out once we have have intake infrastructure in place.

PR #50, opened 2022-04-13
Implement __iter__

douglasdavis:


src/dask_awkward/core.py
```diff
@@ -393,6 +393,10 @@ def __str__(self) -> str:
     def __repr__(self) -> str:  # pragma: no cover
         return self.__str__()
 
+    def __iter__(self):
+        for i in range(len(self)):
+            yield self[i]
```
martindurant, line 396: I feel like this should come with a warning! Does this force a compute of the divisions?

douglasdavis: Yes it will eagerly compute length. I'm writing a test now and it certainly feels _off_ to do this. I took inspiration from `dask.array.Array.__iter__`, I wonder if anyone _ever_ uses it

douglasdavis: Another strange find. `dask.dataframe.Series.__iter__` eagerly computes and returns a generator of values, while `dask.array.Array.__iter__` returns a generator of collectons... I imagine these features are very rarely used (if ever). Perhaps we should hard stop raise NotImplementedError as an unsupported use case?

martindurant: Yes, I think NotImplemented is good, unless we have a compelling use case. The message could point to iterating over values with a function passed to map_partitions.

PR #52, opened 2022-04-29
Awkward behaviors support

douglasdavis:
- [x] Add a function which serves as an equivalent for `x = ak.Array(x, with_name="foo")` but for collections.
- [x] Give the `Array` collection the ability to find behaviors attached to its metadata and call with `map_partitions`.
- [ ] ~~Find a way to communicate `ak.behavior` table when using `distributed` scheduler.~~ will address in a subsequent PR

### DIFF
--- a/src/dask_awkward/core.py
+++ b/src/dask_awkward/core.py
@@ -26,6 +26,7 @@ from numpy.lib.mixins import NDArrayOperatorsMixin
 from dask_awkward.optimize import optimize
 from dask_awkward.utils import (
     empty_typetracer,
+    hyphenize,
     is_empty_slice,
     normalize_single_outer_inner_index,
 )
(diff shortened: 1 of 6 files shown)


src/dask_awkward/core.py
```diff
@@ -434,6 +435,11 @@ def __dir__(self) -> list[str]:
             )
         )
 
+    def possible_behavior_methods(self) -> list[str]:
+        if self._meta is None:
+            return []
+        return list(set(dir(self._meta)) - set(dir(self)) - set(dir(ak.Array)))
```
martindurant, line 441: @jpivarski suggested to look at the definition of `dir`, which adds the fields, so that we can avoid subtracting them again

src/dask_awkward/core.py
```diff
@@ -739,8 +745,28 @@ def __getitem__(self, where: Any) -> Any:
 
         return self._getitem_single(where)
 
+    def _call_behavior(self, method_name: str, *args: Any, **kwargs: Any) -> Any:
+        try:
+            getattr(self._meta, method_name)
+        except AttributeError:
+            raise ValueError(f"{method_name} is not available to this collection.")
```
martindurant, line 752: Suggest that this ca still be an AttributeError

douglasdavis: yea makes sense

src/dask_awkward/core.py
douglasdavis: Yea I plan on thinking about this and experimenting a bit more (linking to https://github.com/scikit-hep/awkward-1.0/issues/1452 for reference)

PR #63, opened 2022-06-24
first partition compute strategy use delayed; typing

douglasdavis:


martindurant:
You'll need to explain this to me :)

PR #76, opened 2022-08-31
Add `AwkwardIOLayer` that with a `project_columns` method.

douglasdavis:
This PR tries to add column projection to an awkward read.

- [x] Update top level `__init__.py` such that Dask's scheduler doesn't need to import `awkward`.
- [x] Adding an `AwkwardIOLayer` blockwise layer that can recreate itself with a new `columns=` argument via a `project_columns` method.
- [ ] ~~Walk the type tracers in a task graph to determine the minimal set of columns that are used.~~ Will do this in follow up PR

(we're going to start with parquet as a first stab)


martindurant:
Dependant on pickle support in the scheduler, or can we set a config to make this package importable? 

Let me know when you want to go over this.

Before merger (add to bullet list?):
I would change `__init__.py` to not import the rest of the repo (or perhaps only /utils) to avoid the scheduler having to import awkward. That would be a bit annoying for users expecting top-level functions, so you can also add a `__getattr__` to the `__init__.py` and make sure that we supply results to `dir()` too.

douglasdavis:
I've programmatically added `dask_awkward` to dask's `distributed.scheduler.allowed-imports` config parameter (in `dask_awkward.config`.

About importing awkward.. yea I'll work on updating the top level __init__.py

douglasdavis:
All components of `dask_awkward` that require `awkward` are now in `dask_awkward.lib`. The new layer is in `dask_awkward.layers` (only depends on `dask`'s `blockwise` module)

We can now do the following:
- Add to `~/.config/dask/dask.yaml`:
```yaml
distributed:
  scheduler:
    allowed-imports:
      - "dask"
      - "distributed"
      - "dask_awkward"
```
- In an environment with `dask[complete]` installed and `dask-awkward` installed **but then explicitly uninstall awkward!**:
  - run `dask-scheduler` and spin up a couple of `dask-worker` processes
- then in an environment with `dask-awkward` (and `dask[complete]`)installed (and *not* uninstalling awkward...)
  - spin up an `IPython` shell,
  - run `client = Client("path-to-scheduler-in-other-environment")`
  - run `client.compute(dak.from_awkward(ak.Array([[1, 2, 3], [4]], npartitions=2)`
 
for some reason I can't seem to get the programmatically set config vars to work when I do this example distributed system setup.

martindurant:
@rjzamora , this is the first part of the optimization process I described. The part where we actually figure out which columns are needed to be able to process the graph is yet to come. The new layer is not yet keeping hold of any general metadata, but I think it will.

PR #80, opened 2022-09-14
Column projection: continue work on optimization add to dak.from_parquet

douglasdavis:
@martindurant Here is what I mentioned earlier today-- like I said,
haven't investigated the failures at all yet!


martindurant:
The cases that fail are where we keep all the columns without any manipulations at all. Removing any column from that doesn't cause an error!

PR #82, opened 2022-09-19
Guard against pyarrow import

douglasdavis:
@martindurant what do you think? This was mentioned here: https://github.com/scikit-hep/uproot5/pull/714#issuecomment-1250099755

If another part of `dask_awkward.lib` (e.g. `dask_awkward.lib.testutils`) is imported then `dask_awkward.lib.io.parquet` gets imported. Can you think of another module-organization method to avoid this other than adding the pyarrow import to the function bodies where necessary?

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -34,7 +34,7 @@ repos:
     language_version: python3
 
 - repo: https://github.com/asottile/pyupgrade
-  rev: v2.37.3
+  rev: v2.38.0
   hooks:
   - id: pyupgrade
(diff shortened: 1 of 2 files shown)


martindurant:
Deferring the import like this is totally fine with me. You can do all sort of tricks with importlib or custom classes, but I really don't think it's worth the complexity if there are to be not too many other occurrences.

The approach used in intake, for reference, is to make a `__getatttr__` in the parent module, which captures `intake.submodule` and does the import only when the attribute is accessed.

Issue #77, opened 2022-09-01
`dak.zip` support for non-`dict` iterables of arrays


masonproffitt:
For example:

```python
>>> import awkward._v2 as ak
>>> import dask_awkward as dak
>>> aa = ak.Array([[1, 2, 3], [], [2]])
>>> a = dak.from_awkward(aa, npartitions=3)
>>> dak.zip([a, a])
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/home/user/miniconda3/envs/func_adl_uproot_rc/lib/python3.10/site-packages/dask_awkward/structure.py", line 510, in zip
    raise DaskAwkwardNotImplemented("ak.zip only supports dictionary inputs.")
dask_awkward.utils.DaskAwkwardNotImplemented: ak.zip only supports dictionary inputs.

If you would like this unsupported call to be supported by
dask-awkward please open an issue at:
https://github.com/ContinuumIO/dask-awkward.
```

I assume the error text is a typo (should say `dak.zip`?), since `ak.zip` certainly supports more than just dictionary input:

```python
>>> ak.zip([aa, aa])
<Array [[(1, 1), (2, ...), (3, 3)], ...] type='3 * var * (int64, int64)'>
```

douglasdavis:
Thanks! fixed by https://github.com/ContinuumIO/dask-awkward/commit/105275b1937cce9a80a352af0b200d4e264f27f7

douglasdavis:
Non-dict iterable of arrays is now supported by `dak.zip`.

masonproffitt:
> Non-dict iterable of arrays is now supported by `dak.zip`.

It looks like `list`s are supported but not iterables in general. For example:

```python
>>> dak.zip((a, a))
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/home/user/miniconda3/envs/func_adl_uproot_rc/lib/python3.10/site-packages/dask_awkward/lib/structure.py", line 568, in zip
    raise DaskAwkwardNotImplemented("dict or list input is supported by dak.zip")
dask_awkward.utils.DaskAwkwardNotImplemented: dict or list input is supported by dak.zip

If you would like this unsupported call to be supported by
dask-awkward please open an issue at:
https://github.com/ContinuumIO/dask-awkward.
```

The exception is also confusing again, since it doesn't say that *only* dict or list input is supported.

douglasdavis:
[...]

PR #85, opened 2022-09-28
add list-of-arrays support for dak.zip

douglasdavis:


douglasdavis:
fixes #77

src/dask_awkward/lib/structure.py
```diff
@@ -506,40 +514,55 @@ def zip(
...
+            ),
+            *colls,
+            label="zip",
+            meta=meta,
+        )
 
-    return map_partitions(
-        _ZipFn(
-            keys,
+    elif isinstance(arrays, list):
```
masonproffitt, line 551: Why is this restricted to `list`s here? This precludes other iterables that should work (like `tuple`s)

PR #86, opened 2022-10-03
Update awkward import (remove `_v2`)

douglasdavis:
- rename import; awkward._v2 to awkward
- remove commented lines

### DIFF
--- a/.github/workflows/pypi-tests.yml
+++ b/.github/workflows/pypi-tests.yml
@@ -28,6 +28,7 @@ jobs:
       run: |
         pip install pip setuptools wheel -U
         pip install dask[complete]
+        pip install git+https://github.com/scikit-hep/awkward@main
         pip install --ignore-installed -q --no-cache-dir -e .[complete]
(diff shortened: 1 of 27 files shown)


.github/workflows/pypi-tests.yml
```diff
@@ -28,6 +28,7 @@ jobs:
       run: |
         pip install pip setuptools wheel -U
         pip install dask[complete]
+        pip install git+https://github.com/scikit-hep/awkward@main
         pip install --ignore-installed -q --no-cache-dir -e .[complete]
```
agoose77, line 32:
```suggestion
        pip install git+https://github.com/scikit-hep/awkward@main
        pip install -q --no-cache-dir -e .[complete]
```

douglasdavis: Thanks a lot! surprised this wasn't an issue earlier... what I get for not carefully adjusting CI configs when taking inspiration from another one of my projects  🤦

PR #92, opened 2022-10-25
use awkward 2.0.0rc1

douglasdavis:


jpivarski:
@agoose77, just what I was saying on https://github.com/scikit-hep/uproot5/pull/765. Thanks, @douglasdavis!

PR #93, opened 2022-10-28
Configurable column projection optimization

douglasdavis:


src/dask_awkward/awkward.yaml
```diff
@@ -8,10 +8,14 @@ awkward:
...
+  # - "brute-force": a brute force search for columns that are
+  #   required for the task graph to complete without raising an
+  #   IndexError (each column is held out and the metadata task graph
+  #   is computed as a requirement test)
+  # - "simple-getitem": checks for calls to getitem in the task graph.
+  #   This optimization is determined faster than the brute-force
+  #   method, but it may not be as complete w.r.t. eliminating
+  #   unncessary reads.
+  # - null, false, or "none": To turn off the optimization
+  column-projection-optimization: "simple-getitem"
```
martindurant, line 21: May as well add "chained" right now, which raises a NotImplemented until it's done.

tests/test_optimize.py
```diff
@@ -26,3 +28,17 @@ def test_requested_columns(caa_parquet):
...
+
+    a = from_parquet([caa_parquet] * 3)
+    with dask.config.set({"awkward.column-projection-optimization": "brute-force"}):
+        a.points.x.compute()
+
+
+def test_config_adjust_2(caa_parquet):
+    a = dak.from_parquet([caa_parquet] * 3)
+    with dask.config.set({"awkward.column-projection-optimization": "brute-force"}):
+        a.points.x.compute()
```
douglasdavis, line 44: @martindurant this test (`test_config_adjust_2`) is failing because the determined necessary columns evaluates to (as expected):
```
["points.x"]
```
and supplying that as the `columns=` arg is causing an empty to array to be created at compute.

The test above it (`test_config_adjust_1`) is working. It uses my old "dumb" `from_parquet` (in my scratch module), which uses `ak.from_parquet` directly instead of the `_load` implementation that your `dak.from_parquet` is using.

I'm wondering if we need to adjust `project_columns` in your implementation to also change the schema? That's the only difference I can see between my scratch implementation (which doesn't use schema) and your implementation that uses `_load`.

This will create a file you can use to test:

```python
a = [{'points': [{'x': 6, 'y': 5}, {'x': 9, 'y': 3}, {'x': 9, 'y': 4}]},
 {'points': []},
 {'points': [{'x': 4, 'y': 1}, {'x': 4, 'y': 8}]},
 {'points': [{'x': 2, 'y': 10}]},
 {'points': [{'x': 5, 'y': 6}, {'x': 2, 'y': 4}, {'x': 7, 'y': 2}]},
 {'points': [{'x': 6, 'y': 5}, {'x': 9, 'y': 3}, {'x': 9, 'y': 4}]},
 {'points': []},
 {'points': [{'x': 4, 'y': 1}, {'x': 4, 'y': 8}]},
 {'points': [{'x': 2, 'y': 10}]},
 {'points': [{'x': 5, 'y': 6}, {'x': 2, 'y': 4}, {'x': 7, 'y': 2}]},
 {'points': [{'x': 6, 'y': 5}, {'x': 9, 'y': 3}, {'x': 9, 'y': 4}]},
 {'points': []},
 {'points': [{'x': 4, 'y': 1}, {'x': 4, 'y': 8}]},
 {'points': [{'x': 2, 'y': 10}]},
 {'points': [{'x': 5, 'y': 6}, {'x': 2, 'y': 4}, {'x': 7, 'y': 2}]}]

a = ak.from_iter(a)
ak.to_parquet(a, "caa.parquet", extensionarray=False)
```

martindurant: Looking. Note that ak.from_parquet also calls _load!

martindurant: It seems that the columns being passed down to ak's loader is `["points.x"]`, but it should be `["points.list.item.x"]`

martindurant: In dak.from_parquet, `parquet_columns` is computed for the given selection of ak columns, but not stored anywhere. When projecting columns, we need to map the chosen columns back to their parquet verbose names, and that list is what gets passed to `_load()`.

douglasdavis: Ah it's because `ak.from_parquet` is converting to the columns while `_load` so load is expecting the converted columns. When we call `_load` in `_file_to_partition` we need to convert

douglasdavis: Or we can make the column projection yield converted columns

martindurant: Yes. But we only know which of these are "listy" and whether they use "item" or "element" while scanning the metadata (which we don't want to repeat), so we probably have to carry a {ak_col: pq_col} mapping around with _FromParquetFn rather than just a list of desired columns, if that class is to keep responsibility for amending the list of columns we pass through.

Ah, but we also have the schema, so never mind - the only thing we need to store is the "item" versus "element", I think, and then we have all the information available within _FromParquetFn.project_columns.

martindurant:
```diff
--- a/src/dask_awkward/lib/io/parquet.py
+++ b/src/dask_awkward/lib/io/parquet.py
@@ -16,9 +16,10 @@ from dask_awkward.lib.io.io import from_map


 class _FromParquetFn:
-    def __init__(self, columns=None, schema=None):
+    def __init__(self, columns=None, schema=None, listsep="list.item"):
         self.columns = columns
         self.schema = schema
+        self.listsep = listsep

     @abc.abstractmethod
     def __call__(self, source):
@@ -30,8 +31,8 @@ class _FromParquetFn:


 class _FromParquetFileWiseFn(_FromParquetFn):
-    def __init__(self, fs, columns, schema):
-        super().__init__(columns=columns, schema=schema)
+    def __init__(self, fs, columns, schema, listsep):
+        super().__init__(columns=columns, schema=schema, listsep=listsep)
         self.fs = fs

     def __call__(self, source):
@@ -43,10 +44,12 @@ class _FromParquetFileWiseFn(_FromParquetFn):
         )

     def project_columns(self, columns):
+        schema = self.schema.select_columns(columns)
         return _FromParquetFileWiseFn(
             fs=self.fs,
-            columns=columns,
-            schema=self.schema,
+            columns=schema.columns(self.listsep),
+            schema=schema,
+            listsep=self.listsep
         )
```

(untested)

PR #94, opened 2022-10-31
feat: JSON: refactor IO for JSON; make input implementations column optimizable

douglasdavis:
More JSON development to enable schema <--> layout/form round trips (motivated by column projection optimizations)


douglasdavis:
Copying @jpivarski's comment from Slack w.r.t. missing things to check for

- `is_indexed`: just pass through; "indexed" are invisible to types, but the case needs to be considered
- `{"__array__": "bytestring"}`: in addition to `"string"`
- `is_regular`: can JSONSchema specify that all lists in a collection have the same type? I think it can.
- `is_unknown`: an unconstrained type, like what you do with `existing_schema is None`
- `is_union`: we don't support UnionArrays when reading ak.from_json with a JSONSchema, but for this direction (to JSONSchema), it would be easy
- `{"__array__": "sorted_map"}` always contains a RecordArray that is a 2-tuple (no field names) that specifies key and value types for a mapping (not used by Awkward yet, but we've reserved the word)
- in principle, you could check `{"__array__": "categorical"}` for the strings that it contains (part of the layout, not the Form/Type) if the nested type is string, then this could be JSONSchema with an enumerated set of strings. (This may be mixing types and values too much. Everything else here is type level.)

src/dask_awkward/lib/io/json.py
```diff
@@ -30,221 +25,222 @@
...
+    *,
+    storage_options: dict[str, Any] | None = None,
+) -> tuple[AbstractFileSystem, str, list[str]]:
+    fs, token, paths = get_fs_token_paths(source, storage_options=storage_options)
+
+    # if paths is length 1, check to see if it's a directory and
+    # wildcard search for JSON files. Otherwise, just keep whatever
+    # has already been found.
+    if len(paths) == 1 and fs.isdir(paths[0]):
+        paths = list(filter(lambda s: ".json" in s, fs.ls(paths[0])))
```
douglasdavis, line 47: use `fs.find`

src/dask_awkward/lib/io/json.py
```diff
@@ -30,221 +25,222 @@
...
+
+    return fs, token, paths
```
douglasdavis, line 49: (maybe) raise if paths is empty raise FileNotFound

PR #97, opened 2022-11-04
More top level structure functions

douglasdavis:


src/dask_awkward/lib/structure.py
```diff
@@ -171,9 +171,25 @@ def copy(array):
...
+        self.kwargs = kwargs
+
+    def __call__(self, arr):
+        return ak.fill_none(arr, self.value, **self.kwargs)
+
+
 @borrow_docstring(ak.fill_none)
-def fill_none(array, value, axis=-1, highlevel=True, behavior=None):
-    raise DaskAwkwardNotImplemented("TODO")
+def fill_none(
```
martindurant, line 184: Are there still many of these? Feels like we can establish a pattern to generate them if so.

douglasdavis: Yea there are still a number of these hanging around. I'm hesitant to generate them because of `axis=`, but the general pattern of creating small callable class has been worked well so far.

PR #99, opened 2022-11-14
parquet fixes

douglasdavis:
this PR fixes `from_parquet` in two areas:

1. the case where the we have an "unnamed root" in a column (e.g. `.list.item.column.list.item`)
2. the case where splitting by row groups created a task graph that was one element too small


douglasdavis:
Ah based on the failing test it looks like the behavior was relied on elsewhere

PR #102, opened 2022-11-15
to_parquet: mkdirs

douglasdavis:
@martindurant does this make sense? (without this `dak.to_parquet(array, "some_dir")` crashes if `some_dir` doesn't exist)


martindurant:
You should probably call `mkdirs(path, exists_ok=True)`, and apply to all filesystems. It will be a no-op usually on s3, but even there you might be asking for a new bucket. You definitely don't want an error because the directory already exists.

PR #105, opened 2022-11-16
Add `with_parameter` and `without_parameters`

douglasdavis:


douglasdavis:
I'm not quite sure how to test this off the top of my head; @jpivarski or @lgray any thoughts?

lgray:
A good test would be to add and remove an empty dictionary, make sure it is there, make sure it is not there.

lgray:
And yes, indeed, check the results after a compute! I'll also be needing to manipulate the parameters of the dak.Array as well. So both sides need to be appropriately persitent and correct.

douglasdavis:
Alright I've added a test that adds and removes a trivial parameter, it makes sure that the collection metadata has the parameter, as well as the computed result.

PR #117, opened 2022-12-08
add test for no awkward imported

douglasdavis:


douglasdavis:
Yeah I'm trying to make sure we can import dask_awkward (and dask_awkward.layers) without import awkward. I also wanted to move the imports to the top level `__init__.py` for a better IDE experience.

On the pyarrow thing. Are you suggesting we also install pyarrow in this new `no-awkward-test` GH action and make sure we can import both dask_awkward and pyarrow but not import awkward?

martindurant:
Sorry, typing in multiple places gets signals interfering...

I mean, that importing dask-awkward for the use of layers should not import awkward even if it is imported. You could test this by making a fake awkward package after pip-removing it, and having that mutate something checkable at import time.
Sorry for mistakenly mentioning arrow.
```python
import os
os.IAMHERE = True
```

douglasdavis:
nixing this for now based on our earlier conversation

PR #118, opened 2022-12-08
add some _ipython* methods and remove some warnings

douglasdavis:
- ipython methods needed to conserve the mechanism we use to check for
  behaviors (methods on the metadata)

- remove string fill value for fill_none tests; raises warning and
  computes first partition due to typetracer operation error


douglasdavis:
automatic metadata determination is failing for `dak.mean`, `dak.var`, and `dak.std` (which raises a warning and computes the first partition) because the `ak.` versions appear to not support operation on typetracers

Issue #123, opened 2022-12-14
`dak.concatenate`


masonproffitt:
Not implemented:

```python
    @borrow_docstring(ak.concatenate)
    def concatenate(
        arrays,
        axis=0,
        merge=True,
        mergebool=True,
        highlevel=True,
        behavior=None,
    ):
>       raise DaskAwkwardNotImplemented("TODO")
E       dask_awkward.utils.DaskAwkwardNotImplemented: TODO
E
E       If you would like this unsupported call to be supported by
E       dask-awkward please open an issue at:
E       https://github.com/dask-contrib/dask-awkward.
```

martindurant:
Of the incoming issues, I suggest this is the easiest one to support, for any value of `axis`. I'm not sure we would support all of the optional argiuments.

masonproffitt:
The main use case here for me is for combining collections of similar objects that need the same operations applied to them, like putting electrons and muons together for applying identical lepton selections

douglasdavis:
`axis>=0` support is in; we can handle future needs with another issue

PR #125, opened 2022-12-14
Add `dak.concatenate`

douglasdavis:
Fix for #123 

Covers `axis>=0`


src/dask_awkward/lib/operations.py
```diff
@@ -0,0 +1,27 @@
...
+    for collection in arrays:
+        metas.append(collection._meta)
+        for k in collection.__dask_keys__():
+            g[(name, i)] = k
+            i += 1
+
+    meta = ak.concatenate(metas)
+
+    hlg = HighLevelGraph.from_collections(name, g, dependencies=arrays)
+    return new_array_object(hlg, name, meta=meta, npartitions=npartitions)
```
martindurant, line 27: If the incoming arrays all know they partition sizes, then we can know the partition sizes of the output?

src/dask_awkward/lib/operations.py
```diff
@@ -0,0 +1,27 @@
+from __future__ import annotations
+
+import awkward as ak
+from dask.base import tokenize
+from dask.highlevelgraph import HighLevelGraph
+
+from dask_awkward.lib.core import Array, new_array_object
+
+
+def concatenate(arrays: list[Array]) -> Array:
```
martindurant, line 10: Missing ak arguments, particularly `axis=`. This version is for `axis=0`.

src/dask_awkward/lib/operations.py
douglasdavis: Yep! working on that now

src/dask_awkward/lib/operations.py
douglasdavis: >This version is for axis=0.

Right, always starting with the easiest 😄 

I'm going through a few experiments to see how much of `axis=` we can support. There will definitely be some restrictions

jpivarski:
For `axis>0`, the lengths of each partition does not change.

```python
>>> one = ak.Array([[1, 2, 3], [], [100, 101]])
>>> two = ak.Array([[4, 5], [10, 11, 12, 13], [102]])
>>> ak.concatenate([one, two], axis=1).show()
[[1, 2, 3, 4, 5],
 [10, 11, 12, 13],
 [100, 101, 102]]
```

You just need to ensure that all of the arrays going into the concatenation, `one`, `two` (and maybe more) all have the same length (error if they don't) and can all be partitioned the same way (maybe repartition if not?). Then you can pass partition `i` of each array to a separate worker that uses Awkward's `ak.concatenate` on all the arrays for partition `i` and call the results for each partition the final result.

This isn't one, like `ak.flatten`, that changes the lengths of each partition, because the length is defined by the number of lists, not the number of elements in them.

douglasdavis:
I'm going to go ahead and merge this in for `axis>=0` support; we can expand on this more in the future.

PR #135, opened 2022-12-17
Use filter to remove task keys from getitem discovery

douglasdavis:
This PR is a followup to #131 - a better check to make sure we don't discover getitem calls that are actually dask tasks e.g.

```python
array[array.x > 5]
```

the `array.x > 5` is a dask task with name `greater-[0-9a-f]{32}`


douglasdavis:
Won't be needed after #141/#148

Issue #121, opened 2022-12-14
`dak.unzip`


masonproffitt:
Now that Uproot 5 (and thus `uproot.dask`) is officially released, I'm going to start making issues for these `NotImplemented` errors when I run into them:

```python
    @borrow_docstring(ak.unzip)
    def unzip(array, highlevel=True, behavior=None):
>       raise DaskAwkwardNotImplemented("TODO")
E       dask_awkward.utils.DaskAwkwardNotImplemented: TODO
E
E       If you would like this unsupported call to be supported by
E       dask-awkward please open an issue at:
E       https://github.com/dask-contrib/dask-awkward.
```

masonproffitt:
The use case here is of course to be able to invert the `zip()` operation, which for me is mainly important for being able to access the individual members of tuples created by `combinations()` or `cartesian()`

lgray:
I started working on this one and while I see a few hacky ways to deal with a tuple of arrays being returned, I couldn't really decide a nice way to do it given the current infrastructure. 

@douglasdavis I think the best way to do this is to maintain the appearance of the awkward interface which return `tuple(ak.Array)`, but it wasn't really clear to me how to make an efficient graph for this using `dask.HighLevelGraph` as there isn't any immediate documentation on indexing into returned products. I guess the evolution of the meta would also have to understand how to interpret the tuple.

PR #136, opened 2022-12-23
Add dak.unzip

lgray:
Fixes https://github.com/dask-contrib/dask-awkward/issues/121

### DIFF
--- a/src/dask_awkward/lib/structure.py
+++ b/src/dask_awkward/lib/structure.py
@@ -8,11 +8,12 @@ import awkward as ak
 import dask.delayed
 import numpy as np
 from awkward._typetracer import UnknownScalar
+from dask.base import tokenize
 
+import dask_awkward
 from dask_awkward.lib.core import (
     Array,
     map_partitions,
-    new_array_object,
     new_known_scalar,
     total_reduction_to_scalar,
 )
@@ -506,28 +507,54 @@ def unzip(
     if not highlevel:
         raise ValueError("Only highlevel=True is supported")
 
+    _unzip_label = "unzip"
+
     # this has to get a little weird right now
     # someone else can probably determine a better impl
 
     meta = ak.unzip(array._meta, highlevel=highlevel, behavior=behavior)
     if isinstance(meta, tuple):
+        len_meta = len(meta)
         unzip_fn = _UnzipFn(highlevel=highlevel, behavior=behavior)
-        bound = dask.delayed(pure=True, nout=len(meta))(unzip_fn)(array)
-        # bound = graph_manip_bind(bound, array)
-        return tuple(
-            new_array_object(
-                bound[imeta].dask,
-                bound[imeta]._key,
-                meta=ameta,
-                divisions=array.divisions,
+
+        temp_delayed = array.to_delayed()
+
+        unzipped_delayed_list = [
+            dask.delayed(pure=True, nout=len(meta))(unzip_fn)(delayed_array)
+            for delayed_array in temp_delayed
+        ]
+
+        unzipped_delayed_by_meta = []
+
+        for ipart, unzipped_delayed in enumerate(unzipped_delayed_list):
+            unzipped_delayed_by_meta.extend(
+                [
+                    unzipped_delayed.__getitem__(
+                        i, dask_key_name=("getitem-" + tokenize(array, i), ipart)
+                    )
+                    for i in range(len_meta)
+                ]
             )
-            for imeta, ameta in enumerate(meta)
-        )
+
+        unzipped_arrays = []
+        for i, imeta in enumerate(meta):
+            unzipped_arrays.append(
+                dask_awkward.from_delayed(
+                    tuple(
+                        unzipped_delayed_by_meta[len(meta) * j + i]
+                        for j in range(array.npartitions)
+                    ),
+                    meta=imeta,
+                    prefix=_unzip_label,
+                )
+            )
+
+        return tuple(unzipped_arrays)
     else:
         return map_partitions(
             _UnzipFn(highlevel=highlevel, behavior=behavior),
             array,
-            label="unzip",
+            label=_unzip_label,
             output_divisions=1,
         )
 
--- a/src/dask_awkward/lib/structure.py
+++ b/src/dask_awkward/lib/structure.py
@@ -531,18 +531,17 @@ def unzip(
         # here I pop out the necessary portion of dak.from_delayed to skip unncessary steps in the dag and remove unnecessary loops
         unzipped_arrays = []
         for i, imeta in enumerate(meta):
-            parts = [
-                unzipped_delayed_list[ipart].__getitem__(
-                    i,
-                    dask_key_name=(
-                        f"{_unzip_label}-getitem-{tokenize(array, i)}",
-                        ipart,
-                    ),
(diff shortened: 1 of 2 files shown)


lgray:
Could a maintainer please run this :-)

lgray:
@douglasdavis all the reducers appear to be failing, my test passes.

lgray:
Fixes #121

lgray:
So I'm *really* surprised this new implementation works, but it does so long as provide an empty typetracer to the actual `ak.unzip` step and then fill in what the typetracer should be when exploding the `getitems` into that tuple. This seems delicate but it's better than generating a bazillion dask-delayed tasks for large input datasets.

lgray:
Could someone kick this PR for tests? Thanks!

lgray:
@jpivarski @martindurant your thoughts on the present implementation would be useful. As noted above I really get the feeling it's delicate, despite being quite effective.

martindurant:
Just trying to wrap my head around the implications of this. Some thoughts:
- we create an intermediate graph layer that has output type awkward, but actually produces a tuple of awkwards
- that layer has an empty typetracer which is not descriptive of what the layer is 
- downstream layers will not work with optimization in the case that not all the branches are actually needed for computing

It seems to me that a cleaner description would be to_delayed(), which makes it explicit that the intermediate thing is not a dask-awkward collection, and that we don't expect to be able to optimize through it.

OR, perhaps better, there should only be a single layer, which fuses unzip/getitem into a single function for each branch. This would return a tuple of dak objects as now. Yes, you would effectively repeat the ak.unzip call for evey branch, but that's a cheap metadata-only operation, so it's fine.

jpivarski:
As I understand it, the issue is that `unzip` returns a tuple, rather than a Dask object, such as a `dak.Array` (or `dask-histogram`, or `dask.array`, etc.). You use an empty type-tracer to make this function look enough like a function that returns a Dask object that `map_partitions` doesn't complain. Is that the issue?

Thinking about it now, `dak.unzip` should not be sending any work to the workers (no calls to `map_partitions`): it's purely a metadata rearrangement. This

```python
ak.unzip(some_ak_array)
```

is entirely equivalent to this

```python
tuple(some_ak_array[field_name] for field_name in ak.fields(some_ak_array))
```

It must be the case that `dak.fields` doesn't even produce a Dask node, right? That function can/should return its result during the graph-creation pass; it should be

```python
def dak.fields(dak_array):
    return ak.fields(dak_array._meta)
```

The getitem-field step, `some_dak_array[field_name]`, is just metadata-rearrangement, but it needs to be a Dask node because it is an operation that needs to run on the workers (the return value is a Dask type). But it has already been implemented, and a `dak.unzip` implementation can use that.

If it's done the above way, then a `dak.unzip(some_dak_array)` in the source code would translate into a bunch of `some_dak_array[field_name]` nodes in the task graph, one for each field, but then the Dask graph consists entirely of nodes that input Dask types and output Dask types, no non-Dask tuples.

Wouldn't this be a better way to implement it? Looking at the Dask graph, it wouldn't be possible to determine that the source code contained `dak.unzip` instead of manually calling getitem-field, but I don't think that reversibility is important.

martindurant:
@jpivarski , I think your description and mine are identical in effect, but I didn't explain it as well :)  Indeed, simply not having `ak.unzip` in the call at all is simpler.

lgray:
Got it - thanks! Addressed this, please run tests :-)

martindurant:
Does this fail if the input is not a record type? ak.unzip also supports fixed-length upmost level (becomes N elements of the tuple) and other (becomes one-element tuple).

I think in this special case, we should add to the docstring, specifying that this function procudes a tuple of dak objects.

jpivarski:
I actually didn't know what it was going to do. What it does in Awkward looks like reasonable behavior:

```python
>>> ak.fields(ak.Array([1, 2, 3]))   # this much I knew; it's intentional
[]
>>> ak.unzip(ak.Array([1, 2, 3]))    # this is an interesting surprise, not unwelcome
(<Array [1, 2, 3] type='3 * int64'>,)
```

In the case of no fields, there's no attempt to get `array[field_name]`, but the resulting tuple is not empty (as the code in my previous comment would have it). It's a length-1 tuple of the original array.

> I think your description and mine are identical in effect

Sorry that I didn't catch that. (We were writing our comments at the same time, without seeing each others'.)

lgray:
```python3
    fields = ak.fields(array._meta)
    if len(fields) == 0:
        return tuple([array])
    else:
        return tuple(array[field] for field in ak.fields(array._meta))
```
? could also do an unzip on the meta of the input array.

jpivarski:
What you have looks right to me (although you can reuse the already-computed `fields` variable).

I don't think you want to run `ak.unzip` on the `array._meta`, since you need to end up with Dask arrays, not plain Awkward arrays (which would be type-tracers, anyway). It has to be Dask-in, Dask-out.

Although I guess you could run `ak.unzip` on the meta and then wrap all outputs as Dask, but then you'd have to preserve the partitioning—it just seems easier to do it manually.

martindurant:
Yes, I think that. There should also be a test for the fixed-top-level ("tuple-like") case - @jpivarski , what is the right condition for that? It should give
```
        return tuple(array[i] for i in range(N))
```
where `N` is the fixed size length.

jpivarski:
The `ak.fields` approach (which returns strings like `"0"`, `"1"`, `"2"` for tuple-type records) works for both tuple-type and named-field records. There's no need to write specialized code for this case.

```python
>>> array = ak.Array([(1, 1.1, "one"), (2, 2.2, "two")])
>>> array
<Array [(1, 1.1, 'one'), (2, 2.2, ...)] type='2 * (int64, float64, string)'>

>>> ak.unzip(array)
(<Array [1, 2] type='2 * int64'>,
 <Array [1.1, 2.2] type='2 * float64'>,
 <Array ['one', 'two'] type='2 * string'>)

>>> ak.fields(array)
['0', '1', '2']

>>> array["0"]
<Array [1, 2] type='2 * int64'>
>>> array["1"]
<Array [1.1, 2.2] type='2 * float64'>
>>> array["2"]
<Array ['one', 'two'] type='2 * string'>
```

(If you were to use `range(N)`, you'd have to ensure that the values are strings, so that they're not interpreted as row indexes: `str(i)`. But `ak.fields` already gives you that.)

martindurant:
Oh ok, thanks for clarifying. Might be worth a test just like your example code.

lgray:
please rerun tests

(maybe talking to the bot works if you're a new contributor?)

src/dask_awkward/lib/structure.py
```diff
@@ -489,8 +489,16 @@ def unflatten(array, counts, axis=0, highlevel=True, behavior=None):
...
-def unzip(array, highlevel=True, behavior=None):
-    raise DaskAwkwardNotImplemented("TODO")
+def unzip(
+    array: Array, highlevel: bool = True, behavior: dict | None = None
+) -> tuple(Array):
+    if not highlevel:
+        raise ValueError("Only highlevel=True is supported")
+    fields = ak.fields(array._meta)
+    if len(fields) == 0:
+        return tuple([array])
```
martindurant, line 499:
```suggestion
        return (array, )
```

PR #137, opened 2022-12-29
Add dak.with_field

lgray:
This is necessary since field assignment into `dak.Array` RecordArrays cannot be done with square brackets.

### DIFF
--- a/src/dask_awkward/lib/structure.py
+++ b/src/dask_awkward/lib/structure.py
@@ -10,11 +10,16 @@ from awkward._typetracer import UnknownScalar
 
 from dask_awkward.lib.core import (
     Array,
+    compatible_partitions,
     map_partitions,
     new_known_scalar,
     total_reduction_to_scalar,
 )
-from dask_awkward.utils import DaskAwkwardNotImplemented, borrow_docstring
+from dask_awkward.utils import (
+    DaskAwkwardNotImplemented,
+    IncompatiblePartitions,
+    borrow_docstring,
+)
 
 if TYPE_CHECKING:
     from numpy.typing import DTypeLike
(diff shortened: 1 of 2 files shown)


martindurant:
Please not that @douglasdavis is on leave, and I will be covering for a while. I'll try to get to all of these cases as fast as I can.

> assignment into dak.Array RecordArrays cannot be done with square brackets

Eventually, we may follow what dask.dataframe does to allow this, but agreed that this is the right thing to do for the present (and should always be allowed anyway).

lgray:
Oh - I'm not necessarily expecting responses during this period anyway, but thank you! Just queuing stuff up.

martindurant:
I would remove "highlevel" as a kwarg here and elsewhere, since for dask-awkward, we are always high-level.

jpivarski:
What if `highlevel` is a keyword argument in all the places where it is in the associated `ak.*` functions, but setting `highlevel=False` raises an exception? That would be an efficient way to communicate to users that they're not missing something—it just can't be done.

martindurant:
Yes, @jpivarski , we could raise exceptions. We would be testing it in a lot of places, though. I think having an argument in the signature that you cannot actually change is somewhat annoying (even if the ak version has it), I would prefer clear documentation.

lgray:
Looking through the rest of `structure.py`, `highlevel` is inconsistently enforced. However, we know it's never allowed. I'd make a choice for dealing with that, but in another PR since it would change nearly every user-facing function. No good reason to block functionality that will not change with an alteration to a higher level design choice.

martindurant:
OK to dealing with highlevel systematically in a new PR.

Issue #140, opened 2023-01-08
Segfault on compute() when loading specific columns in from_parquet.


riga:
Hi!

I really appreciate the feature to load only specific, arbitrarily nested columns from a parquet file 🎉 This really boosts our IO.

However, I spotted a case where calling `compute()` on a `dak` array results in a segfault, depending on which exact columns were passed to `from_parquet()`. You can find a test snippet and an example input file at [this gist](https://gist.github.com/riga/6e04cae06d65cb78f17ae625d83de0b0), but the short story is the following.

If a file contains some nested columns, say

```
array
├─ Jet
│  ├─ pt
│  ├─ mass
│  └─ ... (others!)
└─ ... (others)
```

and I load all columns under "Jet" via `from_parquet(file, columns=["Jet.*"])` everything works as expected. But as soon as I load only a *subset of columns* under "Jet" such as `from_parquet(file, columns=["Jet.pt"])`, `compute()` causes a segfault.

I repeated this on different files to make sure none of them was corrupted, but the error occurred consistently.

- Python 3.10.8
- dask_awkward 2023.1.0
- awkward 2.0.5

douglasdavis:
Hi @riga, thanks for raising the issue with the minimal reproducer! We'll start digging around

martindurant:
@douglasdavis , are you working on this? I suspect it's a mismatch with those .list.item. column name bits that parquet has.

martindurant:
The following test passed
```python
def test_nested_columns(tmpdir):
    tmpdir = str(tmpdir)
    data = [
        {
            "Jet": {
                "pt": 0,
                "mass": 1,
            }
        },
        {
            "Jet": {
                "pt": 1,
                "mass": 2,
            }
        },
    ]
    arr = dak.from_awkward(ak.from_iter(data), 2)
    to_parquet(arr, tmpdir)
    arr1 = dak.from_parquet(tmpdir)
    arr2 = dak.from_parquet(tmpdir, columns=["Jet.*"])
    assert_eq(arr1, arr2)
    arr2 = dak.from_parquet(tmpdir, columns=["Jet.pt"])
    assert "mass" not in arr2.Jet.fields
    assert arr2.Jet.pt.compute().tolist() == [0, 1]
```

Looking at the gist example next.

martindurant:
@jpivarski , a segfault should not happen ever, right?
[...]

PR #141, opened 2023-01-10
New optimization

martindurant:
Using the [new API](https://github.com/scikit-hep/awkward/pull/2027#discussion_r1056548262) available on awkward typetracers

### DIFF
--- a/src/dask_awkward/layers/layers.py
+++ b/src/dask_awkward/layers/layers.py
@@ -72,11 +72,12 @@ class AwkwardIOLayer(Blockwise):
         form = new_meta.layout.form
         _label_form(form, self.name)
         new_labelled_meta, report = ak._typetracer.typetracer_with_report(form)
+        nblock = list(self.numblocks.values())[0][0]
         return (
             AwkwardIOLayer(
                 name=self.name,
                 columns=self.columns,
-                inputs=[None],
+                inputs=[None] * nblock,
                 io_func=lambda *_, **__: ak.Array(new_labelled_meta),
                 label=self.label,
                 produces_tasks=self.produces_tasks,
--- a/src/dask_awkward/layers/layers.py
+++ b/src/dask_awkward/layers/layers.py
@@ -61,7 +61,7 @@ class AwkwardIOLayer(Blockwise):
             if form.is_record:
                 for field in form.fields:
                     _label_form(
-                        form.content(field), f"{start}.{field}" if start else field
+                        form.content(field), f"{start}.{field}"
                     )
             elif form.is_numpy:
                 form.form_key = start
@@ -70,7 +70,7 @@ class AwkwardIOLayer(Blockwise):
 
         new_meta = typetracer_from_form(copy.deepcopy(self._meta.layout.form))
         form = new_meta.layout.form
-        _label_form(form, "")
+        _label_form(form, self.name)
         new_labelled_meta, report = ak._typetracer.typetracer_with_report(form)
         return (
             AwkwardIOLayer(
--- a/src/dask_awkward/layers/layers.py
+++ b/src/dask_awkward/layers/layers.py
@@ -60,31 +60,32 @@ class AwkwardIOLayer(Blockwise):
         def _label_form(form, start):
             if form.is_record:
                 for field in form.fields:
-                    _label_form(form.content(field), f"{start}.{field}" if start else field)
+                    _label_form(
+                        form.content(field), f"{start}.{field}" if start else field
+                    )
             elif form.is_numpy:
                 form.form_key = start
             else:
                 _label_form(form.content, start)
 
-        new_meta = typetracer_from_form(
-            copy.deepcopy(self._meta.layout.form)
-        )
+        new_meta = typetracer_from_form(copy.deepcopy(self._meta.layout.form))
         form = new_meta.layout.form
         _label_form(form, "")
-        new_labelled_meta, report = ak._typetracer.typetracer_with_report(
-            form
+        new_labelled_meta, report = ak._typetracer.typetracer_with_report(form)
+        return (
+            AwkwardIOLayer(
+                name=self.name,
+                columns=self.columns,
+                inputs=[None],
+                io_func=lambda *_, **__: ak.Array(new_labelled_meta),
+                label=self.label,
+                produces_tasks=self.produces_tasks,
+                creation_info=self.creation_info,
+                annotations=self.annotations,
+                meta=new_labelled_meta,
+            ),
+            report,
         )
-        return AwkwardIOLayer(
-            name=self.name,
-            columns=self.columns,
-            inputs=[None],
-            io_func=lambda *_, **__: ak.Array(new_labelled_meta),
-            label=self.label,
-            produces_tasks=self.produces_tasks,
-            creation_info=self.creation_info,
-            annotations=self.annotations,
-            meta=new_labelled_meta,
-        ), report
 
     def project_columns(self, columns: list[str]) -> AwkwardIOLayer:
         if hasattr(self.io_func, "project_columns"):
--- a/src/dask_awkward/layers/layers.py
+++ b/src/dask_awkward/layers/layers.py
@@ -46,26 +46,45 @@ class AwkwardIOLayer(Blockwise):
             annotations=None,
         )
 
-    def project_and_mock(self, columns: list[str]) -> AwkwardIOLayer:
+    def __repr__(self):
+        return f"AwkwardIOLayer<{self.output}>"
+
+    def mock(self) -> Any:
 
         # imported here because it this method should be run _only_ on
         # the Client (which is allowed to import awkward)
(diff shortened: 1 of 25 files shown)


martindurant:
I only fixed the first check in test_optimize, this is to show what I think the idea is

To do:
- account for to_delayed or other layers where we know we must touch all columns

I realise I don't actually know how dask optimised the case where you request multiple outputs, dask.optimise(array1, array2)

tests/test_parquet.py
```diff
@@ -163,7 +163,33 @@ def test_write_roundtrip(tmpdir):
     tmpdir = str(tmpdir)
     arr = dak.from_awkward(ak.from_iter(data), 2)
     to_parquet(arr, tmpdir)
-    arr = dak.from_parquet(tmpdir)
+    dak.from_parquet(tmpdir)
+
+
+def test_nested_columns(tmpdir):
```
martindurant, line 169: Sorry, this doesn't belong here, but in the parquet segfault discussion #140 . It can be removed if problematic.

martindurant:
Still fails tons of things, not sure why.
- sometimes, the number of partitions in an array becomes 1
- processing the typetracer yields UnknownScalar in other places, which cannot be made back into ak.Arrays.

martindurant:
> Is this something that would benefit from UnknownScalar becoming a 0-D array, as is planned?

Probably yes. I am not certain, though.

src/dask_awkward/layers/layers.py
```diff
@@ -46,25 +46,46 @@ def __init__(
...
+        new_meta = typetracer_from_form(copy.deepcopy(self._meta.layout.form))
+        form = new_meta.layout.form
+        _label_form(form, self.name)
+        new_labelled_meta, report = ak._typetracer.typetracer_with_report(form)
+        nblock = list(self.numblocks.values())[0][0]
+        return (
+            AwkwardIOLayer(
+                name=self.name,
+                columns=self.columns,
+                inputs=[None] * nblock,
```
martindurant, line 80: We need all the inputs rather than just the first one now, even though we try only get the first output "partition" while reporting. This is because some actions like aggregations work across all partitions. Can this possibly expand the time it takes to run a graph for optimization purposes?

src/dask_awkward/lib/io/json.py
```diff
@@ -189,6 +189,8 @@ def _from_json_files(
         mode="rb",
         storage_options=storage_options,
     )
+    if isinstance(urlpath, list) and len(urlpath) > len(urlpaths):
+        urlpaths = urlpath
```
martindurant, line 193: I don't know why, but get_fs_token_paths was returning only one path. It *is* reasonable, given that the paths in the tests are often identical, but not what was happening before. This hack fix may fail is passing many glob string, of which most don't find any files (unlikely).

src/dask_awkward/lib/optimize.py
```diff
@@ -97,166 +59,44 @@ def _projectable_io_layer_names(dsk: HighLevelGraph) -> list[str]:
...
+    import awkward as ak
+    layers = dsk.layers.copy()
+    deps = dsk.dependencies.copy()
+    reports = {}
+    for name in _projectable_io_layer_names(dsk):
+        layers[name], report = layers[name].mock()
+        reports[name] = report
+    hlg = HighLevelGraph(layers, deps)
+    # TODO: outlayers should be any place we need all the columns, whether output
+    #  or to_* where data leaves dak
```
martindurant, line 77: as far as I can tell, `outputs` (passed by dask) is equal to `list(hlg.layers.values())[-1].keys()` in all the cases I can see.

src/dask_awkward/lib/optimize.py
```diff
@@ -97,166 +59,44 @@ def _projectable_io_layer_names(dsk: HighLevelGraph) -> list[str]:
...
+    outlayer = list(hlg.layers.values())[-1]
+    out = get_sync(hlg, list(outlayer.keys())[0])
```
martindurant, line 79: Any writing operation (e.g., to_parquet) will at this point attempt to **create files**, although they will be empty.

src/dask_awkward/lib/optimize.py
```diff
@@ -97,166 +59,44 @@ def _projectable_io_layer_names(dsk: HighLevelGraph) -> list[str]:
...
-
+    deps = dsk.dependencies.copy()
+    for name, ttr in reports.items():
+        cols = set(ttr.data_touched)
+        select = []
+        for col in cols:
+            if col is None:
+                continue
+            n, c = col.split(".", 1)
+            if n == name:
```
martindurant, line 99: I think this is always True, at least for the simple linear graphs we have so far

martindurant (commented):
@douglasdavis , some thoughts for the current state of this PR. You can see that a good number of tests pass again, but there are issues with UnknownScalars and other typetracer-typetracer operations.

src/dask_awkward/layers/layers.py
douglasdavis: Yea I wonder that as well. We should be able to check this without too much difficulty.

src/dask_awkward/lib/optimize.py
douglasdavis: Yea I think that's the point of the dask optimization API taking two arguments: the graph and the final necessary key(s). Are you using `list(hlg.layers.values())[-1].keys()` because we had that in the previous implementation?

src/dask_awkward/lib/optimize.py
douglasdavis: Ah I can't immediately thing of a way around this

tests/test_parquet.py
douglasdavis: Doesn't bother me to have have this in- seems like a good test to have anyway!

douglasdavis:
I think I'm nearly up to speed on this new strategy. Might have a couple of questions for you later this afternoon. Definitely nice to see lot's of red in the diff!

lgray:
@nsmith- This seems to be getting close. Your fsspec pr https://github.com/scikit-hep/uproot5/pull/692 might be worth resurrecting.

lgray:
@martindurant 

> I realise I don't actually know how dask optimised the case where you request multiple outputs, dask.optimise(array1, array2)

Just looking at how things are shaking out it appears that we'll be operating in this way almost exclusively for realistic analysis (usually with histograms or other aggregators at the end, `dak.to_parquet` is also likely to be popular here).

lgray:
Some initial explorations of optimized graphs over multiple outputs seem pretty decent though, fwiw.

src/dask_awkward/lib/optimize.py
```diff
@@ -97,166 +60,96 @@ def _projectable_io_layer_names(dsk: HighLevelGraph) -> list[str]:
...
+    for arg in args + tuple(kwargs.values()):
+        if isinstance(arg, ak.Array):
+            arg.layout._touch_data(recursive=True)
+
+
+def _mock_output(layer: Layer):
+    assert len(layer.dsk) == 1
+    layer2 = copy.deepcopy(layer)
+    mp = layer2.mapping.copy()  # why is this always a MetrializedLayer?
+    key = iter(mp).__next__()
```
douglasdavis, line 81: I've been playing around with this today. When an output layer is present if we don't modify each partition's write step the optimization will fail because `dask_awkward.lib.io.parquet._write_partition` is still getting called on a typetracer.

For example, if we have 3 partitions we'll end up with a graph that looks like

```python
{
    (write-parquet-abc123, 0): (_mock_io_func, ..., 0),
    (write-parquet-abc123, 1): (_ToParquetFn, ..., 1),  # <-- raises b/c callable is getting a typetracer
    (write-parquet-abc123, 2): (_ToParquetFn, ..., 2),
}
```

```suggestion
    for k in iter(mp.keys()):
        mp[k] = (_mock_output_func,) + mp[k][1:]
```

martindurant: OK, right - I think this was meant to go in conjunction with the usual route of selecting just one output key, but of course you can't have both those code paths at once.

src/dask_awkward/lib/optimize.py
```diff
@@ -97,166 +60,97 @@ def _projectable_io_layer_names(dsk: HighLevelGraph) -> list[str]:
...
+            arg.layout._touch_data(recursive=True)
+
+
+def _mock_output(layer: Layer):
+    assert len(layer.dsk) == 1
+    layer2 = copy.deepcopy(layer)
+    mp = layer2.mapping.copy()  # why is this always a MetrializedLayer?
+    for k in iter(mp.keys()):
+        mp[k] = (_mock_output_func,) + mp[k][1:]
+    mp[key] = (_mock_io_func, ) + mp[key][1:]
```
douglasdavis, line 83:
```suggestion
```
forgot to remove this line in the previous suggestion

douglasdavis:
closed in favor of #148

PR #143, opened 2023-01-12
Take ak.num meta from form.length_zero_array()

lgray:
Workaround for https://github.com/scikit-hep/awkward/issues/2092.

This just precalculates the meta for axis > 0 cases using the typetracer's `form.length_zero_array()`


douglasdavis:
Jim's bubbling the root issue up on his todo list to have it addressed upstream https://github.com/scikit-hep/awkward/issues/1997#issuecomment-1380570575

PR #130, opened 2022-12-14
Add `dak.argsort` for `axis != 0`

douglasdavis:

PR #146, opened 2023-01-14
Add more ak verbs

### LINKED ISSUE #124: `dak.argsort`
Not implemented:

```python
    @borrow_docstring(ak.argsort)
    def argsort(
        array,
        axis=-1,
        ascending=True,
        stable=True,
        highlevel=True,
        behavior=None,
    ):
>       raise DaskAwkwardNotImplemented("TODO")
E       dask_awkward.utils.DaskAwkwardNotImplemented: TODO
E
E       If you would like this unsupported call to be supported by
E       dask-awkward please open an issue at:
E       https://github.com/dask-contrib/dask-awkward.
```

### LINKED ISSUE #128: `dak.broadcast_arrays`
Not implemented:

```python
  File "/home/user/miniconda3/envs/func_adl_uproot_rc/lib/python3.10/site-packages/dask_awkward/lib/structure.py", line 108, in broadcast_arrays
    raise DaskAwkwardNotImplemented("TODO")
dask_awkward.utils.DaskAwkwardNotImplemented: TODO

If you would like this unsupported call to be supported by
dask-awkward please open an issue at:
https://github.com/dask-contrib/dask-awkward.
```

lgray:
Fixes #124
Fixes #128

add:
- argcartesian
- argcombinations
- isclose
- singletons
- argsort (for axis > 0, from #130)
- sort (for axis > 0)
- appropriate error for copy
- full_like (needs issue on awkward opened for meta calculation)
- unflatten (currently simplistic implementation, needs work to be general since it requires repartitioning, but very important/necessary)
- to_packed
- ravel (with warnings about recordarrays)
- run_lengths (with warning about axis=0)
- from_regular
- to_regular
- broadcast_arrays (#128)

remove:
- concatenate impl in `structure.py`

### DIFF
--- a/src/dask_awkward/lib/structure.py
+++ b/src/dask_awkward/lib/structure.py
@@ -69,6 +69,14 @@ __all__ = (
 )
 
 
+class _ArgCartesianFn:
+    def __init__(self, **kwargs):
+        self.kwargs = kwargs
+
+    def __call__(self, *arrays):
+        return ak.argcartesian(list(arrays), **self.kwargs)
+
+
 @borrow_docstring(ak.argcartesian)
 def argcartesian(
     arrays,
@@ -79,7 +87,27 @@ def argcartesian(
     highlevel=True,
     behavior=None,
 ):
+    if axis == 1:
+        fn = _ArgCartesianFn(
+            axis=axis,
+            nested=nested,
+            parameters=parameters,
+            with_name=with_name,
+            highlevel=highlevel,
+            behavior=behavior,
+        )
+        return map_partitions(fn, *arrays, label="argcartesian", output_divisions=1)
     raise DaskAwkwardNotImplemented("TODO")
+
+
+class _ArgCombinationsFn:
+    def __init__(self, n: int, axis: int, **kwargs: Any):
+        self.n = n
+        self.axis = axis
+        self.kwargs = kwargs
+
+    def __call__(self, array):
+        return ak.argcombinations(array, self.n, axis=self.axis, **self.kwargs)
 
 
 @borrow_docstring(ak.argcombinations)
@@ -94,6 +122,26 @@ def argcombinations(
     highlevel=True,
     behavior=None,
 ):
+    if fields is not None and len(fields) != n:
+        raise ValueError("if provided, the length of 'fields' must be 'n'")
+
+    if axis != 0:
+        fn = _ArgCombinationsFn(
+            n=n,
+            replacement=replacement,
+            axis=axis,
+            fields=fields,
+            parameters=parameters,
+            with_name=with_name,
+            highlevel=highlevel,
+            behavior=behavior,
+        )
+        return map_partitions(
+            fn,
+            array,
+            label="argcombinations",
+            output_divisions=1,
+        )
     raise DaskAwkwardNotImplemented("TODO")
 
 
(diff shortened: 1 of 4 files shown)


douglasdavis (commented):
This is looking good to me. Something I noticed is a warning coming from the `ak.full_like` test:
```
/Users/ddavis/.pyenv/versions/3.11.1/envs/dev/lib/python3.11/site-packages/awkward/operations/ak_full_like.py:160: DeprecationWarning: elementwise comparison failed; this will raise an error in the future.
    if fill_value == 0 or fill_value is _ZEROS:
```
Getting triggered in upstream awkward, but for some reason I'm not able to reproduce it outside of the tests. Do you have a handle on what's going on here?

lgray:
@douglasdavis Yeah - I ran into the same weird irreproducibility problem, and I haven't had time to dig into to it really. Just implementing so far. I can take a look by the end of the week.

For `dak.unflatten` I think we can get away with even this simplistic implementation so long as we can put some better checks on array compatibility. For instance, if we know that `counts` came from an earlier `dak.num` (this is actually a common pattern: `flatten -> do stuff -> unflatten`), which will have counts appropriate to the partitioning of the input array, we can easily allow that array through. I'm not sure if we can do that lazily with what we have here and now, but perhaps with a little analysis of the task graph we could verify things like that?

Otherwise the complete implementation requires the ability to repartition and shuffle dask-awkward arrays, which I think right now is step `N+1` (much much lower priority than optimized reading and faster graph optimization).

I'd like to add yet a few more verbs so we can chew on this. I'd like to check with @nsmith- to make sure we've got all the mostly common ones. I think with the PR we have so far we're pretty good but would like to get regular-use coverage mostly done.

lgray:
@agoose77 just FYI there are a number of TODO's in this PR that come from needing to `.length_zero_array()` a form to create an output form and then a typetracer from that. Instead of the operation on the typetracer completing successfully. I'm not sure if it's worth opening an awkward ticket for each one since it's pretty common.

lgray:
@masonproffitt could you give this branch a try?

lgray:
@douglasdavis pending a bit of user testing this is good to go now.

src/dask_awkward/lib/structure.py
```diff
@@ -467,22 +654,97 @@ def pad_none(
 
 @borrow_docstring(ak.ravel)
 def ravel(array, highlevel=True, behavior=None):
-    raise DaskAwkwardNotImplemented("TODO")
+    if not highlevel:
+        raise ValueError("Only highlevel=True is supported")
+
+    if isinstance(array._meta.layout, ak.contents.recordarray.RecordArray):
+        warnings.warn("ravel may produce inconsistent results for record arrays!")
```
douglasdavis, line 661: What do you mean by inconsistent results here. Should we just not allows it if it can't be consistent?

douglasdavis (commented):
Added one inline question, but other than that if you have nothing else to add I think this is good to go!

src/dask_awkward/lib/structure.py
lgray: They are correct results but out of order row-wise comapared to awkward. So not wrong but possibly confusing if validating against raw awkward.

douglasdavis: OK that seems like a reasonable thing we can field if a question comes up

douglasdavis:
@lgray just want to confirm you have nothing else you'd like to add to this PR and I'll go ahead and merge!

masonproffitt:
> @masonproffitt could you give this branch a try?

I tested this, and the only new problem that came up was #151

PR #148, opened 2023-01-19
Refactor column optimization based on typetracer updates

douglasdavis:
Before this PR we recently created two different optimization strategies for column projection:
- the "simple getitem" strategy: searching for `getitem` layers that use strings or lists of strings
- the "brute-force" strategy: drop fields from the starting IO layer(s) one at a time and see if a compute on the metadata fails

With this PR we now use the new `typetracer_with_report` API from upstream `awkward` to create a mutable report object, compute the task graph on purely the metadata (typetracer arrays), and return a list of necessary columns for each IO layer in the graph.

There are still some outstanding upstream typetracer array issues that need to be ironed out before this optimization works for the current dask-awkward test suite.


src/dask_awkward/lib/optimize.py
```diff
@@ -86,177 +49,122 @@ def _projectable_io_layer_names(dsk: HighLevelGraph) -> list[str]:
...
+        out = get_sync(hlg, list(outlayer.keys())[0])
+    except Exception as err:
+        on_fail = dask.config.get("awkward.optimization.on-fail")
+        if on_fail == "warn":
+            warnings.warn(f"Column projection optimization failed: {type(err)}, {err}")
+            return {}
+        elif on_fail == "pass":
+            return {}
+        else:
+            raise
```
douglasdavis, line 138: @martindurant what do you think about this try-except block?

martindurant: This is reasonable. 

- Should the config parameter be allowed a "-" (as opposed to "_")?
- can we be more specific than Exception? Maybe not.

douglasdavis: > Should the config parameter be allowed a "-" (as opposed to "_")?

I've tested the config and `-` is OK (dask.array uses dashes), the envvar can be written as `DASK_AWKWARD__OPTIMIZATION__ON_FAIL`

> can we be more specific than Exception? Maybe not.

Yea this crossed my mind as well, as of right now it can be (`AttributeError, ValueError, NotImplementedError, AssertionError`) and catch everything in the current test suite but (a) a list of 4 is kind of big IMO and (b) I'm worried about something else popping up that we don't yet know about.

douglasdavis: The default is set to `warn` (pinged you in other comment on the diff for the config file)

> If "pass", maybe there should still be a logging.debug call in that block? Maybe add a comment labeling the branch corresponding to the current default.

Sounds good to me

martindurant: > The default is set to warn

OK, then Exception should be fine, and we may look to narrow it down in the future.

martindurant: As with the previous version of optimising, I think it would be useful to allow users a way to run the find-column code on their own dak object without the rest of optimize and without compute. That will help build understanding and should appear in tutorial docs.

douglasdavis: Yea I've kept the `dask_awkward.lib.inspect.necessary_columns` public function in place. I'll add it to the toplevel `dask_awkward` import

martindurant:
Good to go, right? 
I wonder if there's a convenient way to count how many warnings we are currently generating in the tests - presumably the same number as we had failing tests before.

douglasdavis:
> Good to go, right?

I think it would be reasonable to bring this in as is and start addressing the warnings in follow up PRs

@lgray would you be able to run a quick workflow of yours that you may be experimenting with? I think it would be useful to see how this PR affects things you've been working on.

> I wonder if there's a convenient way to count how many warnings we are currently generating in the tests - presumably the same number as we had failing tests before.

There are actually a few more now that @lgray's #146 PR is in (which is good, just more information we can report upstream as holes in the typetracer)

lgray:
@douglasdavis Sure, I can give this a try later today.

lgray:
@douglasdavis 

```
/Users/lgray/coffea-dev/dask-awkward/src/dask_awkward/lib/optimize.py:137: UserWarning: Column projection optimization failed: <class 'ValueError'>, cannot convert UnknownScalar(dtype('int64')) (type UnknownScalar) to an array element

(https://github.com/scikit-hep/awkward/blob/awkward-cpp-7/awkward-cpp/src/python/content.cpp#L191)
  warnings.warn(f"Column projection optimization failed: {type(err)}, {err}")
```

I don't have any introspection as to what part of my code is causing it though. Interestingly the results come back just fine?

lgray:
The dict I'm passing into `dask.compute` looks like:
```
{'mass': Hist(
  StrCategory([], growth=True, name='dataset'),
  Regular(30000, 0.25, 300, name='mass'),
  storage=Double()) # (has staged fills), 'pt': Hist(
  StrCategory([], growth=True, name='dataset'),
  Regular(30000, 0.24, 300, name='pt'),
  storage=Double()) # (has staged fills), 'cutflow': {'ZJets_pt': dask.awkward<sum-agg, type=Scalar, dtype=Unknown>, 'ZJets_mass': dask.awkward<sum-agg, type=Scalar, dtype=Unknown>, 'Data_pt': dask.awkward<sum-agg, type=Scalar, dtype=Unknown>, 'Data_mass': dask.awkward<sum-agg, type=Scalar, dtype=Unknown>}}
```

lgray:
If I try to ask for `_necessary_columns` on one of the outputs, I get the following:
image

lgray:
Also it would be great if `_necessary_columns` was a public interface! I will need to do a bit of analysis on user code that will require it.

douglasdavis:
We do know about the UnknownSclalar error (this one: `cannot convert UnknownScalar(dtype('int64')) (type UnknownScalar) to an array element`), reported upstream here https://github.com/scikit-hep/awkward/issues/2135

> I don't have any introspection as to what part of my code is causing it though. Interestingly the results come back just fine?

So the route we've gone down here for now is to catch the exceptions that are raised during the optimization and then throw that warning that you are seeing (the warning that says the optimization failed). We do still expect the computation to be correct, we just bailed on the column projection optimization (so I'm glad you see that the results come back just fine). Martin and I were just talking about a way to avoid the exceptions from awkward and still run the optimization. This would be a workaround we can use until the upstream typetracer development is complete and the exceptions no longer get raised by awkward. I'm going to start experimenting with the workaround today.

> Also it would be great if _necessary_columns was a public interface! I will need to do a bit of analysis on user code that will require it.

a `necessary_columns` function makes it into the top level module in this PR (you should be able to call `dak.necessary_columns`)

lgray:
I guess my question is then: can I reliably use `necessary_columns` on workflows that have non-dak operations in them?

douglasdavis:
tl;dr: Since the optimization is based on typetracer arrays, right now the operations in the workflow need to be typetracer array aware.

More explanation taking dask-histogram as an example: If we create a graph from the following code:

```python
import dask_histogram.boost as dhb
import dask_awkward as dak
ds = dak.from_parquet("s3://ddavistemp/hpq", storage_options={"anon": True})
pt = dak.flatten(ds.muons.pt)
h = dhb.Histogram(dhb.axis.Regular(25, 0, 250))
h.fill(pt)
h.compute()
```
This compute will work, but only the dask_histogram optimizations will run because we called compute on the dask_histogram object; that is, this:

https://github.com/dask-contrib/dask-histogram/blob/d3ebda03385ab4892dddb65bded9ee6613bcfc5c/src/dask_histogram/core.py#L274-L276

function is called. Which does vanilla dask optimizations (not specific to dask-awkward). [you can see dask-histogram's optimize is just calling functions from upstream dask](https://github.com/dask-contrib/dask-histogram/blob/d3ebda03385ab4892dddb65bded9ee6613bcfc5c/src/dask_histogram/core.py#L202-L217)

However, if we call `dak.necessary_columns`:

```python
dak.necessary_columns(h)
```

We see

```python
/Users/ddavis/software/repos/dask-awkward/src/dask_awkward/lib/optimize.py:150: UserWarning: Column projection optimization failed: <class 'AssertionError'>, bug in Awkward Array: attempt to convert TypeTracerArray into a concrete array

See if this has been reported at https://github.com/scikit-hep/awkward-1.0/issues
  warnings.warn(f"Column projection optimization failed: {type(err)}, {err}")
```

In this case the graph is getting analyzed by dask-awkward, which knows about `AwkwardInputLayer`; so the optimization is _attempted_, but it fails because there's a step in the graph calling boost_histogram code, which is unaware of awkward typetracers (hence the error where a typetracer arrays is attempting to be converted to a concrete numpy array). This is happening in a boost-histogram fill via this line in dask-histogram:

https://github.com/dask-contrib/dask-histogram/blob/d3ebda03385ab4892dddb65bded9ee6613bcfc5c/src/dask_histogram/core.py#L194-L195

`data` in that code block is a simply typetracer array during the optimization compute pass, and boost-histogram doesn't know anything about typetracer arrays.

It should be possible to add to dask-histogram a feature to make it run dask-awkward optimizations if dask-awkward is installed. But we still need to solve the issue of exceptions getting raised due to external libraries not knowing about typetracers. This is do-able too, but we need to figure out the best way to mock these layers in the type-tracer-only graph.

lgray:
Ah, right this would be taken care of by the `empty_if_typetracer` stuff that @agoose77 is working on, right? Since it's so common I think we should make sure dak -> dask-histogram works nearly out of the box.

agoose77:
Ultimately, any code executed by Dask with a typetracer has the potential to fail. We're planning for `UnknownScalar` to become `TypeTracerArray(..., shape=())`, which will unify the error semantics; any operation requiring concrete values will fail.

There are two facets to this:
- Ensuring that third-party code doesn't operate on typetracer arrays
   - Plan is to add support to `hist` for this e.g. with a structured-histogram protocol that we're thinking about at the moment (see [here](https://gist.github.com/agoose77/f755d96061a36795dda592276d63f506) for some musings of mine on this, not actually planned yet. Incidentally, this wouldn't be much use right now given that `dask-histogram` uses `boost-histogram` instead of `hist`.
- Ensuring that user code doesn't operate on typetracer arrays

In both cases, the current idea that we're pursuing is to export an `empty_if_typetracer` function that returns the equivalent of `form.length_zero_array()` if the array is a typetracer *and* calls `touch_data()`. This is an explicit, manual step, but it will ensure that the user code avoids any conditionals. 

Note that `empty_if_typetracer` was originally dreamt up for Coffea, but it's not clear to me how needed it still is. It would certainly be useful for anything going in to `map_partitions`.

lgray:
@agoose77 There are a few places in coffea that need to call a `map_partitions`, particularly here I just want dask-histogram to yield to typetracer analysis, which (I think) will also solve the dask-histogram issue.

agoose77:
`empty_if_typetracer` would be useful in such a scenario, e.g.

```python
def fill_user_hist(data):
    p = np.sqrt(data.px**2 + data.py**2 + data.pz**2)
    p_is_small = p < 10    
    
    # Maybe this will only be a single-arg function, who knows.
    # Ensure bogus data are available for histogramming
    p, p_is_small = ak.empty_if_typetracer(p, p_is_small)
    
    p_hist = Hist.new.Bool().Reg(128, 0, 1000).Int64()
    p_hist.fill(p_is_small, p)

dak.map_partitions(fill_user_hist, array)
```

lgray:
Ah - I see. This kind of stuff never goes anywhere near the common user in my perception of things, it should be there for them as a pressure relief valve or for power users, but the libraries should do a good portion of common functionality themselves without much complaining.

agoose77:
Yes, this is an escape hatch. Really, it's an L2 function (public API for advanced users) who need to know about typetracers. For developers of libraries, it's less useful because you'd already use `ak.backend(layout)` to determine whether you had a typetracer array, and then you could do `array.layout.form.length_...` if you wanted a zero length array.

So in this case, `empty_if_typetracer` shouldn't come in to if if it's non-awkward library code, e.g. Coffea, dask. If such knowledge exists at the task-graph level in a convenient fashion, we should be able to handle dask-histogram explicitly, but I'll defer to the Anaconda folks as to the practical feasibility of this.

douglasdavis:
I agree agree with @lgray's comment on it being a pressure relief valve. We can benefit (on the dask-awkward side as library authors) from usng `empty_if_typetracer` to help ensure the optimization operates as desired (the implementation of the optimization is never exposed to users as a public API). We will always go with direct operation on typetracers as a first try, but being able to fall back to a mocked function call which can wrap the original function but operate on something created by `empty_if_typetracer` is a good thing to have in the toolbox

douglasdavis:
And second, we can include dask-awkward optimizations in dask-histogram for graphs where dask-awkward's `AwkwardInputLayer` is detected.

agoose77:
Just to be clear, the empty_if_typetracer proposal is something users would need to add themselves to the functions that get passed in to map_partitions. Otherwise, all you can do is materialise/touch *everything* in the array, which really should work most of the time (unless the function actually cares about array lengths). Then of course there's the existing 'run one partition, take the meta' strategy as a third option.

douglasdavis:
> Just to be clear, the empty_if_typetracer proposal is something users would need to add themselves to the functions that get passed in to map_partitions.

Yea this makes sense, we already mock output layers in our typetracer-only-graph (which is one of the main components of this PR!) and touch the data without writing anything to disk. 

```python
def _mock_output_func(*args: Any, **kwargs: Any) -> Any:
    import awkward as ak

    for arg in args + tuple(kwargs.values()):
        if isinstance(arg, ak.Array):
            arg.layout._touch_data(recursive=True)


def _mock_output(layer: Any) -> Any:
    assert len(layer.dsk) == 1

    new_layer = copy.deepcopy(layer)
    mp = new_layer.mapping.copy()
    for k in iter(mp.keys()):
        # this layer tuple is originally is of the form (ak.to_parquet, ...), but we
        # replace the to_parquet or to_json, etc. with our mocked_output_func
        # which just touches the data
        mp[k] = (_mock_output_func,) + mp[k][1:]
    new_layer.mapping = mp
    return 
```

What @martindurant  and I have in mind is to mock (really wrap) all other layers in the typetracer-only-graph and try to call whatever function `func` that layer is calling, and if the function call fails on the pure typetracer we can fall back to calling `func` on an object created by `empty_if_typetracer`.

lgray:
Cool! Got it to work:
image

Code in dask-histogram is like:
```python3
def _blocked_dak(data: Any, *, histref: bh.Histogram | None = None) -> bh.Histogram:
    import awkward as ak
    thedata = data
    print(type(thedata))
    if isinstance(thedata, ak.Array) and ak.backend(thedata) == "typetracer":
        thedata.layout._touch_data(recursive=True)
        thedata = data.layout.form.length_zero_array()

    return clone(histref).fill(thedata)


def _blocked_dak_ma(*data: Any, histref: bh.Histogram | None = None) -> bh.Histogram:
    import awkward as ak
    thedata = list(data)
    for idata, adatum in enumerate(thedata):
        if isinstance(adatum, ak.Array) and ak.backend(adatum) == "typetracer":
            adatum.layout._touch_data(recursive=True)
            thedata[idata] = adatum.layout.form.length_zero_array()

    return clone(histref).fill(*tuple(thedata))
```

lgray:
And furthermore, this output is correct since you need all the 4-momentum variables to calculate that histogram.

lgray:
and the `pt` histogram optimizes correctly too:
image


@douglasdavis want me to make that a PR over to dask-histogram or do you want to cook that one yourself due to reasons I'm not seeing?

lgray:
I have another request here for `dak.necessary_columns`: it would be immediately for the best if we can make it have the same sort of interface as `dask.compute/optimize/visualize` where we can pass in (nested) basic python containers holding dask collections and get the union of the required columns over all dask collections in the input object.

Can likely lift the necessary modifications almost directly from compute or optimize.

lgray:
Arg - something about this branch is causing the `_meta` to loose its parameters when getting files from uproot. Digging.

douglasdavis:
> and the `pt` histogram optimizes correctly too: image
> 
> @douglasdavis want me to make that a PR over to dask-histogram or do you want to cook that one yourself due to reasons I'm not seeing?

A PR in dask-histogram sounds good!

douglasdavis:
> I have another request here for `dak.necessary_columns`: it would be immediately for the best if we can make it have the same sort of interface as `dask.compute/optimize/visualize` where we can pass in (nested) basic python containers holding dask collections and get the union of the required columns over all dask collections in the passed input.

Yes I think this is a great idea :+1:

lgray:
So I've been trying some more exotic things and ran into a problem. I'm trying to apply one of our more complex NanoEvents schemas atop the simple awkward from we get from `uproot.dask`.

The code to do this looks roughly like:
```python3
    @classmethod
    def apply_to_dask(cls, dask_array, version="latest"):
        import dask_awkward

        dask_array, _ = super().apply_to_dask(dask_array)
        schema_instance = cls(json.loads(dask_array.form.to_json()), version)

        meta = dask_awkward.typetracer_from_form(awkward.forms.from_dict(schema_instance.form))
        dask_array = dask_awkward.map_partitions(lambda x: x, dask_array, meta=meta)
        dask_array.layout.behavior = schema_instance.behavior

        return dask_array, schema_instance
```

When I try to then look for, for instance `events.Muon.pt` in decorated awkward array, it appears to be searching for the name "Muon" in the input data, rather than checking the form key of `events.Muon.pt` (which is what we actually want to check), leading to an error.

image

I'm bringing this up since if we get this case dealt with then we can really start debugging coffea with a dask backend.

douglasdavis:
One thing immediately comes to mind but it may not be the issue: perhaps the `AwkwardInputLayer._meta` attribute (the original `_meta` created by the `uproot.dask` call) is somehow missing the schema update step. Without messing with code I feel like this shouldn't be the case, I've been surprised by this stuff before!

I just tried to install `coffea@awkward2_dev` but I'm seeing this:
```
In [14]: test, _ = NanoAODSchema.apply_to_dask(test)
/Users/ddavis/software/repos/dask-awkward/src/dask_awkward/lib/core.py:1315: UserWarning: metadata could not be determined; a compute on the first partition will occur.
  warnings.warn(
---------------------------------------------------------------------------
AttributeError                            Traceback (most recent call last)
Cell In[14], line 1
----> 1 test, _ = NanoAODSchema.apply_to_dask(test)

File ~/software/repos/coffea/coffea/nanoevents/schemas/nanoaod.py:183, in NanoAODSchema.apply_to_dask(cls, dask_array, version)
    180 dask_array, _ = super().apply_to_dask(dask_array)
    181 schema_instance = cls(json.loads(dask_array.form.to_json()), version)
--> 183 meta = dask_awkward.typetracer_from_form(schema_instance._form)
    184 dask_array = dask_awkward.map_partitions(lambda x: x, dask_array, meta=meta)
    185 dask_array.layout.behavior = schema_instance.behavior

File ~/software/repos/dask-awkward/src/dask_awkward/lib/core.py:1751, in typetracer_from_form(form)
   1733 def typetracer_from_form(form: Form) -> ak.Array:
   1734     """Create a typetracer Array from an awkward form.
   1735 
   1736     This functions uses `form` along with :func:`awkward.from_buffers`
   (...)
   1749 
   1750     """
-> 1751     layout = form.length_zero_array(highlevel=False)
   1752     return ak.Array(layout.to_typetracer(forget_length=True))

AttributeError: 'dict' object has no attribute 'length_zero_array'
```
Do you have some local changes that you can share that would help me reproduce this?

lgray:
Oh yeah some stuff changed and I had to patch. Let me push those.

lgray:
@douglasdavis take the current awkward2_dev branch!

douglasdavis:
Thanks!

In `NanoAODSchema.apply_to_dask` it looks like the metadata is getting updated properly (which is why `test.Muon` works, dask-awkward is using the metadata to determine that there is a `Muon` field and it happily creates a new getitem step in the task graph), but I think the task graph that is getting stitched together is (at compute time) instantiating arrays that don't have the new schema/behavior applied.

On this line:
```python
dask_array = dask_awkward.map_partitions(lambda x: x, dask_array, meta=meta)
```
Doesn't this function need to modify the schema/behaviors of `dask_array` (not just return the same array)?

lgray:
@douglasdavis 

Thanks for the pointer, that was a good chunk of things!
So I've finally run into a conceptual problem here and can explain the impasse.

You can find all this as-is in the awkward2_dev branch for playing around with.

The problem can largely be summarized by this picture:
image

The main issue with using a `map_partitions` based approach to applying complex NanoEvents is that we only want to touch the data asked for by `test.Muon.pt`, but we only know which data that is from running the graph backwards which a mapped function cannot have any idea about. We cannot touch all data in in the NanoEvents remapping because that would be deeply inefficient, we only want to read the data we ask for.

This, to me, implies some sort of IO-ish layer (like the `dak.from_buffers` idea or something in uproot.dask) that can take a form with `form_keys` defined and manage the column projections in the way we want.

lgray:
Transcribing this from slack -> I'll try to (in parallel to a `dak.from_buffers` implementation) see if I can patch together what I need embellishing upon the uproot awkward IO layer. If I can get that to work we can certainly back-solve into something less profoundly clunky.

lgray:
Some odd behavior in `dak.necessary_columns` and zips perhaps?

This is using this branch + awkward@main + coffea 0.8 integration branch.

Symptoms as follows for the following outputs calculation:
image

for the two histograms things are fine:
image

For the skim also ok:
image

However for the cutflow counters, because I am only using `muons` `*_pt` returns empty necessary columns when it needs at least one of them (to get the offsets/length) but `*_mass` is fine because it comes from a 4-momentum sum that touches data to create the object:
image
image

martindurant:
@douglasdavis , this sounds like what we talked about. The labeling we currently do only applies to values/leaf buffers, not the index/offsets that you need to get a count. We should label those too, but when they come up in the report without any of their children, that's when we get to make a choice of which child column to actually load.

douglasdavis:
> @douglasdavis , this sounds like what we talked about. The labeling we currently do only applies to values/leaf buffers, not the index/offsets that you need to get a count. 

Yep, agreed!

> We should label those too, but when they come up in the report without any of their children, that's when we get to make a choice of which child column to actually load.

I'll see if I can come up with a working example

martindurant:
@jpivarski we have been experimenting with trying to get the report to handle the case when only offsets get touched in the typetacer report. Specific case is ak.num as discovered by @lgray above:
```python
dak.num(array.field, axis=1)
```
In that case, no field is reported as being necessary because the data is never touched, only the offsets.

With a little digging, we find that the root cause of this seems to be in our labelling a form object and passing this to typetracer_with_report. But that form does not have objects for the offsets, but the type ("i32" usually): it is only the generated layout that has buffer-like objects at those points in the structure. 

Since _attach_report recurses both the form and layout to apply labels from the former to the latter, we did experiment with recursing through a layout instead, but didn't make progress since it's not clear from the outside when an attribute lookup returns a mutable reference versus making a new object.

Summary:
when doing `ak.num(typetracer_array.field, axis=1)`, we would like *something* to show up in the report indicating that the list "array.field" is needed (but not the contents).

agoose77:
@martindurant let me know if I'm missing something, but shouldn't the `ListOffsetForm` form key be used for its offsets? The list _data_, e.g. a `NumpyArray` should have its own form:
```python
import copy
import awkward as ak
from awkward._nplikes.typetracer import TypeTracerReport, TypeTracer, _attach_report


layout = ak.to_layout([[1,2,3], [4,5]])
layout = layout.to_typetracer(forget_length=True)

form = copy.deepcopy(layout.form)
form.form_key = "offsets"
form.content.form_key = "data"

report = TypeTracerReport()
_attach_report(layout, form, report)

print(report.data_touched)
result = ak.num(layout)
print(report.data_touched)
```

In the above code, I only see `"offsets"` in `report.data_touched`. This is in `main`, for context.

jpivarski:
@martindurant, you're right that `ak.num(array)` where `array` is `ListOffsetArray(NumpyArray)` would not touch the `NumpyArray`, and @agoose77 has the right remedy: assigning a `form_key` to the `ListOffsetArray`.

For a Parquet source (not v2), the `ListOffsetArrays` above a `NumpyArray` for column `x` should all be labeled with the same `form_key` as `x` because if you touch any offsets, you're going to need to read the whole column. I know that Parquet v2 has a way around that, but I don't think pyarrow's Parquet reader does.

Then what about `ListOffsetArray(RecordArray(x: NumpyArray, y: NumpyArray))`? The `ListOffsetArray`'s `form_key` should be set to indicate that you need to read something to find the number of records in each list. If this is a Parquet file, you'd need either `x` or `y` to do that. In Awkward v1's lazy arrays, the "smallest" one, usually a column of booleans, was picked.

This also applies to option-type nodes: a function like `ak.is_none` would touch the option-type node (`BitMaskedArray`, `ByteMaskedArray`, `IndexedOptionArray`, or `UnmaskedArray`) but not its contents, though answering a question about which entries are missing would require some data to be loaded (possibly following the same model as the one for lists, above).

`RecordArrays` don't need any `form_keys` because they're metadata-only—all of the questions you'd ask of them can be answered without reading data.

`UnionArrays` need data in principle, but (I think) any operation on a `UnionArray` would also touch all of its constituents, so you'd get these for free.

`IndexedArray` would only come directly from the source if it's a dictionary encoding, and there's an explicit column associated with that.

`EmptyArray` doesn't need any data. Nor do `RegularArrays`.

So in a Form that will populate a report,

  * `NumpyForm`
  * `ListOffsetForm`, `ListForm`
  * `BitMaskedForm`, `ByteMaskedForm`, `IndexedOptionForm` (maybe not `UnmaskedArray`, since it has no data)
  * `IndexedForm` from dictionary encoding

should get `form_keys` that point back to identifiers in the source like column names, and you may need to make some choices about which column is going to provide information to a list-type or option-type when more than one could do it.

src/dask_awkward/lib/optimize.py
```diff
@@ -86,177 +122,191 @@ def _projectable_io_layer_names(dsk: HighLevelGraph) -> list[str]:
...
+                    cnew = c[:-9].rstrip(".")
+                    if cnew not in select:
+                        select.append(f"{cnew}.*")
+                else:
+                    select.append(c)
+        kv[name] = select
+    return kv
+
+
+def _prune_wildcards(columns: list[str], meta: AwkwardArray) -> list[str]:
```
douglasdavis, line 253: @martindurant would be happy to get your take on this function.

It currently has one pitfall (that I know of!)

given a dataset with has parquet columns `foo.bar.x` and `foo.bar.y` (where `foo.bar` is a record array with fields `x` and `y`):

```python
dak.num(foo.bar.y, axis=1).compute()
```
will fail because `foo.bar.*` is what will be found as necessary, but then this function will result in `foo.bar.x` as the column to read (which is technically fine and should result in a correct computed array), but the dask graph will expect to get to `foo.bar.y`.

martindurant: > the dask graph will expect to get to foo.bar.y

What do you mean by this?

douglasdavis: This time it's the fully computed graph (not just the optimization typetracer graph) that errors. The optimization will determine that we only need _any_ field from `foo.bar` (which is technically true for the correct result), but the task graph has that additional getitem layer for retrieving `"y"`. 

I'm wondering if there's another piece of the graph we need to touch during the optimization to avoid dropping the field that was actually explicitly asked for in the original graph

Here's a minimal reproducer with the muons dataset, where `dak.num(ds.muons.eta, axis=1)` can be simplified to `dak.num(ds.muons, axis=1)` so we end up going down the path of just grabbing the first field of `ds.muons` (in this case it happens to be `pt`):

```python
>>> dak.num(ds.muons.eta, axis=1)
dask.awkward<num, npartitions=50>
>>> a = dak.num(ds.muons.eta, axis=1)
>>> a.dask
HighLevelGraph with 4 layers.
<dask.highlevelgraph.HighLevelGraph object at 0x11e8afad0>
 0. read-parquet-1045eb08760cefeedd5f6194e5d1b18f
 1. muons-6a84d262864006fc9042dc8a4de45586
 2. eta-727d9bccb85d3046d4242f492b1a0c92
 3. num-76cb7a5f0a0c6ad8c1c7b3a45d97fa88

>>> a.compute()
  File "/Users/ddavis/software/repos/dask/dask/base.py", line 314, in compute
                ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/ddavis/software/repos/dask/dask/base.py", line 599, in compute
              ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/ddavis/software/repos/dask/dask/threaded.py", line 89, in get
    results = get_async(
              ^^^^^^^^^^
  File "/Users/ddavis/software/repos/dask/dask/local.py", line 511, in get_async
    raise_exception(exc, tb)
  File "/Users/ddavis/software/repos/dask/dask/local.py", line 319, in reraise
    raise exc
  File "/Users/ddavis/software/repos/dask/dask/local.py", line 224, in execute_task
             ^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/ddavis/software/repos/dask/dask/core.py", line 119, in _execute_task
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/ddavis/software/repos/dask/dask/optimization.py", line 990, in __call__
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/ddavis/software/repos/dask/dask/core.py", line 149, in get
             ^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/ddavis/software/repos/dask/dask/core.py", line 119, in _execute_task
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/ddavis/software/repos/dask/dask/core.py", line 119, in <genexpr>
                  ^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/ddavis/software/repos/dask/dask/core.py", line 113, in _execute_task
    return [_execute_task(a, cache) for a in arg]
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/ddavis/software/repos/dask/dask/core.py", line 113, in <listcomp>
    return [_execute_task(a, cache) for a in arg]
            ^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/ddavis/software/repos/dask/dask/core.py", line 119, in _execute_task
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/ddavis/.pyenv/versions/dev/lib/python3.11/site-packages/awkward/highlevel.py", line 951, in __getitem__
    out = self._layout[where]
          ~~~~~~~~~~~~^^^^^^^
  File "/Users/ddavis/.pyenv/versions/dev/lib/python3.11/site-packages/awkward/contents/content.py", line 514, in __getitem__
    return self._getitem(where)
           ^^^^^^^^^^^^^^^^^^^^
  File "/Users/ddavis/.pyenv/versions/dev/lib/python3.11/site-packages/awkward/contents/content.py", line 527, in _getitem
    return self._getitem_field(where)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/ddavis/.pyenv/versions/dev/lib/python3.11/site-packages/awkward/contents/listoffsetarray.py", line 260, in _getitem_field
    self._content._getitem_field(where, only_fields),
    ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/ddavis/.pyenv/versions/dev/lib/python3.11/site-packages/awkward/contents/unmaskedarray.py", line 197, in _getitem_field
    self._content._getitem_field(where, only_fields), parameters=None
    ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/ddavis/.pyenv/versions/dev/lib/python3.11/site-packages/awkward/contents/recordarray.py", line 367, in _getitem_field
    return self.content(where)
           ^^^^^^^^^^^^^^^^^^^
  File "/Users/ddavis/.pyenv/versions/dev/lib/python3.11/site-packages/awkward/contents/recordarray.py", line 316, in content
    out = self.form_cls.content(self, index_or_field)
          ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/ddavis/.pyenv/versions/dev/lib/python3.11/site-packages/awkward/forms/recordform.py", line 153, in content
    index = self.field_to_index(index_or_field)
            ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/ddavis/.pyenv/versions/dev/lib/python3.11/site-packages/awkward/contents/recordarray.py", line 310, in field_to_index
    return self.form_cls.field_to_index(self, field)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/ddavis/.pyenv/versions/dev/lib/python3.11/site-packages/awkward/forms/recordform.py", line 130, in field_to_index
    raise ak._errors.wrap_error(
awkward._errors.FieldNotFoundError: while attempting to slice

    <Array [[{pt: 43.1}, {...}, ..., {...}], ...] type='5994 * var * ?{pt: ...'>

with

    'eta'

Error details: no field 'eta' in record with 1 fields
```

martindurant: I see; that's ... annoying. So the metadata operation getting the field needs that field to exist even though we never touch its data. We could probably spot this pattern and apply an optimization transform, but that sounds like too much increased complexity.

martindurant: @jpivarski maybe you have an opinion here?

agoose77: I'm not hugely following this PR just now (lots of things in parallel), but as I see it we have two options:
1. Use the knowledge that the data aren't important, and use something like a NumPy strides trick to generate dummy data. This wouldn't work for for the JAX backend, but that's a fair tradeoff (JAX is limiting in places like this).
2. Don't apply this kind of optimisation. Not sure how feasible this is

martindurant: I'm not sure we can do the first. We need parquet or JSON to actually do some loading to get the offsets, but have no way to **only** get the offsets. We could modify the call in the real graph as to what we then do with it (e.g., override getitem to get what we loaded, whether it's the right field or not). Sounds a bit painful.

The column optimization should be pretty important. Yes, in this case we are able to load only the record, but it would be the full record to avoid the problem. That record could be all of the data actually. We might choose to make life easy and kick this problem to the future by selecting the whole record.

It is interesting that
```
dak.num(arr.record.field, axis=1) === dak.num(arr.record, axis=1)
```
but users probably don't realise this. The latter form means we can freely choose which column to load. The former might be the right one if the user knows which column should be the easiest to load.

lgray: Some analysis objects can have quite a lot of columns, selecting the whole record only to count it would be extremely expensive in terms of bandwidth off disk (especially for jets, for instance).

douglasdavis: I think right now our best bet is to perform the optimization as-is and be prepared to field issues/questions based on the error message. We can also try to come up with a decent piece of docs explaining this potential obstacle

lgray:
using the tip of this branch + awkward main I'm seeing:
image

douglasdavis:
The indexing definitely needs some safeguards in this code. Would you be able to give a reproducer?

lgray:
I'll get you a reproducer in a bit!

lgray:
Here's a notebook with the reproducer.
[repro_doug.ipynb.zip](https://github.com/dask-contrib/dask-awkward/files/10677367/repro_doug.ipynb.zip)

lgray:
The nastier things are issues in awkward@main with typetracers of regular arrays (so far)

douglasdavis:
I'm not 100% https://github.com/dask-contrib/dask-awkward/pull/148/commits/d62ec672ad2c8d01572c42add556fd3e24030555 does what we'd like to in this case, but it appears to work. For some reason it looks like all the branches in the ROOT file are getting the wildcard syntax, but it's impossible for the ROOT file to have nested fields like a parquet file (right?) so anyway let me know how this looks to you

lgray:
ROOT files can certainly have highly nested structures, I'm not sure if uproot can provide them in a pleasantly columnar fashion off the top of my head @jpivarski ?

jpivarski:
ROOT can definitely have nested fields like a Parquet file, though its representation of them switches from columnar to record-oriented and it gets more complicated to actually read them. But with Uproot 5, AwkwardForth makes many of those cases not-terribly-slow.

This:

https://github.com/scikit-hep/uproot5/blob/main/tests/test_0637-setup-tests-for-AwkwardForth.py

is a suite of ROOT file reading tests that are all sufficiently complicated that AwkwardForth gets invoked (i.e. they're not columnar).

Is it nested record fields that you're wondering about or nested lists? We can provide a test example for just about anything.

jpivarski:
ROOT tries to split objects by their attributes _if_ the `splitLevel` is set to some high value; `splitLevel=99` is the default. I know that the second variable-sized thing always stops ROOT's splitting, but if you had a fixed-size struct inside a fixed-size struct, I don't know whether `splitLevel=99` would put that in separate TBranches or not.

But anyway, anything unsplit means that you have to read the whole thing, so if you have wildcard syntax that can't distinguish between `outer.mid.inner1` and `outer.mid.inner2`, it doesn't matter because you'll have to read the whole `outer`, assuming it's in some unsplit level.

This is not relevant for NanoAOD files. The last time I checked, they're all 100% split.

(AwkwardForth only speeds up the reading of unsplit TBranches. But if you're ever looking for examples of unsplit TBranches, the AwkwardForth tests are a good place to start. In particular, [uproot-small-evnt-tree-nosplit.root](https://github.com/scikit-hep/uproot5/blob/55d1b716446abb24e458f92ed1617121941391db/tests/test_0637-setup-tests-for-AwkwardForth.py#L662-L673) has a `uproot-small-evnt-tree-split.root` counterpart for comparison.)

douglasdavis:
OK I now see why these fields are getting wildcarded. It's because if a form satisfies `form.is_list` we add the wildcard and any `std::vector<T>` branch is going to satisfy that (like `Electron_pt`); for primitive types `T` in these cases it makes sense for no subfields to exist, so it's safe to skip those with the if statement I added in https://github.com/dask-contrib/dask-awkward/pull/148/commits/d62ec672ad2c8d01572c42add556fd3e24030555.

What I still *do not* understand is why the typetracer is reporting these form_keys as touched

When we label the form to run the typetracer only graph we use this recursive function:

```python
def _label_form(form, start):
    if form.is_record:
        for field in form.fields:
            _label_form(form.content(field), f"{start}.{field}")
    elif form.is_numpy:
        form.form_key = start
    elif form.is_list:
        form.form_key = f"{start}.__list__"
        _label_form(form.content, start)
    else:
        _label_form(form.content, start)
```

Each form_key getting set in the `elif form.is_list` block is getting touched. We end up with a from_key like 

```
layer-name-123.Electron_pt.__list__
```

This appears as touched which creates this necessary column:

```
Electron_pt.*
```

But since there are no fields belonging to that particular column, we end up safely skipping it in @lgray's example that only asks for `Muon_{pt,eta,phi,mass}` fields.

douglasdavis:
The calculation that _does not_ work is `out["pt"]`, because it's histogrammed from dot access on the _new_ `muon` field (`muon.pt`), which is built from a call to `dak.zip` with more than just the `Muon_pt` "ingredient" (the zip expects to have `Muon_eta`, `Muon_phi`, and `Muon_mass`). The `out["pt"]` graph ends up only discovering `Muon_pt` as a touched buffer, and the compute fails because the `dak.zip` layer expects the other 3 fields. We need to some how accommodate for this.

douglasdavis:
Alright latest commit (https://github.com/dask-contrib/dask-awkward/pull/148/commits/caaafed48b8abd739030a02ac4ce52350abcd0b3) fixes the `out["pt"]` scenario.

I've added a new layer annotation which directs the optimization to touch all data it finds in a function call; now all arrays passed to `dak.zip` are guaranteed to get touched. The feature is general and any `map_partitions` call can utilize the annotation with

```python
dak.map_partitions(fn, array, annotations={"ak_touch_all": True})
```

Right now `dak.zip` is the only function in the API that uses it.

lgray:
@douglasdavis so now out["pt"] alone, with this annotation, would cause reading of all arrays stuck into the zip?

lgray:
Indeed it does - I would strongly advise against this as the final behavior we want (but I get why things go this way). Reading all columns for a single column of data's worth of info will incur a really heavy penalty.

I'm (finally) working on a nanoevents aware input layer, which I think can avoid initial zipping and thus this problem. However, this may catch users unaware.

douglasdavis:
> Reading all columns for a single column of data's worth of info will incur a really heavy penalty.

Yea I completely understand this, but for dask-awkward to be able to use the upstream implementation of zip we need to ensure the components _passed_ to zip are available during compute, so we can't optimize away downstream unused fields. I suppose we can build a new implementation that is "smarter", but for now I think that would be beyond the scope of this PR.

> I'm (finally) working on a nanoevents aware input layer, which I think can avoid initial zipping and thus this problem. However, this may catch users unaware.

It sounds like this would completely bypass the issue of overzelous-reading-from-disk-because-of-a-zip?

lgray:
The reasoning for going with touch_all totally makes sense for this PR, I guess what I'm saying is that physicists tend to not work that way (a warning may help but we're stubborn folk). Engineering smart common behavior is a good way to go in the long run. 100% agree its for another PR though.

I'm hoping the implementation can avoid it! If it does it would significantly reduce the priority optimizing through zips (for me).

douglasdavis:
While we still have some warnings associated with failed optimizations, I think this may be the first time the PR is green in CI 😝

lgray:
let me give this a try with my notebook of various tortures.

douglasdavis:
(I've updated CI to use `awkward@main` to get things green)

douglasdavis:
@agoose77 here's a notebook that surfaces the every-form-key-labeled-with-duner-list-gets-touched occurrence (you'll need to clone the coffea repo somewhere and use `python<3.11`)
[form-keys-all-get-touched.ipynb.zip](https://github.com/dask-contrib/dask-awkward/files/10701110/form-keys-all-get-touched.ipynb.zip)

lgray:
Oh interesting the optimized graph output changed..
I used to get one fused zip node for each file from the datasets but now I get:
image

and the zip didn't get optimized into a single node (which seems incorrect since it doesn't consolidate data access to a single node)?

lgray:
Ok awkward@main is still waiting for 2227 for my notebook to function.

lgray:
Otherwise looks ok.

agoose77:
@lgray those PRs have now *all* merged, so you should be able to test again.

lgray:
@douglasdavis So I've run into an actual problem with the current version. With the column optimization as it I *cannot* optimize through zips. This is especially problematic for big graphs like the jet energy scale. Because of how the zip is treated none of the unneed calculations are pruned leading to some really significant overhead in processing. You can see this manifest in the task graph above where the data-access bits are not fused into a single node:
image

Any way we can fix this? It's definitely not behavior that we want.

douglasdavis:
I have a hunch. I think the new annotation is overwriting existing annotations. Checking now

douglasdavis:
Ah it's not that annotations are getting overwritten (there were none to begin with), it's the _presence_ of the new annotation that causes the blockwise optimization algorithm to skip the layer: 

https://github.com/dask/dask/blob/6df830a155c241191d478809e2e2b469fa78d913/dask/blockwise.py#L1331-L1336

douglasdavis:
So upstream dask assumes a finite set of possible annotations; f24b092 moves away from annotations and just uses a class attribute. Since the optimization is always run on the client this should be OK w.r.t. dask pickling layers and sending them over the wire to the scheduler.

lgray:
Yep - works as before now! Thanks!

lgray:
So sorry in advance - this needs a few PRs to go in (and I have to wrap up some work in coffea0 before it's easy to repeat. However, I did find some interesting outcomes in the column optimization for things that aren't zips, but are more complex things we tend to find in analysis code.

For instance with coffea nanoevents it's not so uncommon you want to match objects together get the nearest one, and maybe ask for the pt or some other single kinematic quantity. We have some convenience functions that wrap that up as follows, which works fine if daskified:
```
nearest_pt = x.Muon.nearest(x.Jet).pt
```
However when looking at the necessary columns. It's interesting that the column opt picks up the needed columns for Muons correctly (calculating delta_r only needs Muon.eta and Muon.phi), but the *entire* Jet object is requested from disk which is not necessary since we likewise only need the Jet's eta, phi, and pt in this case.
```
>>> dak.necessary_columns(x.Muon.nearest(x.Jet).pt)
{'from-uproot-efbdf44a84a3be6d60f91c4406a73ce9': ['Jet.btagCMVA', 'Jet.electronIdxG', 'Jet.eta', 'Jet.electronIdx2', 'Jet.cleanmask', 'Jet.electronIdx2G', 'Jet.muonSubtrFactor', 'Jet.muonIdxG', 'Jet.muonIdx2G', 'Jet.muEF', 'Jet.btagDeepB', 'Jet.btagCSVV2', 'Jet.muonIdx2', 'Jet.btagDeepFlavB', 'Jet.jercCHPUF', 'Jet.btagDeepFlavC', 'Jet.nElectrons', 'Jet.muonIdx1', 'Jet.pt', 'Jet.chEmEF', 'Jet.electronIdx1', 'Jet.neHEF', 'Jet.jetId', 'Jet.phi', 'Jet.nMuons', 'Muon.eta', 'Jet.rawFactor', 'Jet.muonIdx1G', 'Jet.btagDeepC', 'Jet.bRegRes', 'Jet.bRegCorr', 'Jet.electronIdx1G', 'Jet.chHEF', 'Muon.phi', 'Jet.neEmEF', 'Jet.mass', 'Jet.nConstituents', 'Jet.jercCHF', 'Jet.area', 'Jet.qgl', 'Jet.puId']}
```

I'll follow up with instructions to repeat once everything is in and the recipe is easily shareable.

lgray:
Found a bug that will be run into regularly, here's a repro:
```python3
import uproot
import dask_awkward as dak

print(dak.necessary_columns(dak.broadcast_arrays(x.fixedGridRhoFastjetAll, x.Jet_pt)))
```
prints:
```
{'from-uproot-d77a4895f123e53606d66a8430b53410': ['Jet_pt']}
```

As opposed to needing the data from both columns!

PR #150, opened 2023-01-23
Remove need for explicit calc in dak.with_name

lgray:
Usual meta patch-over with `length_zero_array`

@agoose77 FYI

### DIFF
--- a/src/dask_awkward/lib/structure.py
+++ b/src/dask_awkward/lib/structure.py
@@ -924,11 +924,23 @@ def with_name(
 ) -> Array:
     if not highlevel:
         raise ValueError("Only highlevel=True is supported")
+
+    #  TODO: remove once fixed in awkward
+    meta = typetracer_from_form(
+        ak.with_name(
+            array._meta.layout.form.length_zero_array(),
+            name,
+            highlevel=highlevel,
+            behavior=behavior,
+        ).layout.form
+    )
+
     return map_partitions(
         _WithNameFn(name=name, behavior=behavior),
         array,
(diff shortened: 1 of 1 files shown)


lgray:
Well, that's even weirder. I'm not sure why propagation of behaviors is being lost when I do this?

@agoose77 any idea?

src/dask_awkward/lib/structure.py
```diff
@@ -924,11 +924,23 @@ def with_name(
 ) -> Array:
     if not highlevel:
         raise ValueError("Only highlevel=True is supported")
+
+    #  TODO: remove once fixed in awkward
+    meta = typetracer_from_form(
+        ak.with_name(
+            array._meta.layout.form.length_zero_array(),
```
agoose77, line 931: `array._meta` has a `.behavior` attribute that contains the behavior dict. By unwrapping the meta layout, we never read the behaviors back from this object.

I would assume that we're doing this in other places too. The result of `length_zero_array` needs to be re-wrapped with the original behaviours, perhaps like `ak.Array(array._meta.layout.form.length_zero_array(), behavior=array._meta.behavior)`

lgray: Has to be tacked on to the result of `typetracer_from_form`, but that effectively did the trick. Have a look!

douglasdavis:
This looks good to me. @lgray I'll merge if there's nothing else you'd like to add

lgray:
Please go ahead - just taking care of warnings I see pop up!

PR #152, opened 2023-01-24
UnknownScalar removal

douglasdavis:
This is a followup to scikit-hep/awkward#2154

### DIFF
--- a/src/dask_awkward/lib/core.py
+++ b/src/dask_awkward/lib/core.py
@@ -12,7 +12,12 @@ from typing import TYPE_CHECKING, Any, TypeVar
 import awkward as ak
 import dask.config
 import numpy as np
-from awkward._nplikes.typetracer import MaybeNone, OneOf, TypeTracerArray
+from awkward._nplikes.typetracer import (
+    MaybeNone,
+    OneOf,
+    TypeTracerArray,
+    is_unknown_scalar,
+)
 from awkward.highlevel import _dir_pattern
 from dask.base import DaskMethodsMixin, dont_optimize, is_dask_collection, tokenize
 from dask.blockwise import BlockwiseDep
@@ -157,9 +162,9 @@ class Scalar(DaskMethodsMixin):
         return (self._name, 0)
 
     def _check_meta(self, m: Any) -> Any | None:
-        if not isinstance(m, (TypeTracerArray, MaybeNone, OneOf)):
-            raise TypeError(f"meta must be a typetracer object, not a {type(m)}")
-        return m
+        if isinstance(m, (MaybeNone, OneOf)) or is_unknown_scalar(m):
+            return m
+        raise TypeError(f"meta must be a typetracer object, not a {type(m)}")
 
     @property
     def dtype(self) -> np.dtype | None:
--- a/src/dask_awkward/lib/core.py
+++ b/src/dask_awkward/lib/core.py
@@ -12,7 +12,7 @@ from typing import TYPE_CHECKING, Any, TypeVar
 import awkward as ak
 import dask.config
 import numpy as np
-from awkward._typetracer import MaybeNone, OneOf, UnknownScalar
+from awkward._nplikes.typetracer import MaybeNone, OneOf, TypeTracerArray
 from awkward.highlevel import _dir_pattern
 from dask.base import DaskMethodsMixin, dont_optimize, is_dask_collection, tokenize
 from dask.blockwise import BlockwiseDep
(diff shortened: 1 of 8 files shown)


src/dask_awkward/lib/core.py
```diff
@@ -157,7 +157,7 @@ def key(self) -> Hashable:
         return (self._name, 0)
 
     def _check_meta(self, m: Any) -> Any | None:
-        if not isinstance(m, (MaybeNone, UnknownScalar, OneOf)):
+        if not isinstance(m, (TypeTracerArray, MaybeNone, OneOf)):
```
agoose77, line 160: I'd suggest replacing `isinstance(x, UnknownScalar)` with `awkward._nplikes.typetracer.is_unknown_scalar`, because that way you'll not pick up ND typetracer arrays here. Note - unlike much of awkward, this needs to be explicitly imported

src/dask_awkward/lib/core.py
```diff
@@ -1500,7 +1502,7 @@ def is_typetracer(obj: Any) -> bool:
         if not backend.nplike.known_shape and not backend.nplike.known_data:
             return True
     # scalar-like typetracer
-    elif isinstance(obj, (UnknownScalar, MaybeNone, OneOf)):
+    elif isinstance(obj, (TypeTracerArray, MaybeNone, OneOf)):
```
agoose77, line 1505: Same here.

src/dask_awkward/lib/core.py
douglasdavis: Looks like we get `awkward._nplikes.typetracer.MaybeNone` instances popping up; should `is_unknown_scalar` check for those?

agoose77: I have thought about this; it depends upon what context. I think we might want another function `is_unknown` or something to handle this case. Do you have any example tracebacks? I assume they're not coming from the changes in awkward/main?

douglasdavis: Or should we make this function
```
if is_unknown_scalar(m) or isinstance(m, (MaybeNone, OneOf)):
    # good to go
else:
    # bad

Comment (douglasdavis) [reply src/dask_awkward/lib/core.py:160]: Here's a traceback
```python
>>> dak.sum(ds.muons.pt, axis=None)
  File "/Users/ddavis/software/repos/dask-awkward/src/dask_awkward/lib/reducers.py", line 405, in sum
    return total_reduction_to_scalar(
           ^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/ddavis/software/repos/dask-awkward/src/dask_awkward/lib/core.py", line 1393, in total_reduction_to_scalar
    return new_scalar_object(graph, name_agg, meta=meta)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/ddavis/software/repos/dask-awkward/src/dask_awkward/lib/core.py", line 265, in new_scalar_object
    return Scalar(dsk, name, meta, known_value=None)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/ddavis/software/repos/dask-awkward/src/dask_awkward/lib/core.py", line 107, in __init__
    self._meta: Any = self._check_meta(meta)
                      ^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/ddavis/software/repos/dask-awkward/src/dask_awkward/lib/core.py", line 162, in _check_meta
    raise TypeError(f"meta must be a typetracer object, not a {type(m)}")
TypeError: meta must be a typetracer object, not a <class 'awkward._nplikes.typetracer.MaybeNone'>
```

douglasdavis: After looking at our implementation of `total_reduction_to_scalar` I think this may be an issue on our end!

douglasdavis: My awkward branch is `agoose77/refactor-drop-unknown-scalar`. `MaybeNone` is coming from something like this:

```python
>>> import awkward as ak
>>> ds.muons.pt._meta
<Array-typetracer [...] type='?? * var * float64'>
>>> ak.max(ds.muons.pt._meta, axis=None)
MaybeNone(TypeTracerArray(dtype('float64'), shape=()))
```

agoose77: Thanks Doug. We've now merged this into main, although I'm going to rework the unknown-length concept to use None in the near future. We do expect `MaybeNone`, and in this case it would make sense for `axis=None`. We'll all need to figure out how we want to handle this in dask-awkward.

PR #162, opened 2023-02-15
Adjust `_meta` determination during graph construction (new config parameter, length zero array fallback)

douglasdavis:
This PR changes the way metadata is handled at graph creation time.

1. We've added a configuration parameter `awkward.raise-failed-meta` that, when `True` (default is `False`), will raise an exception during graph creation time if the process of determining a new metadata (typetracer array) fails. For example, in `map_partitions(fn, array)` we always call `fn(array._meta)` to determine the "new" meta on the graph. By default, if this fails we will compute the first partition and issue a warning. The new configuration option will hard fail instead of warning.
2. We've added an additional code path for determining the new meta via a call on a length zero array _instead_ of a high level typetracer array. (the default behavior is to always try on a high level typetracer array). If the new config parameter from (1) is set to `True`, we will never reach this code; but, if we don't want to hard fail after `fn(array._meta)` fails, then we go to the length zero array workaround; which is `fn(array._meta.layout.form.length_zero_array())`. If the later call fails as well, then we issue a warning and fall back to computing the first partition.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -27,7 +27,7 @@ classifiers = [
     "Topic :: Scientific/Engineering",
 ]
 dependencies = [
-    "awkward >=2.0.7",
+    "awkward >=2.0.8",
     "dask >=2022.02.1",
 ]
 dynamic = ["version"]
--- a/src/dask_awkward/awkward.yaml
+++ b/src/dask_awkward/awkward.yaml
@@ -1,5 +1,9 @@
 awkward:
 
+  # This option causes graph creation to fail if the metadata of a
+  # dask-awkward Array collection cannot be determined.
+  raise-failed-meta: False
+
   # This option is for cases where new array collections are created
   # with unknown metadata. The default setting is to compute the first
   # partition of the collection so that we can define the metadata. If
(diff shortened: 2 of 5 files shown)


douglasdavis:
cc @agoose77 

This PR is now in main and it introduces the new config parameter `awkward.raise-failed-meta`. By default it's off, but it can easily be turned on with via the environment variable method of dask configuration:

```
$ DASK_AWKWARD__RAISE_FAILED_META=1 pytest
```

we currently have only 3 failed tests with that option set.

The other configuration parameter that we can play with (from https://github.com/dask-contrib/dask-awkward/pull/148) is `awkward.optimization.on-fail`. When set to `raise` we see 8 failures (during the optimization step when running the whole graph on typetracer objects only)

```
$ DASK_AWKWARD__OPTIMIZATION__ON_FAIL=raise pytest
```

If we run both

```
$ DASK_AWKWARD__RAISE_FAILED_META=1 DASK_AWKWARD__OPTIMIZATION__ON_FAIL=raise pytest
```

we get 10 failures, so there is some overlap (to be expected since this is typetracer-land)

PR #173, opened 2023-03-01
Fix layer order assumption in optimization

### LINKED ISSUE #172: Uproot test fails with latest dask_awkward
https://github.com/scikit-hep/uproot5/blob/main/tests/test_0791-protect-uproot-project_columns-from-dask-node-names.py now fails, I've extracted it for a repro:

```
pip install uproot scikit-hep-testdata dask-awkward
```

```python3
import uproot
import skhep_testdata

da = uproot.dask(skhep_testdata.data_path("uproot-issue-791.root") + ":tree")
assert da[da.int_branch < 0].compute().tolist() == [
    {
        "int_branch": -1,
        "long_branch": -2,
        "float_branch": 3.299999952316284,
        "double_branch": 4.4,
        "bool_branch": True,
    }
]
```
The assertion will fail, only the `int_branch` will be present in the output array.

douglasdavis:
fixes #172

the optimization always touches all fields in an array if it's the last array in a task graph. we were making an incorrect assumption about the output layer always being the last in the list of layer dictionary values. this PR makes use of the HLG topologically sorted layers method

### DIFF
--- a/src/dask_awkward/lib/optimize.py
+++ b/src/dask_awkward/lib/optimize.py
@@ -229,7 +229,7 @@ def _get_column_reports(dsk: HighLevelGraph) -> dict[str, Any]:
         layers[name] = _touch_and_call(layers[name])
 
     hlg = HighLevelGraph(layers, deps)
-    outlayer = list(hlg.layers.values())[-1]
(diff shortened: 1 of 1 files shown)


douglasdavis:
windows failure is a git issue 🤷

PR #175, opened 2023-03-01
Protect against empty project_columns argument

### LINKED ISSUE #174: Possible bug: tests/test_0652_dask-for-awkward.py in uproot tests fails on dask_awkward main
Using dask_awkward main branch:
```python3
test_path = (
    skhep_testdata.data_path("uproot-sample-6.08.04-uncompressed.root") + ":sample"
)
ttree = uproot.open(test_path)
    
ak_array = ttree.arrays()
dak_array = uproot.dask(test_path, library="ak")

assert len(ak_array) == len(dak_array)
```
results in:
```
AssertionError: assert 30 == 0
+  where 30 = len(<Array [{n: 0, b: True, ab: [...], ...}, ...] type='30 * {n: int32, b: bool...'>)
+  and   0 = len(dask.awkward<from-uproot, npartitions=1>)
```

opening this here despite it possibly being in a deeper library @agoose77 @jpivarski

douglasdavis:
fix #174

### DIFF
--- a/src/dask_awkward/layers/layers.py
+++ b/src/dask_awkward/layers/layers.py
@@ -136,7 +136,6 @@ class AwkwardInputLayer(Blockwise):
 
     def project_columns(self, columns: list[str]) -> AwkwardInputLayer:
         if hasattr(self.io_func, "project_columns"):
-
             # TODO: make project_columns call sites never pass in an
             # empty list.
             if len(columns) == 0:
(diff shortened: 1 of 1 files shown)


douglasdavis:
this is _kind of_ a workaround. I'd like to never pass empty lists to project_columns, and for us to instead determine which of the known fields should be plucked for a step that doesn't actually require a specific field, but does need at least one.

PR #176, opened 2023-03-02
reorder column projection

### LINKED ISSUE #171: Field order is changed after `compute`
#148 seems to no longer preserve the order of fields after `compute` (the exact file here is <https://github.com/iris-hep/func_adl_uproot/blob/aed2e68c6643bfc955ba45742874fe303201a467/tests/scalars_tree_file.root>):

```python
>>> import uproot
>>> u = uproot.dask('scalars_tree_file.root')
>>> u.fields
['int_branch', 'long_branch', 'float_branch', 'double_branch', 'bool_branch']
>>> u.compute().fields
['bool_branch', 'double_branch', 'float_branch', 'int_branch', 'long_branch']
```

It looks like the branches get alphabetized. Is this expected? Is there a way to enforce that the order is preserved? ROOT files have a fixed branch ordering, and there's plenty of code that relies on this fact.

martindurant:
(needs test)

Fixes #171

### DIFF
--- a/src/dask_awkward/layers/layers.py
+++ b/src/dask_awkward/layers/layers.py
@@ -143,7 +143,9 @@ class AwkwardInputLayer(Blockwise):
 
             # requested columns to original order; adds on extra columns, which are probably
             # wildcard ones
-            columns = [c for c in self.columns if c in columns] + [c for c in columns if c not in self.columns]
+            columns = [c for c in self.columns if c in columns] + [
+                c for c in columns if c not in self.columns
+            ]
             io_func = self.io_func.project_columns(columns)
             return AwkwardInputLayer(
                 name=self.name,
--- a/src/dask_awkward/layers/layers.py
+++ b/src/dask_awkward/layers/layers.py
@@ -141,6 +141,9 @@ class AwkwardInputLayer(Blockwise):
             if len(columns) == 0:
                 columns = self._meta.fields[:1]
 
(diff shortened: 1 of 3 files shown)


src/dask_awkward/layers/layers.py
```diff
@@ -141,6 +141,11 @@ def project_columns(self, columns: list[str]) -> AwkwardInputLayer:
             if len(columns) == 0:
                 columns = self._meta.fields[:1]
 
+            # requested columns to original order; adds on extra columns, which are probably
+            # wildcard ones
+            columns = [c for c in self.columns if c in columns] + [
+                c for c in columns if c not in self.columns
+            ]
```
douglasdavis, line 148: `self.columns` can be `None` (probably usually is) so I think we need to go through grabbing columns from the form.
```suggestion
            # requested columns to original order; adds on extra columns, which are probably
            # wildcard ones
            form_columns = self._meta.layout.form.columns()
            original = [c for c in form_columns  if c in columns]
            new = [c for c in columns if c not in form_columns]
            columns = new + original
```

martindurant: aha, ok. Are you OK with adding in "new" columns like this? I think they'll arise with num() applied to parquet, which produces columns like "field.*". I had these after the originals, you have them before - this should not matter.

douglasdavis: Ah yea I see, probably best to make it `original + new`

PR #178, opened 2023-03-03
touch 'what' argument by hand in dak.with_field

lgray:
This pull request is meant to inform @agoose77 and @jpivarski that there's something to follow up in awkward-array.

With this patch the complete jetmet correction stack works without problems in dak.

### DIFF
--- a/src/dask_awkward/lib/structure.py
+++ b/src/dask_awkward/lib/structure.py
@@ -890,6 +890,8 @@ class _WithFieldFn:
         self.behavior = behavior
 
     def __call__(self, base: ak.Array, what: ak.Array) -> ak.Array:
+        if ak.backend(what) == "typetracer":
+            what.layout._touch_data(recursive=True)
         return ak.with_field(
             base,
             what,
--- a/src/dask_awkward/lib/structure.py
+++ b/src/dask_awkward/lib/structure.py
@@ -890,6 +890,7 @@ class _WithFieldFn:
         self.behavior = behavior
 
     def __call__(self, base: ak.Array, what: ak.Array) -> ak.Array:
+        # TODO: remove backend handling when touch is handled automatically
         if ak.backend(what) == "typetracer":
             what.layout._touch_data(recursive=True)
         return ak.with_field(


lgray:
This brings up perhaps a deeper question - if we `dak.with_field` and then don't use that field in a `.compute()` should we still read that field? I think the answer is yes, but it probably requires some significant changes to the column optimization.

I came to this problem since the column optimization was doing the right thing, but the `dak.with_field` operation was *always* adding the array from disk no matter what, so if it wasn't being used the actual `ak.with_field` call would fail since the data was not requested in the first place since it wasn't being used in the user-side operation.

agoose77:
The general rule with typetracer-based arrays is that any operation that requires data/shape touches data/shape, but the dask-awkward library should prune unused branches to ensure that only the minimal set of data/shape touches are actually performed. `ak.with_field` *should* touch the data of an array, because it performs broadcasting. However, the actual implementation of `with_field` doesn't contain any touching operations, so there may be some missing work here. I'm track this :)

lgray:
I suppose the major point of conversation here is how to handle `dak.with_field` columns that are added but not used at a given `.compute()`. I imagine this is rare but not infrequent enough to bury. They are almost always added with the intention of later use but occasionally you're digging into some random part of your analysis and only want one branch of an answer instead of the whole thing, and certainly you may have added some data you're not using.

lgray:
To make a thing clear - I would like this to function well in advance of CHEP so I can fulfill my talk's abstract. :-)

So - having `ak.with_field` touch the data since it does broadcasting would be the best solution in the time domain! ^_^

The finer points we can resolve in time.

src/dask_awkward/lib/structure.py
```diff
@@ -890,6 +890,8 @@ def __init__(
         self.behavior = behavior
 
     def __call__(self, base: ak.Array, what: ak.Array) -> ak.Array:
```
douglasdavis, line 892:
```suggestion
    def __call__(self, base: ak.Array, what: ak.Array) -> ak.Array:
        # TODO: remove backend handling when touch is handled automatically
```

PR #184, opened 2023-03-07
Keep optimization-dropped fields as a data-less component of Array

### LINKED ISSUE #180: Add data-less fields to necessary columns optimization
For things like `ak.zip` and `ak.with_field` to work alongside the new necessary columns optimization, we need to make sure that fields accessed by `zip` and `with_field` function calls that end up _not_ being necessary for some final compute still exist after the optimization (but without containing any data). we should be able to do this with the following steps 

- we have an original layout for an IO layer
- we calculate the columns we actually need, and so have a modified layout
- we should like to fill this out to the full original layout by attaching zero-data arrays, so that the result can be used in things like `zip` and `with_field`

douglasdavis:
Attempt to fix #180

We want to be able to use the necessary columns optimization with something like this:

```python
import dask_awkward as dak
import numpy as np
import uproot

y = dak.zip({"pt": x.Muon_pt, "eta": x.Muon_eta})

# add another field lazily (we will never actually use it)
z = dak.with_field(y, x.Muon_phi, "phi")

# function which uses pt and eta (but not phi)
def some_func(pt, eta):
    return pt * np.cosh(eta)

w = some_func(z.pt, z.eta)

# currently fails because there is a node in the graph that expects
# `Muon_phi` (but it gets dropped by the optimization as unused
# (correctly!))
w.compute()
```

where `Muon_phi` is indeed kept as a field, but contains no data in memory (because it is unnecessary for the final computed result)


douglasdavis:
@jpivarski @agoose77 - made some decent progress here. Something is _partially_ working but could use some input.

Current state (feel free to look through the whole diff of the PR but I'll try to write it out with some explanation):

1. After determining the necessary columns we now check them against the original set of columns to figure out what is "removed" (really just "not read from disk")
2. Now also pass this list of removed columns to the `project_columns` interface. This is so the callable class that is going to do the actual `ak.from_xyz` call at the dask node knows about the other fields that will not be read from disk.
3. When the column-projectable callable is actually called at the Dask node, we tack on the removed columns _after_ the main (existing) read of `ak.from_xyz`:
```python
def __call__(self, source):
    a = ak.from_parquet(source, columns=self.columns)
    # before this PR we'd return a now

    # this PR adds this logic mock_dataless are the columns we want to mock w/o data:
    for entry in self.mock_dataless or []:
        # split on "." so we get lists of strings for fields of fields
        entries = entry.split(".")
        if len(entries) == 1:
            a = ak.with_field(
                a,
                np.broadcast_to(np.int64(0), (len(a),)),
                entries[0],
            )
        else:
            ## This will error if the top level field is not already available
            a = ak.with_field(
                a,
                np.broadcast_to(np.int64(0), (len(a),)),
                entries,
            )

        return a
```
(You can wee for this PoC I'm just using a int64 0).

Two issues right now.
- We're looping over all of the "dataless" fields and creating a new ak.Array each time. This appears to be N^2 with the number of columns; so it looks like we'd benefit from some kind of bulk augmentation function.
- The way it's written above fails in the `else` block if we have some subfield structure (like `foo.bar.baz` getting turned into `["foo", "bar", "baz"]`) and the first field is not already present in the array.

Happy to hear your thoughts on this!

agoose77:
@douglasdavis it's late my end, so this is just a superficial response. 

This looks interesting! One note that I can add is that we would like for the order of the columns to be the same whether all fields are read, or only a subset. It might be nice for a different interface to specify missing columns, e.g. a dict (which are now ordered by default) that indicates if the field should be read:
```python
columns = {
	'x': True,
	'y': True,
	'z': False
}
```

Will look over properly tomorrow! :)

agoose77:
I've rebased onto `main`, made some fixes for changes in awkward, and handled a form-layout deviation case (unmasked arrays for non-unmasked forms)

agoose77:
The unhandled case that's failing the CI is described in [this Gist](https://gist.github.com/agoose77/c877840a8d4a9ada1060b40343f66cbf#gistcomment-4568691). I believe it's a duplicate of https://github.com/apache/arrow/issues/30043, meaning it's likely not going to be fixed any time soon. We can drop trivial nodes like `UnmaskedArray`, so that's probably what needs to be done in this case.

agoose77:
6483a08 handles (option, no-option) for a (layout, form) pair, which fixes this bug. I think that's the only place that it's needed; we only see this option being introduced if one reads a subset of columns. If one reads a subset of columns, then re-hydration _always_ happens.

I suppose that a user could only read a subset of columns, which would trigger this bug (in plain Awkward, too, I suppose).

src/dask_awkward/lib/optimize.py
```diff
@@ -474,5 +474,7 @@ def _prune_wildcards(columns: list[str], meta: AwkwardArray) -> list[str]:
         else:
             if imeta.fields:
                 good_columns.append(f"{col[:-2]}.{imeta.fields[0]}")
+            else:
+                good_columns.append(col[:-2])
```
agoose77, line 478: @douglasdavis 

As I understand it, this pruning logic works at the "column" level, rather than buffer level. 

For completeness, a restriction of this approach is that it would involve over-reading in the case that the underlying data are stored as separate buffers, i.e. `array.offsets` can be read separately from `array.content`. I vaguely recall discussing this in a meeting a while ago. Does anyone else recall if we touched upon this?

This patch ensures that if a column has _no_ child fields, we include the column. Are you happy with this change? I don't want to make a mis-step here.

douglasdavis: > As I understand it, this pruning logic works at the "column" level, rather than buffer level. Is this a fair description? I.e., we work at the buffer level (form keys) in Awkward, but in this library we speak about columns (which may be composites of buffers).

Yes! this code is definitely written around the idea of columns

> If so, a restriction of this approach is that it would lead to over-reading in the case that the underlying data are stored as separate buffers, i.e. if array.offsets can be read separately from array.content. I vaguely recall discussing this in a meeting a while ago. Does anyone else recall if we touched upon this?

I _kind of_ recall this as well, but I don't recall us discussing any further than what you've mentioned (i.e. the conversation was pretty much "it could happen" and not a discussion about solutions/workarounds)

> This patch ensures that if a column has no child fields, we just include the column itself. Are you happy with this change? I don't want to make a mis-step, but that seems like the appropriate fall-back to generalise this logic to non-record arrays.

> The consequence of this change is that the reproducer in this PR description no longer fails because we must read Muon_phi. That is to say, the expected behavior is that Muon_phi be read, because we need its offsets (which can only be read if the column itself is entirely read).

I need to re familiarize myself with this code a bit before concocting a reasonable answer!

douglasdavis:
@agoose77 I've been giving this PR a scan and also taking it for a spin locally with a Parquet dataset. I've not run into any problems

I've also started working on adding an API (starting with a decorator) that will allow libraries (at this point that just means uproot 😛) to wrap the concrete callable such that `unproject_layout` will be called on the returned array with the original form. As things currently stand, in the builtin parquet IO support in dask-awkward, we can write in an `unproject_layout` call ourselves, but we definitely want to make it usable by others.

agoose77:
@douglasdavis fab! Locally, I am experimenting with modifying unprojection to create fake arrays instead of content / index objects. I'll see if that bears much fruit!

src/dask_awkward/lib/optimize.py
lgray: @agoose77 yes - that's what we do for most nanoevents. We have some pieces of our DSL for loading bits of columns, offsets, content, etc. We take apart each `Muon_*` and index it by `nMuon`, for instance.

lgray:
Should I give this a spin with nanoevents / uproot soon?

douglasdavis:
@lgray Yea I think so, I'm typing up a little update on an API for rehydrating an array in the method that instantiates it

douglasdavis:
I made a new branch to experiment with the API I mentioned in my comment above. The commit which adds two different possible interfaces: https://github.com/douglasdavis/dask-awkward/commit/d91ddae5d04eae5702ce1a8cab8fa84e881b97e0

The first via a decorator interface and the second via a mixin interface.

In both cases the functor that does the reading from disk just needs a `self.original_form` attribute. You can see in that commit above how I've implemented it for the builtin `from_parquet` reading functor `_FromParquetFileWiseFn`; naming and things can be changed but so far this works for me locally. (I've done a little task graph surgery testing to see that we indeed get an array with the complete form but a bunch if 0 buffers on the fields that are optimized away

src/dask_awkward/lib/optimize.py
douglasdavis: Going back to @agoose77's comment:

> This patch ensures that if a column has no child fields, we just include the column itself. Are you happy with this change? I don't want to make a mis-step, but that seems like the appropriate fall-back to generalise this logic to non-record arrays.

> The consequence of this change is that the reproducer in this PR description no longer fails because we must read Muon_phi. That is to say, the expected behavior is that Muon_phi be read, because we need its offsets (which can only be read if the column itself is entirely read).

> I assume that to avoid the read of Muon_phi, the proper solution is to write an interface to the IO layer that can encode the appropriate prior (that Muon_ branches all share offsets). I believe this is what the remapping logic by @lgray does, right?

At this point I'm happy with the change if it fixes something and we're still able to avoid unnecessary reads (for example through the mechanism @lgray has with remapping).

I am just generally curious how you came across that triggered the need to add the
```python
else:
    good_columns.append(col[:-2])
```
branch

lgray: I'll give it a try properly early next week and report back!

agoose77:
This PR now depends upon https://github.com/scikit-hep/awkward/pull/2479 which introduces support for a placeholder array.

src/dask_awkward/lib/optimize.py
agoose77: > I am just generally curious how you came across whatever triggered the need to add the

Just saw this - At the time I'd been playing around with getting uproot working.

IIRC I was following this optimisation through, and encountered a situation where a dataless offsets buffer appeared alongside a non-dataless offsets buffer in a context that demanded both arrays not be dataless. I think it was broadcasting. Tracking this back, I could see that the wildcard `data_touched` entry was not being considered, which led me to this statement.

agoose77:
@douglasdavis I'm thinking more about https://github.com/dask-contrib/dask-awkward/pull/184#discussion_r1202611928

This PR addresses the problem of keeping the array metadata the same (e.g. `array.fields`) by including columns that are never ultimately touched.
e.g. 
```python
array = ak.Array(
    ak.contents.ListOffsetArray(
        ak.index.Index64(...),
        ak.contents.RecordArray(
            [...],
            fields=["x", "y"]
        )
    )
)

result = ak.sum(array.x)
```

In the above example, we always need to touch the offsets given in the `ListOffsetArray`, but we don't need to load the content corresponding to field `y`. 

The example given in the body of this PR is actually a slightly different problem: 
```python
x_rec = ak.Array(
    ak.contents.ListOffsetArray(
        ak.index.Index64(...),
        ak.contents.RecordArray(
            [...],
            fields=["x"]
        )
    )
)
y = ak.Array(
    ak.contents.ListOffsetArray(
        ak.index.Index64(...),
		...
    )
)
array = ak.with_field(x_rec, y, "y")
result = ak.sum(array.x)
```

Here, we would normally touch both `x_rec.layout.content.offsets` and `y.layout.offsets` in order to perform the broadcasting required by `with_field`. That "fix" is discussed above in https://github.com/dask-contrib/dask-awkward/pull/184#discussion_r1202611928 to ensure that this column is read. In other words, this PR _right now_ reads `y`, even though it doesn't survive to the final result.

This is technically correct. For example, in this concrete example,
```python
import awkward as ak
x_rec = ak.to_regular([
    [{'x': 1}],
    [{'x': 2}]
])
y = ak.Array([
    [1, 2, 3],
    [4, 5, 6, 7]
])

array = ak.with_field(x_rec, y, "y")
result = ak.sum(array.x)
```
the result is `11`, despite `x_rec.x` summing to `3`; the offsets of `y` lead to repetition of `x`. We cannot further optimise this because the result _does_ depend upon both sets of offsets. 

We could conceivably update Awkward to implement asymmetric broadcasting, such that `ak.with_field(x_rec, y, "y")` only broadcasts `y` against `x`; we've been considering that in other contexts, so it's natural to raise it here. With this change, the _value_ of the result would not depend upon the offsets of `y`. But, crucially, `y` could easily have offsets that are not asymmetrically broadcastable with `x_rec` that would raise an exception at runtime. If we implemented an optimisation such that we just ignore offsets during broadcasting if they don't affect the result _value_, then the above would succeed. But, if `y` ended up in the task graph, this earlier result would then fail because the broadcast would be properly evaluated.

In other words, creating records via `ak.with_field` cannot be considered (isn't) metadata-only, even though it's lightweight[^perf]. Transient fields will have offsets touched, so they must come from the same list if we're to ensure that there is no IO penalty to them. That's why the `uproot.dask` form rewriting solves this particular case; it can both retain the important context that both sets of offsets have the same form key, and it also can build records without broadcasting.


@jpivarski I think this should be on your radar, though we can discuss it later today. 

[^perf]: broadcasting lists requires comparing offsets, although we could make it even faster if we add an `this.offsets is that.offsets` optimisation.

douglasdavis:
Thanks for that detailed writeup, @agoose77! Please correct me if I'm misunderstanding here, but this is the general takeaway I'm getting:

The need for form rehydration has popped up a few times and this PR addresses that need. However, the example I give at the top of the PR _is not_ the best example to give as the motivation for form rehydration. And in the PR as it exists you've ensured that the original need is met (form rehydration with the `unproject_layout` function. And still in this PR you ensure that necessary columns for things that involve broadcasting (like the `zip` example that I originally wrote out) are covered in the wildcard pruning logic of the `optimize.py` file. Am I following this correctly?

agoose77:
@douglasdavis succinctly put, that is exactly it!

douglasdavis:
thanks @jpivarski and @agoose77 for helping to get this feature in!

PR #185, opened 2023-03-07
Add ability to have file prefix in to_parquet

douglasdavis:
@lgray is this what you had in mind?

```python
to_parquet(a, "/output/path", prefix="abc")
```
will write files of the form (50 partitions):
```
/output/path/abc-part00.parquet
...
/output/path/abc-part49.parquet
```

while
```python
to_parquet(a, "/output/path")
```
remains
```
/output/path/part00.parquet
...
/output/path/part49.parquet
```

(took the opportunity to add a bit of typing to this PR as well)


martindurant:
I think dask dataframe allows you to supply a template like "file-*.parqet"

lgray:
Yes, exactly! Thank you!

PR #187, opened 2023-03-14
Allow dak.from_parquet to apply form remappings like uproot

lgray:
Work in progress!

This will allow coffea to create nanoevents from parquet files.

### DIFF
--- a/docs/ht-behaviors.rst
+++ b/docs/ht-behaviors.rst
@@ -1,0 +1,58 @@
+Behaviors [experimental]
+------------------------
+
+.. warning::
+
+   Awkward-Array behaviors are currently experimental in dask-awkward!
+
+Awkward-Array's behaviors_ feature provides a powerful mechanism for
+attaching methods and properties to Awkward Array. There is support in
+dask-awkward for using behaviors, but the feature is currently
+experimental and still in design development; there is no guarantee of
+API stability at this time! WIth that caveat, it's still possible to
+use behaviors now.
+
+Here's a brief example (we suggest reading the :doc:`upstream
+documentation <awkward:reference/ak.behavior>` for more information on the
+topic).
+
+.. code:: python
+
+   import awkward as ak
+   import dask_awkward as dak
+
+   behavior: dict = {}
+
+   @ak.mixin_class(behavior)
+   class Point:
+       def distance(self, other):
+           return np.sqrt(
+               (self.x - other.x) ** 2 + (self.y - other.y) ** 2
+           )
+
+   points1 = ak.Array([
+        [{"x": 1.0, "y": 1.1}, {"x": 2.0, "y": 2.2}, {"x": 3, "y": 3.3}],
+        [],
+        [{"x": 4.0, "y": 4.4}, {"x": 5.0, "y": 5.5}],
+        [{"x": 6.0, "y": 6.6}],
+        [{"x": 7.0, "y": 7.7}, {"x": 8.0, "y": 8.8}, {"x": 9, "y": 9.9}],
+    ])
+
+   points2 = ak.Array([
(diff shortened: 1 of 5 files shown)


lgray:
This one is done now and passing integration tests in coffea.

martindurant:
I wonder, why do you need the parquet IO layer to have this functionality? It is far simpler conceptually to have a normal parquet IO layer and then map_partitions applying the form transform as the next one.

lgray:
@martindurant Thanks for the question, and I was trying to do it the way you are asking first but the problem comes down to the way that forms, ak.from_buffers, and column projection work as they are right now.

If I try to do this as a regular map_partitions and not at the data source then I do not have access to which columns need to be extracted from the "base form" (the awkward form of the file) and I am forced to read the entire file in because I cannot do `select_columns` on the "rendered form" (the awkward form I present to a user) to get the columns I need to read on the base form (and I am therefore forced to read the entire based form regardless of what data I need downstream since ak.from_buffers is completely eager. 

Only in the io layers can I simultaneously manipulate the rendered and base forms to keep them lined up in terms of necessary data since it's at the io layers where you know t.

Given that I've worked through all this and I understand how it all works a bit better I can give the way you mention another shot but did give it a good shot and could never get it to work right.

martindurant:
So an array that is touched following from_buffers does not propagate back to the original array? Maybe that's fixable, or general tree arrangement could be achieved without having to unpack and repack the buffers again; those seem like better options from my vantage, but require work in awkward upstream.

lgray:
Yeah the problem is that the `from_buffers` needs to be at the root of the task graph, where as a regular node it does not know about what what's getting touched later.

I can check again to see what's happening when it's running the typetracer, if it's at the root maybe it sees all the data that's touched further down / in leaves? I'll give it a try (probably on Monday).

lgray:
Ah - right - to specify the problem exactly. It is because the typetracer that's not in an iolayer does not have access to the result of the typetracer report (nor the report itself) to see what columns are requested. Since you cannot determine which data are touched you must request it all, which is incredibly inefficient.

Essentially if I had the typetracer's report during the column optimization step at the node that's doing the form transformation, then this would be possible.

lgray:
In particular if we could get access to the report from the root record array in the rendered form we could do this easily.

However, it's not attached at that point in the layout so we can't see it:
https://github.com/scikit-hep/awkward/blob/main/src/awkward/_nplikes/typetracer.py#L174-L177

lgray:
I think I take my usual stance here. I'd like to have this functionality in so I can proceed forward, and I am 100% happy to revisit and improve this as interfaces change. If we can drop the need for remapping forms at the source I would be rather happy since I could completely abstract away the choice of file type in coffea.

lgray:
I recall in a previous meeting that we agreed to let this in for parquet? 

Are there any refinements to the interface here that we'd want to make? (I'd have to make corresponding updates in uproot as well).

The only one I can think of is a hook for users to get relevant metadata from different parts of the file (but that's also a bit more streamlined with parquet compared to ROOT.

martindurant:
Yes, I think so. @douglasdavis remembers how we wanted to have a function or interface that all interested IO layers could participate in. I think it was to be a function call.

lgray:
Right now that's implemented as passing in something that conforms to an interface I defined some time ago.

So we need to make a common entry point setter/getter or a new constructor argument?

lgray:
In particular this interface: https://github.com/dask-contrib/dask-awkward/blob/main/src/dask_awkward/lib/io/io.py#L32-L48

douglasdavis:
I'm not partial to any particular way to accomplish the feature! Just having an abstraction that can be easily used in input layers in uproot, core dask-awkward, etc. with just a few lines of code (whichout reinventing the wheel) I think is ideal.

agoose77:
@lgray I'm going to close this as the column optimisation logic has moved quite far since it was first authored.

I'll follow up on Slack with you about next steps.

PR #189, opened 2023-03-17
fix: dak.cartesian was not tracing properly, fix data touching, behaviors

lgray:
@yimuchen if you take this PR you should be able to continue with your tests.

### DIFF
--- a/src/dask_awkward/lib/structure.py
+++ b/src/dask_awkward/lib/structure.py
@@ -234,6 +234,16 @@ class _CartesianFn:
         self.kwargs = kwargs
 
     def __call__(self, *arrays):
+        if ak.backend(*arrays) == "typetracer":
+            for array in arrays:
+                array.layout._touch_data(recursive=True)
+            return ak.cartesian(
+                list(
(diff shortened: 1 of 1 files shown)


lgray:
Original symptoms were:
```
function call on just metas failed
```
and then the resulting arrays were stripped of their behaviors... I'll add a fix to the `length_zero_array()` fallback as well.

lgray:
@agoose77 The problem with the data touching here was a bit interesting. In the nested form of `cartesian` if I asked for the unzipped components of the cartesian `["0"]` or `["1"]` then instead of needing to get the required data from both parts of the nested object since that defines the shape of the array it would only get the data related to that particular subarray and the calculation would fail since it couldn't do the original cartesian. 

With touching all typetracers before calling the cartesian it works fine.

agoose77:
I'm aware of some typetracer problems with the cartesian function here: https://github.com/scikit-hep/awkward/pull/2295/ 

I'm not sure if it's related to this one, though. (adds to pile!)

PR #191, opened 2023-03-20
Recurse through nested contents for arrays used in `getitem`

### LINKED ISSUE #188: Unable to use a concatenated array of records in masks
For example:

```python
>>> import awkward as ak, dask_awkward as dak
>>> a = ak.Array([[{'a': 1}]])
>>> da = dak.from_awkward(a, 1)
>>> da_concat = dak.concatenate([da, da], axis=1)
>>> da_concat.compute()
<Array [[{a: 1}, {a: 1}]] type='1 * var * {a: int64}'>
>>> da_concat[da_concat.a > 0]
  File "/home/user/iris-hep/src/dask-awkward/src/dask_awkward/lib/core.py", line 876, in _getitem_single
    dtype = where.layout.dtype.type
AttributeError: 'ListOffsetArray' object has no attribute 'dtype'

During handling of the above exception, another exception occurred:

  File "/home/user/iris-hep/src/dask-awkward/src/dask_awkward/lib/core.py", line 922, in __getitem__
    return self._getitem_single(where)
  File "/home/user/iris-hep/src/dask-awkward/src/dask_awkward/lib/core.py", line 878, in _getitem_single
    dtype = where.layout.content.dtype.type
AttributeError: 'IndexedArray' object has no attribute 'dtype'

```

masonproffitt:
Resolves #188.

I don't know if there's a preferred way to handle this, but this was the solution that I found. I believe the `contents` nesting can be arbitrarily deep, so I think there has to be a loop over them, unless there's an `awkward` shortcut that I'm not aware of.

### DIFF
--- a/src/dask_awkward/lib/core.py
+++ b/src/dask_awkward/lib/core.py
@@ -872,10 +872,10 @@ class Array(DaskMethodsMixin, NDArrayOperatorsMixin):
             return self._getitem_outer_int(where)
 
         elif isinstance(where, Array):
-            try:
-                dtype = where.layout.dtype.type
-            except AttributeError:
-                dtype = where.layout.content.dtype.type
+            layout = where.layout
+            while not hasattr(layout, "dtype"):
+                layout = layout.content
+            dtype = layout.dtype.type
             if issubclass(dtype, (np.bool_, bool, np.int64, np.int32, int)):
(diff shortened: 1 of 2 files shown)


douglasdavis:
LGTM!

PR #197, opened 2023-03-22
Check backend of `meta=` arguments in creation functions

douglasdavis:


douglasdavis:
@lgray this should do the trick w.r.t. leaking data-carrying types as `meta=`

PR #200, opened 2023-03-23
Use inspect to determine if we're passing in additional array contexts.

### LINKED ISSUE #193: Trying to call a behavior method fails with `unexpected keyword argument '__dask_array__'`
Example:

```python
>>> import awkward as ak, dask_awkward as dak, vector
>>> vector.register_awkward()
>>> a = ak.Array([{'x': 1, 'y': 1}])
>>> da = dak.from_awkward(a, 1)
>>> dv = dak.with_name(da, 'Vector2D')
>>> dv.deltaphi(dv)
  File "/home/user/iris-hep/src/dask-awkward/src/dask_awkward/lib/core.py", line 990, in wrapper
    return self._call_behavior_method(attr, *args, **kwargs)
  File "/home/user/iris-hep/src/dask-awkward/src/dask_awkward/lib/core.py", line 932, in _call_behavior_method
    return getattr(self._meta, method_name)(
TypeError: Planar.deltaphi() got an unexpected keyword argument '__dask_array__'
```

The plain `awkward` version works as expected:

```python
>>> v = ak.with_name(a, 'Vector2D')
>>> v.deltaphi(v)
<Array [0] type='1 * float64'>
```

lgray:
For methods we check for `__dunder__` parameters in the function signature.

For properties we check for the existence of a single additional parameter which *is* the array context `self`.

Fixes #193 

Also fixes a possible bug in `__setitem__` where it wouldn't raise if you assigned to some random thing.

### DIFF
--- a/src/dask_awkward/lib/core.py
+++ b/src/dask_awkward/lib/core.py
@@ -1,5 +1,6 @@
 from __future__ import annotations
 
+import inspect
 import keyword
 import logging
 import operator
@@ -521,7 +522,7 @@ class Array(DaskMethodsMixin, NDArrayOperatorsMixin):
             isinstance(where, str)
             or (isinstance(where, tuple) and all(isinstance(x, str) for x in where))
         ):
-            TypeError("only fields may be assigned in-place (by field name)")
+            raise TypeError("only fields may be assigned in-place (by field name)")
 
         if not isinstance(what, Array):
             raise DaskAwkwardNotImplemented(
@@ -959,15 +960,33 @@ class Array(DaskMethodsMixin, NDArrayOperatorsMixin):
 
     def _call_behavior_method(self, method_name: str, *args: Any, **kwargs: Any) -> Any:
         """Call a behavior method for an awkward array.
-        Please note that the raw dask_awkward arguments are forwarded to the
-        awkward Mixin class, and the user must deal with when to make calls
-        to map_partitions.
+        If the function signature has __dunder__ parameters it is assumed that the
+        user wants to do the map_partitions dispatch themselves and the _meta's
+        behavior is called.
+        If there are no __dunder__ parameters in the function call then the function
+        is wrapped in map_partitions automatically.
         """
         if hasattr(self._meta, method_name):
-            array_context = kwargs.pop("__dask_array__", self)
-            return getattr(self._meta, method_name)(
-                *args, __dask_array__=array_context, **kwargs
+            themethod = getattr(self._meta, method_name)
+            thesig = inspect.signature(themethod)
+            ndunder = sum(
+                [
+                    1
+                    if len(param) > 4
+                    and param.startswith("__")
+                    and param.endswith("__")
+                    else 0
+                    for param in thesig.parameters
+                ]
             )
+            if ndunder > 0:
+                return themethod(*args, **kwargs)
+            return self.map_partitions(
+                _BehaviorMethodFn(method_name, **kwargs),
+                *args,
+                label=hyphenize(method_name),
+            )
+
         raise AttributeError(
             f"Method {method_name} is not available to this collection."
         )
@@ -978,22 +997,32 @@ class Array(DaskMethodsMixin, NDArrayOperatorsMixin):
         if a user follows the pattern:
 
         class SomeMixin:
-            def get_the_property(self, __dask_array__ = None):
-                ...
 
-            the_property = property(get_the_property)
+            @property
+            def the_property()
 
-        This pattern is caught and reissued as a method call to the "get_the_property"
-        method, passing self as __dask_array__.
+            @property
+            def a_property(array_context=None) # note: this can be any name
 
-        If get_the_property is not defined, and the_property is decorated with
-        @property, then then no calling context is needed and the property is
-        mapped directly.
+        This pattern is caught if the property has an argument that single
+        argument is assumed to be the array context (i.e. self) so that self-
+        referenced re-indexing operations can be hidden in properties. The
+        user must do the appropriate dispatch of map_partitions.
+
+        If there is no argument the property call is wrapped in map_partitions.
         """
         if hasattr(self._meta.__class__, property_name):
-            if hasattr(self._meta.__class__, f"get_{property_name}"):
-                return self._call_behavior_method(
-                    f"get_{property_name}",
+            thegetter = getattr(self._meta.__class__, property_name).fget.__get__(
+                self._meta
+            )
+            thesig = inspect.signature(thegetter)
+
+            if len(thesig.parameters) == 1:
+                binding = thesig.bind(self)
+                return thegetter(*binding.args, **binding.kwargs)
+            elif len(thesig.parameters) > 1:
+                raise RuntimeError(
+                    "Parametrized property cannot have more than one argument, the array context!"
                 )
             return self.map_partitions(
                 _BehaviorPropertyFn(property_name),
--- a/tests/test_behavior.py
+++ b/tests/test_behavior.py
@@ -1,7 +1,5 @@
 from __future__ import annotations
 
-from typing import Any
-
 import awkward as ak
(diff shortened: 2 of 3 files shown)


lgray:
@agoose77 @douglasdavis @masonproffitt This should fix that issue. Please review / comment / alter as needed. :-)

agoose77:
Thanks @lgray!

I think it might be worth discussing what features we want out of this. In my view:
1. Don't break existing naive behaviour classes
2. Support specialised behaviour where data-less arrays aren't useful
3. Support specialised behaviour where partition-wise implementations are not self-evident

@jpivarski and I discussed this, and I see one motivation for the original `__dask_awkward__` approach is for operations that don't have a natural associative operation. Mathematically, I think that is to say that behaviour methods naturally map onto monoids. For behaviour methods returning histograms, that is the set of histograms which combine under addition. For arrays, that's the set of arrays which combine under concatenation.

I think (2) can be realised with `ak.backend()`; if an array is backed by typetracer, then the user can optionally realise it with `length_zero_array()` and then convert the result back to typetracer with `ak.to_backend()`. Or, we perhaps want to implement scikit-hep/awkward#2127.

(3) is not possible without the user specifying the associative combination operation. An example would be a behaviour method that returns a scalar. Behaviour authors would need to define a dask partition-wise combination function. 

This makes me think that the original (i.e. reverted) API should be sufficient for (1) and (2).
(3) is trickier, and I think warrants a decorator, e.g.

```python
def _total_sum_agg(this, that):
	return do_some_more_sums(this, that)

class MyBehavior(ak.Array):
    @dak.aggregated(_total_sum_agg)
    def total_sum(self):
        return ak.sum(self, axis=None)
```

Ignore the API; I'm not in a good API design mindset just now! 

@lgray the immediate question that comes to mind is "is (2) the problem that you have right now, or is it (3)"?

lgray:
All it really want out of this is to be able to specify some array context so that I can reference back to an original set of global offsets for arrays correctly and hence cut on gen particle lists (which are self-referential arrays where the value at a given position is the index of a parent particle, for instance) and maintain the gen-particle parentage correctly.

To reply to your criteria:
1) This cannot not break any existing behavior that didn't use this functionality in the first place since it is opt-in only.
2) I do not know what this means or how this implementation violates it
3) I do not know what this means or how this implementation violates it

Trying to read it a different way to attempt to understand 2 & 3, I think this supports specialized behavior in a fairly covering way since the user is given control of all inputs to the behavior methods being called?

I am not sure why this has anything to do with histograms, and typetracers work with this implementation naturally.

I think the best way forward is to demonstrate something that this new interface breaks.

lgray:
Reading it again... I think I get what 2 and 3 mean now, that took a couple passes.

For 2 in my specific case since I need to both type-trace the fact that I need to call to the root-node of the dask array for the analysis statement.

For 3 I only really care about this for arrays, personally.

An example in coffea nanoevents would be:
```
muon_parents = events.GenPart.distinctParent[abs(events.GenPart.pdgId) == 13].pdgId.compute()
```

If I do not have the ability to specify an array context then I will get the wrong answer since the indices for the parentage will be wrong from the `abs(pdgId) ==13` cut.

lgray:
@masonproffitt can you check that this branch solves your issue? vector works for me now but just in case there's some corner I didn't catch.

agoose77:
@lgray it's late with me so I might not be computing what the problem with the naive `map_partitions` logic is; what part of that code doesn't just work if you perform it chunk-wise and then concatenate the results (as dask-awkward would)?

> Reading it again... I think I get what 2 and 3 mean now, that took a couple passes.

I've elaborated a touch. The motivation here is that we _know_ behaviours can do more than just return arrays, and it is worth answering the question of "what's the least work we can get away with that gives us freedom to grow". We mightn't need to implement `dask.aggregate` right now, but it is something we'll probably need in the long future.

lgray:
@agoose77 Ah! I see where you are coming from. I think we should factor this into two issues.

We certainly know that behaviors can return more than just arrays, and I agree it is worth answering the question about how to get things in good shape for further growth.

However, the original scope of this function did not make any attempt to address this and I think the first order of business is to restore the functionality that @masonproffitt wants for vector.

See: https://github.com/dask-contrib/dask-awkward/blob/2023.2.0/src/dask_awkward/lib/core.py#L910-L929 from before the breaking PR was introduced.

I propose we confirm things are good for Mason, and take the discussion on expanding the interface to an issue and then an alterative PR.

lgray:
As far as code that doesn't work with naive `map_partitions` logic it's this: 

https://github.com/CoffeaTeam/coffea/blob/awkward2_dev/coffea/nanoevents/methods/nanoaod.py#L204-L212
which calls into:
https://github.com/CoffeaTeam/coffea/blob/awkward2_dev/coffea/nanoevents/methods/base.py#L205-L236

This is necessary to do (the take operation on the back-reference to the original global index) if we want to maintain the validity of cross references through masking operations on the original array. It is a specific quality of NanoAOD that we are exploiting with nanoevents that people quite like to use!

douglasdavis:
With that docs page in I'm satisfied to merge this. AFAIK @lgray and @masonproffitt are the only people who have pushed the buttons on behaviors, and they are both aware of the situation here.

masonproffitt:
> @masonproffitt can you check that this branch solves your issue? vector works for me now but just in case there's some corner I didn't catch.

Yeah, this seems fine to me

src/dask_awkward/lib/core.py
```diff
@@ -959,15 +960,33 @@ def __getitem__(self, where: Any) -> AwkwardDaskCollection:
...
+            ndunder = sum(
+                [
+                    1
+                    if len(param) > 4
+                    and param.startswith("__")
+                    and param.endswith("__")
+                    else 0
+                    for param in thesig.parameters
+                ]
+            )
```
agoose77, line 981: I'm not sure about checking for `__`. Unless we know that we're going to want to declare a set of parameters, instead I'd prefer an explicit test for `_dask_array_` (using single underscores, as dunder is usually reserved for Python protocols).

src/dask_awkward/lib/core.py
```diff
@@ -978,22 +997,32 @@ def _call_behavior_property(self, property_name: str) -> Any:
         if a user follows the pattern:
 
         class SomeMixin:
-            def get_the_property(self, __dask_array__ = None):
-                ...
 
-            the_property = property(get_the_property)
+            @property
+            def the_property()
```
agoose77, line 1002:
```suggestion
            @property
            def the_property(self):
                ...
```

src/dask_awkward/lib/core.py
```diff
@@ -959,15 +960,33 @@ def __getitem__(self, where: Any) -> AwkwardDaskCollection:
...
+                    else 0
+                    for param in thesig.parameters
+                ]
+            )
+            if ndunder > 0:
+                return themethod(*args, **kwargs)
+            return self.map_partitions(
+                _BehaviorMethodFn(method_name, **kwargs),
+                *args,
+                label=hyphenize(method_name),
```
agoose77, line 987: @lgray this looks like the same problem (3) as I was describing, albeit more general in that you're not just dealing with specifying the associative operation.

I still don't understand your use case, but I've not spent a great deal of time yet to figure it out.

I'm worried that this is solving too specific a problem; I'd like to find the more general problem and solve that. I will give this more thought, but I'm on holiday/thesis submission the next week, so time is thin on the gruond:/

src/dask_awkward/lib/core.py
agoose77: The same thoughts apply here as https://github.com/dask-contrib/dask-awkward/pull/200/files#r1148017561

douglasdavis:
Alrighty putting this in to allow @lgray to continue pushing ahead, we're all in agreement that this part of dask-awkward is unstable with plans for further development (as mentioned in the docs page)

PR #203, opened 2023-03-30
feat: added an unproject_layout function

jpivarski:
This is 99% of an `unproject_layout` function. The only public function in this new file is

```python
def unproject_layout(form: Form, layout: Content) -> Content
```

where the `form` is the "original Form," the structure of what you want this function to return as an array (low-level `Content`), and `layout` is the "projected array," which has a subset of the fields of the "original Form."

You can also verify (by assertion) that they are

```python
def compatible(form: Form, layout: Content) -> bool
```

before running the recursive function (`_unproject_layout`).

@agoose77 and I were talking about this after the dask-awkward meeting. The difficult part to handle is the union type, as well as interactions between the union type and an option type. When constructing the projected array, unnecessary unions get simplified (they _have_ to, because of some rules we're established), so that part of the projected array can have different `Content` nodes than the corresponding `Form` nodes in the "original Form." When we fully handle that, it will probably only be a few special cases (no more than 3, I'm guessing). That's what the

```python
# handle other cases that come up here...
```

is; we expect to have to handle some cases in which the `Form` class doesn't exactly correspond to the `Content` class.

The test function tests _a few_ cases, but that gets us past the obvious bugs. It will be easier to test thoroughly in the context of the full Dask machinery, since that provides the projected arrays in a way that I don't have to build manually. Once you've hooked this up, I can get a full suite of Awkward structures that we can run through, you provide the projections, and then we handle the special cases as we encounter them.

--------------

There is not very much of this that is duplicated from `ak.from_buffers`; only passing down the predicted `length` through the recursion, and even that is simplified in this case. It makes a lot more sense as a specialized function than trying to shoehorn it into `ak.from_buffers`.

### DIFF
--- a/src/dask_awkward/unproject_layout.py
+++ b/src/dask_awkward/unproject_layout.py
@@ -1,0 +1,362 @@
+import math
+
+import numpy as np
+
+import awkward as ak
+from awkward._nplikes import nplike_of
+from awkward.forms import (
+    Form,
+    EmptyForm,
+    NumpyForm,
+    BitMaskedForm,
+    ByteMaskedForm,
+    IndexedForm,
+    IndexedOptionForm,
+    ListForm,
+    ListOffsetForm,
+    RegularForm,
+    UnmaskedForm,
+    RecordForm,
+    UnionForm,
+)
+from awkward.contents import (
+    Content,
+    EmptyArray,
+    NumpyArray,
+    BitMaskedArray,
+    ByteMaskedArray,
+    IndexedArray,
+    IndexedOptionArray,
+    ListArray,
+    ListOffsetArray,
+    RegularArray,
+    UnmaskedArray,
+    RecordArray,
+    UnionArray,
+)
+
+
+class DummyIndex:
+    def __init__(self, length, nplike):
+        self._length = length
+        self._nplike = nplike
+
+    def __len__(self):
+        return self._length
+
+    @property
+    def length(self):
+        return self._length
+
+
+class DummyIndex8(DummyIndex, ak.index.Index8):
+    @property
+    def dtype(self):
+        return np.dtype(np.int8)
+
+
+class DummyIndexU8(DummyIndex, ak.index.Index8):
+    @property
+    def dtype(self):
+        return np.dtype(np.uint8)
+
+
+class DummyIndex32(DummyIndex, ak.index.Index8):
+    @property
+    def dtype(self):
+        return np.dtype(np.int32)
+
+
+class DummyIndexU32(DummyIndex, ak.index.Index8):
+    @property
+    def dtype(self):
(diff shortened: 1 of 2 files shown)


douglasdavis:
Thanks a log @jpivarski; looks like I've run into a first uncovered case:

```
AssertionError: unexpected combination: <class 'awkward.forms.recordform.RecordForm'> and <class 'awkward.contents.unmaskedarray.UnmaskedArray'>
```

here's the "good array" layout:

```
{
    "class": "RecordArray",
    "fields": [
        "luminosityBlock",
        "muons"
    ],
    "contents": [
        "int64",
        {
            "class": "ListOffsetArray",
            "offsets": "i32",
            "content": {
                "class": "UnmaskedArray",
                "content": {
                    "class": "RecordArray",
                    "fields": [
                        "pt",
                        "eta"
                    ],
                    "contents": [
                        "float64",
                        "float64"
                    ]
                }
            }
        }
    ]
}
```

and the original form:

```
{
    "class": "RecordArray",
    "fields": [
        "run",
        "luminosityBlock",
        "event",
        "MET",
        "muons",
        "gen"
    ],
    "contents": [
        "int64",
        "int64",
        "int64",
        {
            "class": "RecordArray",
            "fields": [
                "pt",
                "phi"
            ],
            "contents": [
                "float64",
                "float64"
            ]
        },
        {
            "class": "ListOffsetArray",
            "offsets": "i32",
            "content": {
                "class": "RecordArray",
                "fields": [
                    "pt",
                    "eta",
                    "phi",
                    "mass",
                    "charge"
                ],
                "contents": [
                    "float64",
                    "float64",
                    "float64",
                    "float64",
                    "int64"
                ]
            }
        },
        {
            "class": "ListOffsetArray",
            "offsets": "i32",
            "content": {
                "class": "RecordArray",
                "fields": [
                    "pt",
                    "eta",
                    "phi",
                    "pdgId"
                ],
                "contents": [
                    "float64",
                    "float64",
                    "float64",
                    "int64"
                ]
            }
        }
    ]
}
```

The example I'm working with here;

```python
ds = dak.from_parquet("s3://ddavistemp/hpq", storage_options={"anon": True})
ds.muons.pt + ds.muons.eta - ds.luminosityBlock
```

In the projected array's form we end up getting an UnmaskedArray on top of the RecordArray for the `muons` field

PS I'm working in PR #184 (I've merged your code into it, but will leave this open as we iterate)

douglasdavis:
Another failed combination popped up in tests in #184 
```
FAILED tests/test_parquet.py::test_columns - AssertionError: unexpected combination: <class 'awkward.forms.bitmaskedform.BitMaskedForm'> and <class 'awkward.contents.unmaskedarray.UnmaskedArray'>
```

PR #210, opened 2023-04-04
optimise linear layer chains

martindurant:


martindurant:
I suspect this fails when reducing in any layer (i.e,. number of indices decreases), but for the purposes of dask-awkward we might be fine, where there is always one index of iteration: the partition.

douglasdavis:
related: https://github.com/dask/dask/issues/9795

martindurant:
@rjzamora , you might be interested to look at this. It implements a linear-time "fuse" operation on chains of Blockwise layers which is much faster that standard dask's layer fuse for very large numbers of layers. As it stands, the algorithm passes the specific two tests written for it, but causes many other dask-awkward tests to fail. I believe that there are two most likely causes:
- we should be excluding reductive layers from consideration, but aren't. It would be even better if we could cope with these, but the number of indices changes and that would need special attention.
- something is not quite right with the keys included in the subgraph versus the keys mentioned in the various attributes of a fused layer.

Any thoughts you might have on this would be appreciated!

src/dask_awkward/lib/optimize.py
```diff
@@ -42,6 +42,10 @@ def all_optimizations(
     else:
         # Perform dask-awkward specific optimizations.
         dsk = optimize(dsk, keys=keys)
+
+        # blockwise layer chaining optimization.
```
lgray, line 46: this should go in `def optimize` below so it gets called when dask_histogram is the final layer.

martindurant:
Good to go?

lgray:
As far as I can tell from torture tests, yes!

PR #217, opened 2023-04-05
feat: support Ellipsis as first arg in multi arg getitem

douglasdavis:


src/dask_awkward/lib/core.py
```diff
@@ -892,6 +892,9 @@ def _getitem_tuple(self, where: tuple[Any, ...]) -> Array:
             if issubclass(dtype, (np.bool_, bool, np.int64, np.int32, int)):
                 return self._getitem_outer_bool_or_int_lazy_array(where)
 
+        elif where[0] is Ellipsis:
+            return self._getitem_trivial_map_partitions(where)
```
martindurant, line 896: Is it not the case that ellipsis can evaluate to zero indices?

douglasdavis: Ah yes for something not deeply nested we can end up with something like

```python
In [25]: a = ak.Array({"x": [1, 2, 3]})

In [26]: a
Out[26]: <Array [{x: 1}, {x: 2}, {x: 3}] type='3 * {x: int64}'>

In [27]: a[..., 0]
Out[27]: <Record {x: 1} type='{x: int64}'>
```

martindurant: This will be what is desired in most sensible user code. We, again, need to know the number of possible axes.

douglasdavis: Right, I think a check on `ndim` vs `len(where)` will give us info we need to determine if we're good to use the trivial case

PR #222, opened 2023-04-08
Remove workaround in cartesian; increase minimum awkward version

lgray:
Will not function correctly until https://github.com/scikit-hep/awkward/pull/2375 is merged and released.

I'll add a new pin to this PR once it is out.

@jpivarski @agoose77


lgray:
@agoose77 have a look, merge when you have time - if there's other old to-do's I can drop I'm happy to add that in as well.

agoose77 (approved):

PR #227, opened 2023-04-12
Modify workaround for touching `what` in dak.with_name

### LINKED ISSUE #223: Fields added by `dak.with_field` cannot be optimized away
Not much more to say than in the title.

```python3
import uproot
import dask_awkward as dak

x["something_silly"] = x.Muon_looseId
print(dak.necessary_columns(x.Muon_pt))
```

prints out:
```
{'from-uproot-ffd643ed3e3f19249385270346179fbb': ['Muon_looseId', 'Muon_pt']}
```

and should print out:
```
{'from-uproot-ffd643ed3e3f19249385270346179fbb': ['Muon_pt']}
```

lgray:
This can be relaxed now but is still a consequence of #223.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -27,7 +27,7 @@ classifiers = [
   "Topic :: Scientific/Engineering",
 ]
 dependencies = [
-  "awkward >=2.1.2",
+  "awkward >=2.1.3",
   "dask >=2022.02.1",
 ]
 dynamic = ["version"]
--- a/src/dask_awkward/lib/structure.py
+++ b/src/dask_awkward/lib/structure.py
@@ -890,6 +890,9 @@ class _WithFieldFn:
         self.behavior = behavior
 
     def __call__(self, base: ak.Array, what: ak.Array) -> ak.Array:
+        # TODO: remove backend handling when touch is handled automatically
+        if ak.backend(what) == "typetracer":
+            what.layout._touch_data(recursive=False)
         return ak.with_field(
             base,
             what,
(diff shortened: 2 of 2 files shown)


lgray:
Do not merge until https://github.com/scikit-hep/awkward/pull/2395 is merged!

martindurant:
We should be able to make a test showing that necessary_columns does the right thing, no?

lgray:
Yes - but it seems this is also a partial fix (since 2395 is a partial fix). Working out the edge cases.

lgray:
@agoose77 FYI - the partial fix is indeed partial. It's a bit strange that I need to touch some of these, at least I can relax the over-aggressive touching.

lgray:
So digging a bit deeper this change appears to be a consequence `with_field` always spawning a `map_partitions` which means the data in `what` must be read and attached to the array (so we *must* touch `what` but we don't need to touch anything below what unless it's specifically called for). 

The behavior at the type tracer level seems reasonable:
```python3
muon = ttarray.muon
print("muon: ", report.data_touched)
muon["weight"] = ttarray.weight
print("assign:", report.data_touched)
``` 
yields:
```
muon:  []
assign: ['muon_list_outer!']
```

Which is correct behavior since that never accesses `weight`. similarly, in the more complex test the equivalent of `weight` is assigned but never accessed.

So, actually, instead of anything to do with fixing #223 it is an example of it. Since the typetracer is doing the right thing and we are only running into this because we do not cull un-used columns that are attached by `with_field`.

lgray:
Please merge this one when awkward pushes its next release as the associated fix is now in main.

PR #229, opened 2023-04-14
Automate releases with GH actions

douglasdavis:


.github/workflows/pypi-release.yml
```diff
@@ -0,0 +1,32 @@
...
+        fetch-depth: 0
+
+    - name: Setup Python
+      uses: actions/setup-python@v4
+      with:
+        python-version: "3.10"
+
+    - name: Build
+      run: |
+        python3 -m pip install pip wheel build -U
```
martindurant, line 25: hatch ?

douglasdavis: It will get installed in a temp environment automatically by 

    python3 -m build

martindurant:
OK, let's merge. We won't know for sure until we try it the first time! We may need to apply tag multiple times in that case, should things go wrong.

PR #230, opened 2023-04-14
feat: Allow `dak.where`, `dak.with_field` and `__setitem__` to take some non-array arguments

### LINKED ISSUE #226: Automate wrapping of float/int in awkward operations
There are certain operations that fail when passing in a float or int object 

```python
import awkward as ak
import dask_awkward as dak
import numpy as np

arr = ak.zip({
    'x': ak.from_regular(np.random.random(size=(10, 10))),
    'y': ak.from_regular(np.random.random(size=(10, 10)))
})
arr = dak.from_awkward(arr, npartitions=1)

arr = dak.where(arr.x > 0.5, 1.0, arr.y) # 'float' object has no attribute 'npartitions'
arr['z'] = 2.0 # DaskAwkwardNotImplemented: Supplying anything other than a dak.Array to __setitem__ is not yet available!
```

The user side solution is rather simple, instead of floats or int, we wrap it in `dak.ones_like(arr.x) * <int>`, but it would be nice if this can be handled automatically.

lgray:
Partially fixes #226 for `where` and `with_field`

Allows users to pass in non-dak arguments for certain arugments of where and with_field.

@yimuchen

### DIFF
--- a/src/dask_awkward/lib/core.py
+++ b/src/dask_awkward/lib/core.py
@@ -526,9 +526,9 @@ class Array(DaskMethodsMixin, NDArrayOperatorsMixin):
         ):
             raise TypeError("only fields may be assigned in-place (by field name)")
 
-        if not isinstance(what, Array):
+        if not isinstance(what, (Array, Number)):
             raise DaskAwkwardNotImplemented(
-                "Supplying anything other than a dak.Array to __setitem__ is not yet available!"
+                "Supplying anything other than a dak.Array, or Number to __setitem__ is not yet available!"
             )
 
         from dask_awkward.lib.structure import with_field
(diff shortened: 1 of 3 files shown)


lgray:
#226 asks for a general, automated fix for uses cases like `ak.where(arr > x, 1.0, arr)`, but this offers specific fixes that can get analysts moving along.

lgray:
I am not sure why these tests are failing. Looks like some recent update in a dependency breaks things horribly.

martindurant:
Can you temporarily pin dask to the previous version to see if that's it? dask 2023.4.0 is 2 hours old.

lgray:
Yep - it's the most recent dask.

martindurant:
OK, let's make a new issue for that, and I'll look into it next week.

martindurant (commented):
Let's try to hit those extra branches, as reported by codecov

lgray:
@martindurant ok should be good now. Squash this one when you merge it to clean up the clutter - was hacking on this between other things.

PR #232, opened 2023-04-17
dask_un/pack are gone!

### LINKED ISSUE #231: dask_awkward tests fail with dask 2023.4.0
To be seen if this requires repinning, etc.

martindurant:
Fixes #231

@lgray , this will ned careful vetting before allowing it in.

### DIFF
--- a/src/dask_awkward/layers/__init__.py
+++ b/src/dask_awkward/layers/__init__.py
@@ -1,3 +1,3 @@
-from dask_awkward.layers.layers import AwkwardInputLayer
+from dask_awkward.layers.layers import AwkwardBlockwiseLayer, AwkwardInputLayer
 
-__all__ = ("AwkwardInputLayer",)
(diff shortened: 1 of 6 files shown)


lgray:
I am pretty sure I am not qualified to review this, but it needs to go in first in terms of PRs to merge.

@jpivarski @agoose77 could you please give this a look? Thanks!

martindurant:
I think given the tests here, if column pruning was still working ok, we can be happy with this. It should only really affect distributed-dask, though.

I will need to push a version bound on dask before merger, hold on.

@douglasdavis FYI: the consequence of this is that a HLG that has been through pickle/unpickle no longer has _meta objects on its graph's layers. That's fine for the scheduler. However, this new graph cannot any. longer be optimised by our own code. Although we could made code to regenerate the _meta objects, I don't think there's any need.

PR #234, opened 2023-04-17
fix: only do column optimization on one partition of input dataset

### LINKED ISSUE #233: dask_awkward column optimization scales linearly with number of partitions in dataset
This causes major problems for the typically heavy HEP task graphs (for a complete dataset optimization times would be measured in hours).

The fix is something along the lines of only running the column optimization once per AwkwardInputLayer by artificially reducing the number of partition to analyze to 1 for the column optimization.

lgray:
Fixes #233

Posting a PR with @martindurant's patch to lodge here results of conversation / hacking in slack.

With some of the suggested changes in https://github.com/dask-contrib/dask-histogram/pull/74, it makes the complete analysis chain optimizable looking at only one partition per input dataset.

### DIFF
--- a/tests/test_core.py
+++ b/tests/test_core.py
@@ -324,20 +324,20 @@ def test_scalar_dtype() -> None:
     assert c.dtype is None
 
 
-def test_scalar_pickle(daa: Array) -> None:
-    import pickle
-
-    s = 2
-    c1 = new_known_scalar(s)
-    s_dumped = pickle.dumps(c1)
-    c2 = pickle.loads(s_dumped)
-    assert_eq(c1, c2)
-    s1 = dak.sum(daa["points"]["y"], axis=None)
-    s_dumped = pickle.dumps(s1)
-    s2 = pickle.loads(s_dumped)
-    assert_eq(s1, s2)
-
-    assert s1.known_value is None
+# def test_scalar_pickle(daa: Array) -> None:
+#    import pickle
+#
+#    s = 2
+#    c1 = new_known_scalar(s)
+#    s_dumped = pickle.dumps(c1)
+#    c2 = pickle.loads(s_dumped)
+#    assert_eq(c1, c2)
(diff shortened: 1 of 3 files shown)


src/dask_awkward/#__init__.py#
```diff
@@ -0,0 +1,19 @@
+from dask_awkward import config  # isort:skip; load awkward config
```
martindurant, line 1: This file is a debug remnant?

lgray:
@martindurant what further testing needs to be done for this one (I've run it through the large analysis and everything works fine).

The only odd bit is `visualize()` run before `compute()` with this patch appears to somehow spoil the necessary columns (if I avoid the visualize or do `visualize(optimize_graph=True)` the problem disappears. Intensely weird but not critical IMO.

lgray:
@martindurant what else is needed here?

tests/test_core.py
```diff
@@ -324,20 +324,20 @@ def test_scalar_dtype() -> None:
...
-    s_dumped = pickle.dumps(c1)
-    c2 = pickle.loads(s_dumped)
-    assert_eq(c1, c2)
-    s1 = dak.sum(daa["points"]["y"], axis=None)
-    s_dumped = pickle.dumps(s1)
-    s2 = pickle.loads(s_dumped)
-    assert_eq(s1, s2)
-
-    assert s1.known_value is None
+# def test_scalar_pickle(daa: Array) -> None:
```
martindurant, line 327: Why does this fail? Can it be reinstated? If not, we should either comment or remove it. I think I had a snippet to re-add the _meta objects to the layers, if that's what we need.

lgray: Yes that snippet will resurrect that test

martindurant: Do you need me to apply it?

lgray: I am presently in a car so it would be appreciated :-)

PR #235, opened 2023-04-19
fix: make ak.singletons workaround operate correctly in all cases

lgray:
Present implementation of singletons was failing column optimization when passed RecordArrays that were then accessed later.

### DIFF
--- a/src/dask_awkward/lib/structure.py
+++ b/src/dask_awkward/lib/structure.py
@@ -700,29 +700,36 @@ def run_lengths(array, highlevel=True, behavior=None):
     )
 
 
+class _SingletonsFn:
+    def __init__(self, axis, **kwargs):
+        self.axis = axis
+        self.kwargs = kwargs
+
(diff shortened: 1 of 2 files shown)


martindurant:
Will look shortly. Would appreciate a test case showing the previous failure now fixed.

lgray:
I'll see if I can boil it down. It popped up in a reasonably complex analysis query, it can take some time.

lgray:
@martindurant I couldn't get it to outright fail but I got it to cause the column optimization failure which eventually causes the failure in a more complex query. 

I made it so that test fails if a warning is emitted, which emulates the outcome in the more complex case. 

The test fails with previous impl and passes now, good enough?

martindurant:
Yes, that's reasonable.

martindurant:
Maybe we should keep a list in an issue of these "will remove workaround" situations.
Will merge when green.

PR #240, opened 2023-04-20
fix: remove zero_length_array meta calculation in full_like

lgray:
It is no longer necessary.


lgray:
@agoose77 Just to make sure I didn't miss anything how does https://github.com/scikit-hep/awkward/pull/2333 affect this PR to remove an old workaround?

agoose77:
@lgray it shouldn't. That PR fixes the case that array-like arguments are evaluated with `==`, rather than limiting it to scalars.

lgray:
Thanks (for indulging my paranoia)! :-)

PR #242, opened 2023-04-20
fix: remove _meta workaround in run_lengths

lgray:
it is no longer necessary.


martindurant:
These are looking familiar...

PR #243, opened 2023-04-20
example: dask.visualize(optimize_graph=False) causes later computation to fail

lgray:
@martindurant Here's what I was talking about in #234.

Here I can only get the column optimization to fail, and I raise it to an error, but in more complex situations this causes complete computation failure (instead of just reading the whole file which is also bad).


lgray:
OK - this is annoying to reproduce in a test. 
Just try the test locally or copy paste it and you'll see what I'm talking about. 

I cannot get it to fail correctly here.

martindurant:
So I have a fix, but I can't pretend to understand it. I wonder if something similar was happening when I was trying and failing to get layer fusing to work the simple way, since the exception feels the same.

lgray:
This fix is causing other failures to occur all over the place (uproot.dask, dask_histogram, etc.)

src/dask_awkward/layers/layers.py
```diff
@@ -31,14 +31,29 @@ def mock(self):
...
+    def __setstate__(self, state):
+        import awkward as ak
+
+        self.__dict__.update(state)
+        if "_meta" in state:
+            if hasattr(self._meta, "length_zero_array"):
+                self._meta = ak.Array(
+                    self._meta.length_zero_array(highlevel=False).to_typetracer(
+                        forget_length=True
+                    )
```
lgray, line 54: this needs to also pass the array behavior to the final high level array that is being made.

martindurant:
OK, so there were two things here: the pickle stuff that was leftovers from before, and clearing the _cache thing on layers. I think that to fix the original issue, maybe only the second was needed and the first is adding noise.

Also: this says we definitely need a more complete test suite in this project, as well as CI runs checking for breakages in other packages.

martindurant:
Here is just the "arcane" fix without the pickle part, and it now passes both things. 

I suspect that we are being caught between dask operations and our own optimization mutating objects, then using  deepcopy() as a workaround - which relies on sensible pickle/unpickle.

lgray:
Ok - I'll be able to check something really complex later this morning.

lgray:
I have tried this on the most complex example I have and this patch gives the correct answer running visualization before / after a compute. visualization and compute never fail.

I'm running on little torture test still but everything looks quiet positive!

lgray:
Yep - looks good!

PR #251, opened 2023-04-28
fix: drop workarounds for typetracer, use `highlevel=False`

agoose77:
This PR removes the last `length_zero_array()` usages in the dask-awkward operations, and also accounts for an upcoming deprecation of `highlevel=True` in `length_zero_array` (scikit-hep/awkward#2435)

### DIFF
--- a/src/dask_awkward/lib/core.py
+++ b/src/dask_awkward/lib/core.py
@@ -1737,7 +1737,7 @@ def to_meta(objects: Sequence[Any]) -> tuple[Any, ...]:
 
 def length_zero_array_or_identity(obj: Any) -> Any:
     if is_awkward_collection(obj):
-        return obj._meta.layout.form.length_zero_array(behavior=obj.behavior)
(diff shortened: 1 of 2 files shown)


lgray:
Ah, I'll have to make some changes in coffea as well, thanks for the heads up!

lgray:
I'll also check this to make sure there are no regressions in the adl-benchmark examples, and the larger full analysis code I have sitting around (which could serve well as a test along with a 100 event file or so)

martindurant:
Good to see the lines of red!

lgray:
Fixes #252

lgray:
I guess this one is waiting on next awkward?

lgray:
Ah - looks like we need to merge this to get the awkward@main tests functioning again. @martindurant

martindurant:
Upstream fix was included in https://github.com/scikit-hep/awkward/releases/tag/v2.1.4

PR #253, opened 2023-05-02
feat: add `repartition` method

martindurant:
repartition function: each output partition is made from one or more input partition with slicing.

Possible simplification could be added to only merge whole partitions without slicing, which would mean no need to compute lengths beforehand.


martindurant:
@lgray - if you like this implementation, I can also do slicing in a very similar way.

lgray:
```python3
import uproot
events = uproot.dask({"https://github.com/CoffeaTeam/coffea/raw/master/tests/samples/nano_dy.root": "Events"})
events.Muon_pt.repartition(npartitions=2)
```
fails with a rather long error message using this branch.

martindurant:
It's calculating the divisions that fails; looking...

martindurant:
Ah, it's a case of "needs rehydration". We optimize away all columns except Muon_pt, but when calling map_partitions(num), which has no "necessary columns", dak is picking column `run` (not sure how) to operate on as enough to be able to compute number of rows. So then the getitem is failing. 

With the change I just pushed, it's enough to run 
```
events.eager_compute_divisions()
```
(before column select) and then the repartition works.

lgray:
Basic uproot works, next level of error:

```python3
from coffea.nanoevents import NanoEventsFactory
events = NanoEventsFactory.from_root(
    "https://github.com/CoffeaTeam/coffea/raw/master/tests/samples/nano_dy.root",
    chunks_per_file=4,
    permit_dask=True,
).events()
events.eager_compute_divisions()
```

results in
```
...
File ~/coffea-dev/dask-awkward/src/dask_awkward/lib/core.py:90, in _finalize_array(results)
     84 else:
     85     msg = (
     86         "Unexpected results of a computation.\n "
     87         f"results: {results}"
     88         f"type of first result: {type(results[0])}"
     89     )
---> 90     raise RuntimeError(msg)

RuntimeError: Unexpected results of a computation.
 results: [10, 10, 10, 10]type of first result: <class 'int'>
```

martindurant:
^ this succeeded for me  at 778d89c
```
In [3]: events.divisions
Out[3]: (0, 10, 20, 30, 40)
```

lgray:
OK - looks to have been some version skew - reinstalled everything and it works.

lgray:
Very good!
image

And corresponding `.compute()` functions.

lgray:
What. Now it's not working again after resetting the notebook.

martindurant:
```
> /Users/mdurant/code/dask-awkward/src/dask_awkward/lib/core.py(90)_finalize_array()
     74     elif isinstance(results, tuple) and all(
     75         isinstance(r, (int, np.integer)) for r in results
     76     ):
     77         return ak.Array(list(results))
```

Adding `list` to the `isinstance` appears to fix it, but I don't know why this case is different.
but we have a list rather than a tuple?

martindurant:
(pushed the diff, if you want to try)

lgray:
Pleasant:
image

(with and without distributed client)

lgray:
Though given the fact we need `.eager_compute_divisions()` called ahead of time for it to function, probably something we don't yet want to expose to users. Fine for experimentation.

martindurant:
Rehydration will fix that. We could in theory also use the result of necessary columns to inform what num does when computing the division values, but I feel like that workaround should not be necessary.

lgray:
Using necessary columns to hint repartition would allow us to present a stable interface with no fiddly-ness.
Then we just make it better over time in the background.

+1

martindurant:
I think that it would be quite hard to achieve, though, and I gather #184 is about ready (@agoose77 )

src/dask_awkward/lib/core.py
```diff
@@ -711,6 +742,7 @@ def _partitions(self, index: Any) -> Array:
 
         # if a single partition was requested we trivially know the new divisions.
         if len(raw) == 1 and isinstance(raw[0], int) and self.known_divisions:
+            # TODO: don't we always know the divisions?
```
douglasdavis, line 745: I know uproot sometimes doesn't report the sizes of the partitions (I think it can make the `uproot.dask` call a bit faster to ignore determining the sizes of the partitions), and if we have something read from JSON we don't know the divisions (only npartitions)

martindurant:
@douglasdavis , look at the lovely red lines [91d1356](https://github.com/dask-contrib/dask-awkward/pull/253/commits/91d1356d7e99c935babf2c20e35ac49d4525300c)

douglasdavis:
yea that does look quite a bit easier to read; we can always come back to this PR if we need to revisit the layer

martindurant:
@douglasdavis : something shows about our column opt failures here.

If the tasks are like
```
                _zero_getitem,
                (self.name, i),
                slice(slice_start, slice_end, step),
                rest,
```
it fails, but if they are
```
                _zero_getitem,
                (self.name, i),
                slice(slice_start, slice_end, step)
```
(i.e, only one axis, which is the case in the "single" test) it passes. We need to investigate this!

martindurant:
The graph being run during column optimisation for the fail case is:
```
{('from-json-ccb017520b83fc52095908cdff28ce65', 0): (subgraph_callable-6231d24a-271b-4710-b599-3770d658fd8c, None),
 ('getitem-08a6b8d2e456920c909b54d9db4e8640', 0): (<function _zero_getitem at 0x11ec9d6c0>,
                                                   ('from-json-ccb017520b83fc52095908cdff28ce65', 0),
                                                   slice(0, 5, 1),
                                                   (slice(None, None, None),)),
 ('getitem-08a6b8d2e456920c909b54d9db4e8640', 1): (<function _zero_getitem at 0x11ec9d6c0>,
                                                   ('from-json-ccb017520b83fc52095908cdff28ce65', 1),
                                                   slice(0, 5, 1),
                                                   (slice(None, None, None),))}
```
i.e., one of the output partitions has no corresponding input partition.

douglasdavis:
Hmm looks like the optimization is culling to the point of having only 1 partition (this was something that was added recently, correct?) but the fact that a repartition layer exists in the graph is making it such that we end up with outputs from the repartition (low level) layer in the graph expecting culled inputs

lgray:
@douglasdavis yes, for the column optimization pass - if we are to scale to actual analysis in HEP - we do the graph optimization on a mocked partition with the same HLG structure. Otherwise the number of calls to awkward / size of materialized graph / resulting execution time is far to much for an optimization step.

lgray:
Means we probably just need a `MockableRepartitionLayer` in dask-awkward so we can do what we need.

martindurant:
Or rather, a generalised mock-materialised , since this version no longer makes a new layer type. That would cover slice too.

lgray:
Is this good to give another spin to from my side?

martindurant:
Yes please. We should have it only plucking the first partition during optimisation

lgray:
Cool - I have some other things to take care of first but I'll get to this.

douglasdavis:
as of commit 42b5d3a optimization is successfully _running_ (that is, no exception raised), but it's being overzelous. I'm investigating

douglasdavis:
OK it's actually not as bad as I thought.

Two cases:

bad:

```python
>>> new = ds.repartition(5)
>>> dak.necessary_columns(new.muons.pt + new.run)
{'read-parquet-422b6358da1d54c615e6af00a19b4dae': ['MET.pt',
  'run',
  'event',
  'muons.phi',
  'gen.pt',
  'gen.pdgId',
  'luminosityBlock',
  'muons.charge',
  'gen.phi',
  'muons.eta',
  'muons.mass',
  'MET.phi',
  'gen.eta',
  'muons.pt']}
```

good:

```python
>>> dak.necessary_columns((ds.muons.pt + ds.run).repartition(5))
{'read-parquet-9754d11080a65942dd864e05f2eda183': ['run', 'muons.pt']}
```

the order of the repartition call matters. I haven't put any thought into possible solutions here to make sure we don't touch everything in the "bad" case.

PR #255, opened 2023-05-03
tidy: remove unnecessary loop over layers for mocking

lgray:
Already done once!


martindurant:
Looks like you are right. Was this costing time?

lgray:
No, not taking any time, but could lead to bug in the future. Better to remove.

PR #258, opened 2023-05-10
fix: import `typetracer_with_report` from L2 module

agoose77:
> **Warning**
> This will require a pin on a future release of awkward (2.2.1). 

We're gradually exporting any clearly-needed features from Awkward into a public (L2) module, in this instance `ak.typetracer`. This PR makes use of these symbols.


martindurant:
Looks fine, but of course we must wait

PR #264, opened 2023-05-13
fix: only run linear-chains optimization on AwkwardBlockwiseLayers

lgray:
This was clobbering linear chains from dask.array, like `x * 0.8` as seen over in the tests of: https://github.com/dask-contrib/dask-histogram/pull/82

Have verified locally that this fixes the test failures for that otherwise unrelated PR, while also not spoiling the optimization for dak layers.


lgray:
Yes, I still see the full benefit (thousands of layers removed in the torture test).

PR #267, opened 2023-05-19
feat: add comprehensive `axis=0` and `axis=None` reduction

agoose77:
- ~[x] Support positional reducers~
- [x] Support trivial value reducers
- [x] Support reducers with different combining operations e.g. `ak.count`

### DIFF
--- a/src/dask_awkward/lib/core.py
+++ b/src/dask_awkward/lib/core.py
@@ -1478,7 +1478,6 @@ def _from_iter(obj):
     representation of the input iterable-of-typetracers.
 
     """
-    breakpoint()
     try:
         return ak.from_iter(obj)
     except (ValueError, TypeError):
--- a/src/dask_awkward/lib/core.py
+++ b/src/dask_awkward/lib/core.py
@@ -1539,9 +1539,18 @@ def _next_chunk_partition_offset(chunk: ak.Array, axis: int | None):
 def _chunk_reducer_positional(
     chunk: ak.Array, axis: int | None, *, reducer: Callable
 ) -> PartialPositionalReductionType:
+    # TODO: this is private Awkward code. We should figure out how to export it
+    # if needed
+    layout, = ak._do.remove_structure(
+        ak.to_layout(chunk),
+        flatten_records=False,
+        drop_nones=False,
+        keepdims=True,
+        allow_records=False,
+    )
+    chunk = ak.Array(layout, behavior=chunk.behavior)
+
     index = reducer(chunk, axis=axis, keepdims=True, mask_identity=True)
-    # TODO: chunk needs to be flattened in the same way that reducer(chunk, axis=None, keepdims=True)
-    # does in order to preserve the index-ability
     # Adjust the index to be absolute w.r.t the original array
     return index, chunk[index], _next_chunk_partition_offset(chunk, axis)
 
--- a/src/dask_awkward/lib/core.py
+++ b/src/dask_awkward/lib/core.py
@@ -6,9 +6,9 @@ import logging
 import operator
 import sys
 import warnings
-from itertools import accumulate, islice
 from collections.abc import Callable, Hashable, Mapping, Sequence
 from functools import cached_property, partial
+from itertools import accumulate, islice
 from numbers import Number
 from typing import TYPE_CHECKING, Any, Literal, TypeVar
 
@@ -1498,7 +1498,11 @@ PartialReductionType = ak.Array
 
 
 def _chunk_reducer_trivial(
-    chunk: ak.Array | PartialReductionType, axis: int | None, *, reducer: Callable, mask_identity: bool
+    chunk: ak.Array | PartialReductionType,
+    axis: int | None,
+    *,
+    reducer: Callable,
+    mask_identity: bool,
 ) -> PartialReductionType:
     return reducer(chunk, keepdims=True, axis=axis, mask_identity=mask_identity)
 
--- a/src/dask_awkward/lib/core.py
+++ b/src/dask_awkward/lib/core.py
@@ -1559,6 +1559,16 @@ def _prepare_axis_none_chunk(chunk: ak.Array) -> ak.Array:
     return ak.Array(layout, behavior=chunk.behavior)
 
 
+def _ragged_index(container: ak.Array, index: ak.Array) -> ak.Array:
+    def apply(layout: ak.contents.Content, **kwargs) -> ak.contents.Content | None:
+        if layout.is_numpy:
+            flat_container = ak.ravel(container, highlevel=False)
+            return flat_container[layout]
+
+    # Take list structure from `index`
+    return ak.transform(apply, index)
+
+
 # Exterior reductions need starts!
 def _chunk_reducer_positional(
     chunk: ak.Array, is_axis_none: bool, *, reducer: Callable
@@ -1567,7 +1577,11 @@ def _chunk_reducer_positional(
         chunk, axis=-1 if is_axis_none else 0, keepdims=True, mask_identity=True
     )
     # Adjust the index to be absolute w.r.t the original array
-    return index, chunk[index], _next_chunk_partition_offset(chunk, is_axis_none)
+    return (
+        index,
+        _ragged_index(chunk, index),
+        _next_chunk_partition_offset(chunk, is_axis_none),
+    )
 
 
 # All reductions need concatenation
@@ -1598,8 +1612,8 @@ def _tree_node_reducer_positional(
     )
     # Indices are already absolute!
     return (
-        partial_index[final_index_index],
-        partial_value[final_index_index],
+        _ragged_index(partial_index, final_index_index),
+        _ragged_index(partial_value, final_index_index),
         partial_length,
     )
 
@@ -1616,7 +1630,7 @@ def _finalise_reducer_positional(
     final_index_index = reducer(
         partial_value, axis=-1, keepdims=True, mask_identity=True
     )
-    final_index = partial_index[final_index_index]
+    final_index = _ragged_index(partial_index, final_index_index)
 
     # The return type of positional reducers is unaffected by option-type values
     # i.e., we can safely remove the mask if it is not required
@@ -1712,15 +1726,14 @@ def non_trivial_reduction(
     # For axis=None, let's remove the structure to make the internal operations
     # axis=-1
     if axis is None:
-        prepared_array = map_partitions(
-            _prepare_axis_none_chunk, array, meta=empty_typetracer()
-        )
+        prepared_array = map_partitions(_prepare_axis_none_chunk, array)
     else:
         prepared_array = array
 
     chunked_result = map_partitions(
         chunked_fn,
         prepared_array,
+        # TODO: fix this meta
         meta=empty_typetracer(),
     )
 
--- a/src/dask_awkward/lib/core.py
+++ b/src/dask_awkward/lib/core.py
@@ -8,7 +8,7 @@ import sys
 import warnings
 from collections.abc import Callable, Hashable, Mapping, Sequence
 from functools import cached_property, partial
-from itertools import accumulate, islice
+from itertools import accumulate
 from numbers import Number
 from typing import TYPE_CHECKING, Any, Literal, TypeVar
 
--- a/src/dask_awkward/lib/core.py
+++ b/src/dask_awkward/lib/core.py
@@ -1586,12 +1586,12 @@ def _concat_reducer_positional(
 
 # Interior reductions don't need starts
 def _tree_node_reducer_positional(
-    partial: PartialPositionalReductionType, *, reducer: Callable
+    partial: PartialPositionalReductionType, is_axis_none: bool, *, reducer: Callable
 ) -> PartialPositionalReductionType:
     # `partial` comes from `concat`
     partial_index, partial_value, partial_length = partial
     final_index_index = reducer(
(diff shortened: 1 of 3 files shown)


lgray:
@iasonkrom when this one is passing tests give it a try!

src/dask_awkward/lib/core.py
```diff
@@ -1493,6 +1494,187 @@ def _from_iter(obj):
...
+        concat_fn = _concat_reducer_positional
+        finalize_fn = _finalise_reducer_positional
+
+        # TODO HACK!
+        from dask_awkward import from_awkward
+
+        if not array.known_divisions:
+            array.eager_compute_divisions()
+        starts = ak.Array(array.divisions)[:-1]
+        tree_node_args = [from_awkward(starts, npartitions=len(starts))]
```
agoose77, line 1601: @martindurant what's the best way of doing this properly — starmapping a sequence in `map_partitions`? Do we need to extend `map_partitions` to handle this, or perhaps support `dask.array` in `map_partitions`?

martindurant: > support dask.array in map_partitions

That sounds like a bad idea

src/dask_awkward/lib/core.py
```diff
@@ -1478,6 +1478,7 @@ def _from_iter(obj):
     representation of the input iterable-of-typetracers.
 
     """
+    breakpoint()
```
agoose77, line 1481:
```suggestion
```

src/dask_awkward/lib/core.py
agoose77: Perhaps I should circumvent `map_partitions` and do this by hand then?

agoose77:
@martindurant @douglasdavis this PR is mostly complete from a functionality perspective. I would appreciate a nudge as to where to take it next. Some thoughts:
- `axis=0` and `axis=None` reduction are very similar. We can improve the `axis=None` case to handle all the reducer arguments that `axis=0` supports here. How should we combine these functions? The trivial case is to first `ak.flatten` the array, which is ultimately what happens when one passes `axis=None` to an `ak.XXX` reducer[^1], and then pass this to `axis=0` reduction.
- I've not tacked `std` or `var`, because these require custom reduction and combining logic. I wonder whether the helper `axis_0_reduction` should become thinner so that we can just pass in the various tree-reduction helpers for these special cases directly?

chrispap95:
Hi, I have been trying to run code that looks like: `dak.sum(events.weight, axis=0)`. `events.weight` is just a simple 1d array of floats. @lgray suggested that I give this PR a try so I am reporting here what I saw just in case it helps with development and debugging. 

My versions are:
```
awkward.__version__='2.2.1'
dask_awkward.__version__='2023.6.1.dev29+g1f3a6a5'
```

When try to run `dak.sum(events.weight, axis=0).compute()`  I get the following error message:
```
  File "/uscms_data/d1/chpapage/fastjet_test_releases/env/lib/python3.9/site-packages/awkward/highlevel.py", line 950, in __getitem__
    out = self._layout[where]
  File "/uscms_data/d1/chpapage/fastjet_test_releases/env/lib/python3.9/site-packages/awkward/contents/content.py", line 536, in __getitem__
    return self._getitem(where)
  File "/uscms_data/d1/chpapage/fastjet_test_releases/env/lib/python3.9/site-packages/awkward/contents/content.py", line 553, in _getitem
    return self._getitem_field(where)
  File "/uscms_data/d1/chpapage/fastjet_test_releases/env/lib/python3.9/site-packages/awkward/contents/recordarray.py", line 471, in _getitem_field
    return self.content(where)
  File "/uscms_data/d1/chpapage/fastjet_test_releases/env/lib/python3.9/site-packages/awkward/contents/recordarray.py", line 406, in content
    out = self.form_cls.content(self, index_or_field)
  File "/uscms_data/d1/chpapage/fastjet_test_releases/env/lib/python3.9/site-packages/awkward/forms/recordform.py", line 144, in content
    index = self.field_to_index(index_or_field)
  File "/uscms_data/d1/chpapage/fastjet_test_releases/env/lib/python3.9/site-packages/awkward/contents/recordarray.py", line 400, in field_to_index
    return self.form_cls.field_to_index(self, field)
  File "/uscms_data/d1/chpapage/fastjet_test_releases/env/lib/python3.9/site-packages/awkward/forms/recordform.py", line 123, in field_to_index
    raise ak._errors.FieldNotFoundError(
awkward._errors.FieldNotFoundError: no field 'genEventSumw' in record with 1 fields

The above exception was the direct cause of the following exception:

  File "/uscms_data/d1/chpapage/fastjet_test_releases/env/lib/python3.9/site-packages/coffea/processor/executor.py", line 1720, in _work_function
    to_compute = processor_instance.process(events)
  File "/uscms_data/d3/chpapage/fastjet_test_releases/SUEPCoffea_dask/workflows/GenSumWeightExtract.py", line 15, in process
    output = ak.sum(events.genEventSumw, axis=0).compute()
  File "/uscms_data/d1/chpapage/fastjet_test_releases/env/lib/python3.9/site-packages/dask/base.py", line 310, in compute
  File "/uscms_data/d1/chpapage/fastjet_test_releases/env/lib/python3.9/site-packages/dask/base.py", line 595, in compute
  File "/uscms_data/d1/chpapage/fastjet_test_releases/env/lib/python3.9/site-packages/dask/threaded.py", line 89, in get
    results = get_async(
  File "/uscms_data/d1/chpapage/fastjet_test_releases/env/lib/python3.9/site-packages/dask/local.py", line 511, in get_async
    raise_exception(exc, tb)
  File "/uscms_data/d1/chpapage/fastjet_test_releases/env/lib/python3.9/site-packages/dask/local.py", line 319, in reraise
    raise exc
  File "/uscms_data/d1/chpapage/fastjet_test_releases/env/lib/python3.9/site-packages/dask/local.py", line 224, in execute_task
  File "/uscms_data/d1/chpapage/fastjet_test_releases/env/lib/python3.9/site-packages/dask/core.py", line 121, in _execute_task
  File "/uscms_data/d1/chpapage/fastjet_test_releases/env/lib/python3.9/site-packages/dask/optimization.py", line 992, in __call__
  File "/uscms_data/d1/chpapage/fastjet_test_releases/env/lib/python3.9/site-packages/dask/core.py", line 151, in get
  File "/uscms_data/d1/chpapage/fastjet_test_releases/env/lib/python3.9/site-packages/dask/core.py", line 121, in _execute_task
  File "/uscms_data/d1/chpapage/fastjet_test_releases/env/lib/python3.9/site-packages/dask/core.py", line 121, in <genexpr>
  File "/uscms_data/d1/chpapage/fastjet_test_releases/env/lib/python3.9/site-packages/dask/core.py", line 121, in _execute_task
  File "/uscms_data/d1/chpapage/fastjet_test_releases/env/lib/python3.9/site-packages/awkward/highlevel.py", line 962, in __getitem__
    return wrap_layout(out, self._behavior, allow_other=True)
  File "/uscms_data/d1/chpapage/fastjet_test_releases/env/lib/python3.9/site-packages/awkward/_errors.py", line 56, in __exit__
    self.handle_exception(exception_type, exception_value)
  File "/uscms_data/d1/chpapage/fastjet_test_releases/env/lib/python3.9/site-packages/awkward/_errors.py", line 71, in handle_exception
    raise self.decorate_exception(cls, exception)
awkward._errors.FieldNotFoundError: no field 'weight' in record with 1 fields

This error occurred while attempting to slice

    <NanoEventsArray [{run: 1}, {...}, ..., {...}, {run: 1}] type='117 * event'>

with

    'weight'
```
This is the same as the error message from previous versions.
When I try `dak.sum(events.weight.compute(), axis=0)` I get:
```
  File "/uscms_data/d1/chpapage/fastjet_test_releases/env/lib/python3.9/site-packages/coffea/processor/executor.py", line 1720, in _work_function
    to_compute = processor_instance.process(events)
  File "/uscms_data/d3/chpapage/fastjet_test_releases/SUEPCoffea_dask/workflows/GenSumWeightExtract.py", line 15, in process
    output = ak.sum(events.genEventSumw.compute(), axis=0)
  File "/uscms_data/d1/chpapage/fastjet_test_releases/env/lib/python3.9/site-packages/dask_awkward/lib/reducers.py", line 414, in sum
    return non_trivial_reduction(
  File "/uscms_data/d1/chpapage/fastjet_test_releases/env/lib/python3.9/site-packages/dask_awkward/lib/core.py", line 1733, in non_trivial_reduction
    chunked_result = map_partitions(
  File "/uscms_data/d1/chpapage/fastjet_test_releases/env/lib/python3.9/site-packages/dask_awkward/lib/core.py", line 1469, in map_partitions
    npartitions=deps[0].npartitions,
IndexError: list index out of range
```
I am not sure if calling compute before doing `dak.sum()` makes any sense at all, but this error message is different than before. (it used to throw a `NotImplemented` exception)

agoose77:
Hmm, now to get `axis=0` positional reducers working!

agoose77:
Jim and I had a meeting to discuss the challenge of handling positional reducers in a low-debt, performant manner. We decided to implement only value-based reducers to start (in our meeting we discussed dropping `axis=0`, but I think the same argument applies to `axis=None`) as a first implementation.

jpivarski:
`axis=0` is more exotic/corner case than `axis=None`, particularly since `axis=None` is the default. We'll likely see requests for `axis=None` before we see any requests for `axis=0`.

agoose77:
Right. When I say "niche", I mean — likelihood of _wanting_ to `axis=None` for large datasets ;)

src/dask_awkward/lib/core.py
```diff
@@ -1592,6 +1592,185 @@ def _from_iter(obj):
...
+
+    # Build graph
+    prepared_array_graph = prepared_array.__dask_graph__()
+    layers = {name_finalize: dftr, name_chunked: chunked, **prepared_array_graph.layers}
+    dependencies = {
+        name_finalize: {name_chunked},
+        name_chunked: {prepared_array.name},
+        **prepared_array_graph.dependencies,
+    }
+    graph = HighLevelGraph(layers, dependencies)
```
agoose77, line 1760: @douglasdavis would you be able to cast your eyes over this and check that it is reasonable?

douglasdavis: yep will do!

douglasdavis: @agoose77 this is looking good to me.

One thing I want to clarify: It looks to me that you're building the graph with `HighLevelGraph(...)` instead of `HighLevelGraph.from_collections` because `chunked` is a layer instead of a collection. Is there a reason you're making `chunked` via `partitionwise_layer` instead of via `map_partitions` (which will just make a new collection that you can pass to `DataFrameTreeReduction`)?

src/dask_awkward/lib/core.py
```diff
@@ -1593,6 +1593,185 @@ def _from_iter(obj):
...
+
+
+def _chunk_reducer_non_positional(
+    chunk: ak.Array | PartialReductionType,
+    is_axis_none: bool,
+    *,
+    reducer: Callable,
+    mask_identity: bool,
+) -> PartialReductionType:
+    return reducer(
```
douglasdavis, line 1606: this function is failing with typetracer task graphs (so the necessary columns opt fails).

repro:
```python
>>> import dask
>>> with dask.config.set({"awkward.optimization.on-fail": "raise"}):
...     dak.min(ds.muons.pt, axis=None).compute()
```
pretty long message, but here's the end:

```
ValueError: arrays to concatenate do not have the same depth for negative axis=-1

This error occurred while calling

    ak.concatenate(
        arrays = [<Array-typetracer [...] type='1 * 1 * ?float64'>, ('min-pre...
        axis = -1
        mergebool = True
        highlevel = True
        behavior = None
    )
```

when #277 is in we should be able to see things like this in tests :)

Also errors with `axis=0`

```
ValueError: cannot use non-negative axis on a nested list structure of variable depth (negative axis counts from the leaves of the tree; non-negative from the root)

This error occurred while calling

    ak.min(
        array = <Array-typetracer [...] type='15 * union[var * ?float64, st...'>
        axis = 0
        keepdims = True
        initial = None
        mask_identity = True
```

src/dask_awkward/lib/core.py
agoose77: Technically there's no reason for this PR as it stands. I originally needed this because some of the tree reduction steps returned tuples rather than Arrays. I think on-balance I'd remove this, though; the commits in this PR include the previous stab towards positional reducers should we see fit to restore them.

src/dask_awkward/lib/core.py
agoose77: Fab! I can fix this.

agoose77: Could you provide a reproducer that I can run?

agoose77:
@douglasdavis LGTM!

src/dask_awkward/lib/core.py
```diff
@@ -1601,6 +1601,176 @@ def _from_iter(obj):
...
+    # TODO: this is private Awkward code. We should figure out how to export it
+    # if needed
+    (layout,) = ak._do.remove_structure(
+        ak.to_layout(chunk),
+        flatten_records=False,
+        drop_nones=False,
+        keepdims=True,
+        allow_records=False,
+    )
+    return ak.Array(layout, behavior=chunk.behavior)
```
agoose77, line 1655: I'll need to export this, but we can release without it.

douglasdavis:
Awesome, thanks!

PR #270, opened 2023-05-26
feat: add `head`, `sample` and axis 0 slicing

martindurant:
prelude to slicing on axis 0

### DIFF
--- a/src/dask_awkward/lib/core.py
+++ b/src/dask_awkward/lib/core.py
@@ -32,8 +32,8 @@ from dask.utils import IndexCallable, funcname, key_split
 from numpy.lib.mixins import NDArrayOperatorsMixin
 from tlz import first
 
-from dask_awkward.layers import AwkwardBlockwiseLayer
-from dask_awkward.lib.optimize import all_optimizations, mock_materialized_layer
+from dask_awkward.layers import AwkwardBlockwiseLayer, AwkwardMaterializedLayer
+from dask_awkward.lib.optimize import all_optimizations
 from dask_awkward.typing import AwkwardDaskCollection
 from dask_awkward.utils import (
     DaskAwkwardNotImplemented,
@@ -705,7 +705,7 @@ class Array(DaskMethodsMixin, NDArrayOperatorsMixin):
         dsk = {(name, i): tuple(key) for i, key in enumerate(new_keys)}
         graph = HighLevelGraph.from_collections(
             name,
-            dsk,
+            AwkwardMaterializedLayer(dsk, previous_layer_name=self.name),
             dependencies=[self],
         )
 
@@ -864,7 +864,7 @@ class Array(DaskMethodsMixin, NDArrayOperatorsMixin):
         }
         hlg = HighLevelGraph.from_collections(
             name,
-            dsk,
+            AwkwardMaterializedLayer(dsk, previous_layer_name=self.name),
             dependencies=[partition],
         )
         if isinstance(new_meta, ak.Record):
@@ -925,8 +925,11 @@ class Array(DaskMethodsMixin, NDArrayOperatorsMixin):
                 (self.divisions[i + 1] - self.divisions[i]) // step + divisions[-1]
             )
             remainder = remainder % step
-        hlg = HighLevelGraph.from_collections(name, dask, dependencies=[self])
-        hlg.layers[name].mock = mock_materialized_layer(hlg.layers[name], self.name)
+        hlg = HighLevelGraph.from_collections(
+            name,
+            AwkwardMaterializedLayer(dask, previous_layer_name=self.name),
+            dependencies=[self],
+        )
         return new_array_object(
             hlg,
             name,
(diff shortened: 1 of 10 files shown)


martindurant:
@douglasdavis , I reckon this is a pretty easy one to allow in quickly. I don't know how effective it would be at culling massive graphs, but extra work on that doesn't belong here.

douglasdavis:
yep looks good to me!

martindurant:
@douglasdavis , added an attempt at mocking a materialised layer for axis 0 slicing, which makes column optimisations pass. It's probably brittle! We should come up with a set of tests first, then fix problems. This will also apply to repartition, though.

Also, we need to decide whether head() and sample() should be dak functions only, as @jpivarski suggested, or if we keep one or both as they are. I think having them as functions is fine, but there is an argument for head() (only) as a method also.

douglasdavis:
> @douglasdavis , added an attempt at mocking a materialised layer for axis 0 slicing, which makes column optimisations pass. It's probably brittle! We should come up with a set of tests first, then fix problems. This will also apply to repartition, though.

great! I'll take a look and test it out

> Also, we need to decide whether head() and sample() should be dak functions only, as @jpivarski suggested, or if we keep one or both as they are. I think having them as functions is fine, but there is an argument for head() (only) as a method also.

I think we decided to make `head` a method (the precedent set by dataframe is huge here). I would argue that `sample` should probably be treated in the same way (`pandas.DataFrame.sample` being a method); though NumPy does have a function for random sampling (`numpy.random.choice`)

martindurant:
`assert 0.3 < (len(arr) / len(daa)) < 0.7` - I should calculate what the actual chance of this failing is, since it's using random probabilities. Perhaps even better, I should change sample() to allow to set the numpy random seed? I'm not sure that will work when the tasks can be running simultaneously.

src/dask_awkward/lib/core.py
```diff
@@ -872,6 +872,69 @@ def _getitem_outer_int(self, where: int | tuple[Any, ...]) -> Any:
...
+                rest,
+            )
+            outpart += 1
+            remainder += (self.divisions[i + 1] - self.divisions[i]) % step
+            divisions.append(
+                (self.divisions[i + 1] - self.divisions[i]) // step + divisions[-1]
+            )
+            remainder = remainder % step
+        hlg = HighLevelGraph.from_collections(name, dask, dependencies=[self])
+        hlg.layers[name].mock = mock_materialized_layer(hlg.layers[name], self.name)
```
douglasdavis, line 929: this attach-a-free-function-as-instance-method is clever! I was thinking we would need to subclass `MaterialzedLayer` and add a `mock` method

martindurant: I was feeling burned from making layers, but yes that would work of course. It's not actually a method, just an attribute here (making a true dynamic method is possible too).

martindurant: done

martindurant:
(when this is in, repartition should use the same mock)

src/dask_awkward/lib/inspect.py
```diff
@@ -78,3 +79,27 @@ def necessary_columns(*args: Any, traverse: bool = True) -> dict[str, list[str]]
...
+
+
+def sample(arr, factor: int | None = None, probability: float | None = None):
+    """Decimate the data to a smaller number of rows
+
+    Must give either factor or probability
+
+    Parameters
+    ----------
+    arr: dask-awkward array
```
douglasdavis, line 91:
```suggestion
    arr : dask_awkward.Array
        Array collection to sample
```

src/dask_awkward/lib/inspect.py
```diff
@@ -78,3 +79,27 @@ def necessary_columns(*args: Any, traverse: bool = True) -> dict[str, list[str]]
...
+    factor: if given, every Nth row will be kept. The counting restarts for each
+        partition, so reducing the row count by an exact factor is not guaranteed
```
douglasdavis, line 93:
```suggestion
    factor : int, optional
        if given, every Nth row will be kept. The counting restarts for each
        partition, so reducing the row count by an exact factor is not guaranteed
```

src/dask_awkward/lib/inspect.py
```diff
@@ -78,3 +79,27 @@ def necessary_columns(*args: Any, traverse: bool = True) -> dict[str, list[str]]
...
+    probability: a number between 0 and 1, giving the chance of any particular
+        row surviving. For instance, for probability=0.1, roughly 1-in-10
+        rows will remain.
```
douglasdavis, line 96:
```suggestion
    probability : float, optional
        a number between 0 and 1, giving the chance of any particular
        row surviving. For instance, for probability=0.1, roughly 1-in-10
        rows will remain.
```

douglasdavis (commented):
a couple of docstring nitpicks but other than that this looks good to me

douglasdavis:
I know there exists a sphinx plugin to handle this- I'd like to eventually get all sphinx types from the typehints, it's just not high on my todo list right now :)

douglasdavis:
Do you have anything else in mind for this PR or good to go?

martindurant:
> Do you have anything else in mind for this PR or good to go?

I feel like the mock will need revisiting, but maybe we can make more extensive test cases for it out of real workflows after merging. I don't have a clear idea of how it can be improved now, since I don't know how it will break.

PR #272, opened 2023-06-02
Fix mask with from_awkward

jrueb:
Optimization so far failed when applying `dak.mask` when that mask that is applied was generated using something like `dak.from_awkward`.

The `_get_column_reports` method was not consistently dealing with the fact that `AwkwardInputLayer.mock` also returns a type tracer report.


jrueb:
I marked this as draft because there is another bug raising an error during optimization. The bug happens here https://github.com/dask-contrib/dask-awkward/blob/3ce8e4ef128f262509097d542a02061778528aa8/src/dask_awkward/lib/core.py#L488-L493
This sets the type of the last layer in the graph to the output type. The the case of `mask` and `from_awkward`, this will overwrite the type of the mask (which is supposed to be something like `## * bool`) upon which awkward will complain that the mask needs to be bool or that it has the wrong structure.
I would also like to fix this but I cannot make sense of this code. What layer is supposed to have its `_meta` changed and why?

douglasdavis:
Hi @jrueb, thanks for the contribution! I'll dive in to looking at this at some point this week.

douglasdavis:
This looks like a hole in our interface with the way we've defined the `mock` method for layers (it has grown a bit organically as needed). We should probably make it such that any layer class that implements the `mock` method needs to return a two element tuple from that method.

@martindurant what do you think? Should `AwkwardMaterializedLayer.mock` return `(new_materialized_layer, None)` (instead of just `new_materialized_layer`)? I'm thinking we need to define `mock`'s return signature to be:

```python
class MyLayer:
    ...

    def mock(self) ->tuple[MyLayer, TypeTracerReport | None]:
        ...
```

And another question for @martindurant, do you happen to quickly recall the purpose of the code @jrueb quoted in his comment https://github.com/dask-contrib/dask-awkward/pull/272#issuecomment-1573299479? (I'm struggling to quickly remember, still thinking about it/tracking it down, but wanted to pick your brain just in case)

douglasdavis:
Here's what I have in mind

```diff
diff --git a/src/dask_awkward/layers/layers.py b/src/dask_awkward/layers/layers.py
index ad37585..e504a3a 100644
--- a/src/dask_awkward/layers/layers.py
+++ b/src/dask_awkward/layers/layers.py
@@ -28,7 +28,7 @@ class AwkwardBlockwiseLayer(Blockwise):
         nb = layer.numblocks
         layer.numblocks = {k: tuple(1 for _ in v) for k, v in nb.items()}
         layer.__dict__.pop("_dims", None)
-        return layer
+        return layer, None
 
     def __getstate__(self) -> dict:
         d = self.__dict__.copy()
@@ -211,7 +211,7 @@ class AwkwardMaterializedLayer(MaterializedLayer):
                 else v
                 for v in task
             )
-            return MaterializedLayer({(name, 0): task})
+            return MaterializedLayer({(name, 0): task}), None
 
         # failed to cull during column opt
-        return self
+        return self, None
diff --git a/src/dask_awkward/lib/optimize.py b/src/dask_awkward/lib/optimize.py
index bb8bfcd..dd0f83b 100644
--- a/src/dask_awkward/lib/optimize.py
+++ b/src/dask_awkward/lib/optimize.py
@@ -348,11 +348,7 @@ def _get_column_reports(dsk: HighLevelGraph) -> dict[str, Any]:
             layers[name], report = lay.mock()
             reports[name] = report
         elif hasattr(lay, "mock"):
-            mocking = lay.mock()
-            if isinstance(mocking, tuple):
-                layers[name], reports[name] = mocking
-            else:
-                layers[name] = mocking
+            layers[name], _ = lay.mock()
 
     for name in _ak_output_layer_names(dsk):
         layers[name] = _mock_output(layers[name])
```

martindurant:
Yes to the consistent interface, for sure.

Not certain about that code, but every layer we control should ideally have a _meta, and I think in this case it is being set for the first time. I don't follow why the DaskAwkwardArray's meta should disagree with the meta of its output layer.

douglasdavis:
I've just pushed a commit that fixes the interface; we still need to address the layer meta issue.

douglasdavis:
I'd like to get this in because it fixes the `mock` interface. Regarding the failed columns optimization-- @jrueb would you mind opening a dedicated issue for that?

PR #277, opened 2023-06-07
refactor: harden tests by adjusting current array-providing fixtures

douglasdavis:
This PR is an experiment in hardening our existing tests.

Many of our tests are performed by using the `daa` and `caa` fixtures.

- `daa` is a dask-awkward `Array` that is created via `dak.from_json`. 
- `caa` is a concrete awkward `Array` created from `ak.from_json` followed by `ak.concatenate`. 

the structure of the array is:

```
n * {
    points: var * {
        x: int64,
        y: int64
    }
}
```
(a record array with a single entry (`points`) which each contain a variable number of `(x, y)` pairs)

This makes it easy to write a test like:

```python
def test_sum_axis1(daa, caa):
    assert_eq(
        dak.sum(daa.points.x, axis=1),
        dak.sum(caa.points.x, axis=1),
    )
```

Currently dask-awkward JSON reads do not support the `project_columns` interface (though it is something on our todo list).

So `daa.points.x` will not be optimized to read only the `points.x` field.

In this PR I've made the `daa` fixture originate via a `dak.from_parquet` call.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -57,6 +57,7 @@ docs = [
 test = [
   "aiohttp",
   "distributed",
+  "pandas",
   "pyarrow",
   "pytest >=6.0",
   "pytest-cov >=3.0.0",
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -37,13 +37,15 @@ Homepage = "https://github.com/dask-contrib/dask-awkward"
 "Bug Tracker" = "https://github.com/dask-contrib/dask-awkward/issues"
 
 [project.optional-dependencies]
+io = [
+  "aiohttp",
+  "pyarrow",
+]
 complete = [
   "aiohttp",
   "pyarrow",
-  "pytest >=6.0",
-  "pytest-cov >=3.0.0",
-  "requests >=2.27.1",
 ]
+# `docs` and `test` are separate from user installs
 docs = [
   "dask-sphinx-theme >=3.0.2",
   "pyarrow",
@@ -51,10 +53,6 @@ docs = [
   "pytest >=6.0",
   "pytest-cov >=3.0.0",
   "requests >=2.27.1",
-]
-io = [
-  "aiohttp",
-  "pyarrow",
 ]
 test = [
   "aiohttp",
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -27,7 +27,6 @@ classifiers = [
   "Topic :: Scientific/Engineering",
 ]
 dependencies = [
-  "numpy <=1.24.3",
   "awkward >=2.2.2",
   "dask >=2023.04.0",
 ]
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -27,6 +27,7 @@ classifiers = [
   "Topic :: Scientific/Engineering",
 ]
 dependencies = [
+  "numpy <=1.24.3",
   "awkward >=2.2.2",
   "dask >=2023.04.0",
 ]
--- a/src/dask_awkward/lib/core.py
+++ b/src/dask_awkward/lib/core.py
@@ -2062,10 +2062,17 @@ def typetracer_from_form(form: Form) -> ak.Array:
 
 
 def make_unknown_length(array: ak.Array) -> ak.Array:
-    return ak.Array(
-        ak.to_backend(
-            array,
-            "typetracer",
-            highlevel=False,
-        ).to_typetracer(forget_length=True)
-    )
+    """Make any highlevel Array a highlevel typetracer Array with unknown length.
+
+    Parameters
+    ----------
+    array : ak.Array
+        Array of interest
+
+    Returns
+    -------
+    ak.Array
+        Highlevel typetracer Array with unknown length.
+
+    """
+    return ak.Array(ak.to_layout(array).to_typetracer(forget_length=True))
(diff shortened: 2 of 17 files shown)


douglasdavis:
Something that's popping up in a number of tests: the `_unproject_layout` function's final `else` case raises due to an unexpected combination. We're getting a case where the form is `None` and the layout is a `RecordArray` content.

```
AssertionError: unexpected combination: <class 'NoneType'> and <class 'awkward.contents.recordarray.RecordArray'>
```

a test to trigger this (not the only one):

```python
$ pytest tests/test_inspect.py::test_sample
```

full list of failures due to that assertion:

```
FAILED tests/test_core.py::test_compute_typetracer - AssertionError: unexpected combination: <class 'NoneType'> and <class 'awkward.contents.recordarray.RecordArray'>
FAILED tests/test_core.py::test_scalar_to_delayed[False] - AssertionError: unexpected combination: <class 'NoneType'> and <class 'awkward.contents.recordarray.RecordArray'>
FAILED tests/test_core.py::test_optimize_chain_single - AssertionError: unexpected combination: <class 'NoneType'> and <class 'awkward.contents.recordarray.RecordArray'>
FAILED tests/test_getitem.py::test_single_int - AssertionError: unexpected combination: <class 'NoneType'> and <class 'awkward.contents.recordarray.RecordArray'>
FAILED tests/test_getitem.py::test_single_partition - AssertionError: unexpected combination: <class 'NoneType'> and <class 'awkward.contents.recordarray.RecordArray'>
FAILED tests/test_inspect.py::test_sample - AssertionError: unexpected combination: <class 'NoneType'> and <class 'awkward.contents.recordarray.RecordArray'>
FAILED tests/test_io.py::test_to_and_from_delayed[False] - AssertionError: unexpected combination: <class 'NoneType'> and <class 'awkward.contents.recordarray.RecordArray'>
FAILED tests/test_io.py::test_to_delayed[False] - AssertionError: unexpected combination: <class 'NoneType'> and <class 'awkward.contents.recordarray.RecordArray'>
FAILED tests/test_io.py::test_to_bag - AssertionError: unexpected combination: <class 'NoneType'> and <class 'awkward.contents.recordarray.RecordArray'>
FAILED tests/test_io.py::test_to_dataframe[False] - AssertionError: unexpected combination: <class 'NoneType'> and <class 'awkward.contents.recordarray.RecordArray'>
FAILED tests/test_structure.py::test_full_like[datetime64] - AssertionError: unexpected combination: <class 'NoneType'> and <class 'awkward.contents.recordarray.RecordArray'>
```

simple repro:

```python
import dask_awkward as dak
import awkward as ak

arr = ak.Array(
    [
        {"x": [1,2,3], "y": 0},
        {"x": [4,5], "y": 1},
    ]
)
ak.to_parquet(arr, "test.parquet")
darr = dak.from_parquet("test.parquet")
samp = dak.sample(darr, probability=0.8)
samp.compute()
```
I'll see if I can track it down some, but tagging @agoose77 for visibility

agoose77:
Thanks @douglasdavis. `unproject_layout` always needs a `Form`, so I would expect the caller to ensure that it passes in a valid `Form` object before calling the function?

douglasdavis:
thanks @agoose77 that was quick! Yes I tracked down the issue. I was thinking we perhaps hit a bad combo in the recursion, but I was wrong. We are currently explicitly passing  a `None` form at an explicit call site in cases _when the column optimization fails_. Just pushed a simple fix to bypass running _unproject_columns when the column opt fails.

pyproject.toml
```diff
@@ -27,6 +27,7 @@ classifiers = [
   "Topic :: Scientific/Engineering",
 ]
 dependencies = [
+  "numpy <=1.24.3",
```
agoose77, line 30: I'd suggest moving this to the CI resolution, if required; we'll patch this upstream in short order, and then make a new release!

src/dask_awkward/lib/core.py
```diff
@@ -2051,3 +2059,13 @@ def typetracer_from_form(form: Form) -> ak.Array:
...
+
+
+def make_unknown_length(array: ak.Array) -> ak.Array:
+    return ak.Array(
+        ak.to_backend(
+            array,
+            "typetracer",
+            highlevel=False,
+        ).to_typetracer(forget_length=True)
+    )
```
agoose77, line 2071: This can just be `to_layout(array).to_typetracer(...)` :)

pyproject.toml
douglasdavis: right, I did this as a temporary sanity check just to see if GH actions would pass :) - gone now!

tests/test_behavior.py
```diff
@@ -31,6 +31,7 @@ def non_dask_method(self, _dask_array_=None):
         return _dask_array_
 
 
+@pytest.mark.xfail(reason="NumPy 1.25 mixin __slots__ change")
```
agoose77, line 34: We could use a conditional fail (check `awkward>XXX || numpy>=YYY`), or just leave it until we fix it!

douglasdavis: thanks for the suggestion!

src/dask_awkward/lib/core.py
douglasdavis: :+1:

PR #278, opened 2023-06-09
fix: divisions after slicing

douglasdavis:
based on iris-hep slack discussions

### DIFF
--- a/src/dask_awkward/lib/core.py
+++ b/src/dask_awkward/lib/core.py
@@ -923,7 +923,10 @@ class Array(DaskMethodsMixin, NDArrayOperatorsMixin):
             outpart += 1
             remainder += (self.divisions[i + 1] - self.divisions[i]) % step
             divisions.append(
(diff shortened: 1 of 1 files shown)


src/dask_awkward/lib/core.py
```diff
@@ -923,7 +923,10 @@ def _getitem_slice_on_zero(self, where: tuple[slice, ...]):
             outpart += 1
             remainder += (self.divisions[i + 1] - self.divisions[i]) % step
             divisions.append(
-                (self.divisions[i + 1] - self.divisions[i]) // step + divisions[-1]
+                min(
+                    (self.divisions[i + 1] - self.divisions[i]) // step + divisions[-1],
+                    divisions[-1] + slice_end,
```
martindurant, line 928: It won't be god enough. I'll get back to you. Also the test isn't right :|
```
assert out.compute().tolist() == daa[where, rest].compute().tolist()
```
->
```
assert out.compute().tolist() == daa.compute()[where, rest].tolist()```

martindurant:
#729 should do it

PR #279, opened 2023-06-09
fix: zeroth slicing divisions

martindurant:


douglasdavis:
We talked about his earlier today but going ahead and posting the comment for my own reference to come back and take a stab at this!

---

Looks like we still have a couple of more complicated slice bad divisions

for example:

```python
>>> concrete = ak.Array([[1, 2, 3], [4], [5, 6, 7], [8, 9]] * 25)
>>> lazy = dak.from_awkward(concrete, npartitions=4)
```
^^ total length of 100 elements with 25 per partition

Taking a relatively complex slice:

```python
>>> lazy[10:68:5].compute().tolist() == concrete[10:68:5].tolist()
True
```
OK good. 

```python
>>> len(lazy[10:68:5])
15
>>> len(lazy[10:68:5].compute())
12
```
not good.

```python
>>> lazy[10:68:5].divisions
(0, 5, 10, 15)
```
and not good

this is another case of divisions reporting the wrong thing from the def __len__(self): return divisions[-1]. It looks like what's happening is the original divisions are being reported but with the correct number of partitions.

every fifth element from 10 to 68 is equivalent to every fifth element from 10 to 65; with this we should expect to touch partitions 0, 1, and 2 from the original array

- partition 0 being (10, 25) (length 3) (items 10, 15, and 20), 
- partition 1 being [25, 50) (length 5) (items 25, 30, 35, 40, 45),
- partition 2 being [50, 65] (length 4) (items 50, 55, 60, 65)

Also things don't look good with

```python
>>> lazy[-30:].divisions
(0, 25, 50)
>>> len(lazy[-30:])
50
```

douglasdavis:
I think we can use [`dask.array._slice_1d`](https://github.com/dask/dask/blob/24adabbd90137a24ffdd18867ffe95e335bc8418/dask/array/slicing.py#L345) and not reinvent the wheel!

martindurant:
I'll add your actual tests from above, because test_getitem_zero_slice_divisions already passed with the existing code, after correcting the test itself.

I had a look at _slice_id and got it to pass some tests, but it wasn't as simply plug-and-play as you might have hoped.

martindurant:
Oh, I see the new test is the same, so I'm not sure what went wrong?! This is going green.

douglasdavis:
Ok now I'm super confused, I think there's a chance I didn't have an updated branch when I was playing with this yesterday-- huge face palm!

douglasdavis:
(I also played around with `_slice_1d` and got something working, but if this is working no sense in refactoring)

douglasdavis:
OK yea I definitely was testing this with an old status of the branch. Always fun to waste a couple hours 😝 . I've just hardened the divisions test that I wrote and I think this is good to go.

PR #294, opened 2023-06-21
feat(map_partitions): allow for dask collections to be passed as kwargs or as part of higher level structures

lgray:
This will allow more expressive signatures as inputs to map_partitions.

Resulting code is pleasantly clean.

### DIFF
--- a/src/dask_awkward/lib/core.py
+++ b/src/dask_awkward/lib/core.py
@@ -1919,7 +1919,7 @@ def to_length_zero_arrays(objects: Sequence[Any]) -> tuple[Any, ...]:
     return tuple(map(length_zero_array_or_identity, objects))
 
 
-def map_meta(fn: Callable, *args: Any, **kwargs: Any) -> ak.Array | None:
+def map_meta(fn: Callable, *args: Any, traverse: bool=True, **kwargs: Any) -> ak.Array | None:
     arg_colls, arg_repack = unpack_collections(args, traverse=traverse)
     kwarg_colls, kwarg_repack = unpack_collections(kwargs, traverse=traverse)
 
--- a/src/dask_awkward/lib/core.py
+++ b/src/dask_awkward/lib/core.py
@@ -10,7 +10,7 @@ import warnings
 from collections.abc import Callable, Hashable, Mapping, Sequence
 from functools import cached_property, partial
 from numbers import Number
-from typing import TYPE_CHECKING, Any, Literal, TypeVar
+from typing import TYPE_CHECKING, Any, Literal, TypeVar, overload
 
 import awkward as ak
 import dask.config
@@ -1888,21 +1888,33 @@ def meta_or_identity(obj: Any) -> Any:
     return obj
 
 
+@overload
 def to_meta(objects: Sequence[Any]) -> tuple[Any, ...]:
-    """In a sequence convert Dask Awkward collections to their metas.
+    ...
+
+
+@overload
+def to_meta(objects: dict[str, Any]) -> dict[str, Any]:
+    ...
+
+
+def to_meta(objects):
+    """Convert sequence or dict of Dask Awkward collections to their metas.
 
     Parameters
     ----------
-    objects : Sequence[Any]
-        Sequence of objects.
+    objects : Sequence[Any] or dict[str, Any]
+        Sequence or dictionary of objects to retrieve metas from.
 
     Returns
     -------
-    tuple[Any, ...]
-        The sequence of objects where collections have been replaced
-        with their metadata.
+    tuple[Any, ...] or dict[str, Any]
+        The sequence of objects (or dictionary) where collections have
+        been replaced with their metadata.
 
     """
+    if isinstance(objects, dict):
+        return {k: meta_or_identity(v) for k, v in objects.items()}
     return tuple(map(meta_or_identity, objects))
 
 
--- a/src/dask_awkward/lib/core.py
+++ b/src/dask_awkward/lib/core.py
@@ -1954,8 +1954,9 @@ def map_meta(
             )
         pass
     try:
-        lzas = to_length_zero_arrays(args)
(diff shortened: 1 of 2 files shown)


lgray:
OK there we go - got all the original tests passing.

lgray:
@douglasdavis @martindurant @agoose77 I think this is complete, would appreciate a review! 

It turns out we actually need this over in coffea for correctly dealing with the triton inference server. Packing arrays according to its interface makes it much more clear to the user what to do to get a meaningful answer.

src/dask_awkward/lib/core.py
```diff
@@ -1530,22 +1564,43 @@ def map_partitions(
     This is effectively the same as `d = c * a`
 
     """
-    token = token or tokenize(fn, *args, meta, **kwargs)
-    label = label or funcname(fn)
+    token = token or tokenize(base_fn, *args, meta, **kwargs)
+    label = label or funcname(base_fn)
```
douglasdavis, line 1568: purely cosmetic nit:
```suggestion
    label = hyphenize(label or funcname(base_fn))
```

src/dask_awkward/lib/core.py
```diff
@@ -1530,22 +1564,43 @@ def map_partitions(
...
         opt_touch_all=opt_touch_all,
-        **kwargs,
     )
-    deps = [a for a in args if is_dask_collection(a)] + [
-        v for _, v in kwargs.items() if is_dask_collection(v)
-    ]
 
     if meta is None:
-        meta = map_meta(fn, *args, **kwargs)
+        meta = map_meta(fn, *arg_deps_expanded, *kwarg_deps)
```
douglasdavis, line 1603: if the function expects or requires named kwargs does `*kwarg_deps` not expand enough information (that is, why don't we need dict `**` unrolling here?). It looks like tests are working so it seems this is indeed correct! I'm just wondering how

src/dask_awkward/lib/core.py
```diff
@@ -1449,22 +1458,45 @@ def partitionwise_layer(
...
+    def __init__(self, the_fn, arg_repackers, kwarg_repacker, arg_lens_for_repackers):
+        self.fn = the_fn
+        self.arg_repackers = arg_repackers
+        self.kwarg_repacker = kwarg_repacker
+        self.arg_lens_for_repackers = arg_lens_for_repackers
+
+    def __call__(self, *args_deps_expanded):
+        args = []
+        len_args = 0
+        for repacker, n_args in zip(self.arg_repackers, self.arg_lens_for_repackers):
```
douglasdavis, line 1471: this is a place where some inline comments summarizing what's going on here will be useful for future maintenance. 

something along the lines of ...
```
The purpose of keeping track of lengths and slicing arguments is 
so we can separate what were originally unnamed args vs named kwargs.
for example. we then use the repackers to recreate the structure
necessary to pass to the desired callable.
```

src/dask_awkward/lib/core.py
```diff
@@ -1530,22 +1564,43 @@ def map_partitions(
...
     name = f"{label}-{token}"
+    kwarg_deps, kwarg_repacker = unpack_collections(kwargs, traverse=traverse)
+    deps, _ = unpack_collections(*args, *kwargs.values(), traverse=traverse)
+
+    arg_deps_expanded = []
+    arg_repackers = []
+    arg_lens_for_repackers = []
+    for arg in args:
+        this_arg_deps, repacker = unpack_collections(arg, traverse=traverse)
+        if len(this_arg_deps) > 0:
```
douglasdavis, line 1578: just for my own clarification: does this case just mean that the `arg` is not a dask collection, correct?

douglasdavis (commented):
Thanks @lgray! Overall this looks pretty good to me. I left a few inline comments/questions. I think it would also be a good idea to have a couple more tests mixing args and kwargs with map_partitions

src/dask_awkward/lib/core.py
lgray: `kwarg_deps` is guaranteed to be a un-nested list of unlabeled arguments (the first return value of `unpack_collections`). It's the repack function that carries around the actual args/kwargs structure, if there is any.

src/dask_awkward/lib/core.py
lgray: Almost, it means arg does not contain any dask collections (where the collection could be arg itself) so we should not repack it. This is what lets the blockwise Index arguments to continue to work as normal, since instead of packing it up we ship it.

src/dask_awkward/lib/core.py
lgray: Yeah it's a bit odd that kwargs_deps is actually a flat list of arguments that need to be repacked into kwargs.
I'll figure out a better name.

lgray:
A pure kwargs example I think is sufficient. We test pure args elsewhere in the code, and one args + kwargs example is  enough to show it functions correctly.

I guess I can add one with a scalar arg that's broadcasted or something.

lgray:
@douglasdavis all comments addressed, tests extended :-)

douglasdavis:
Yea the `structured_function` test is a good one!

I wrote this one for some local tinkering- may be worth adding

def my_func(aaa, bbb, *, ccc=None, ddd=None):
    return (aaa + bbb) ** ccc - ddd

```python
a = ak.Array([[1,2,3,],[4,5],[6,7,8]])
b = ak.Array([[-10, -10, -10], [-10, -10], [-10, -10, -10]])
c = ak.Array([0, 1, 2])
d = 1

aa = dak.from_awkward(a, npartitions=2)
bb = dak.from_awkward(b, npartitions=2)
cc = dak.from_awkward(c, npartitions=2)
dd = d

res = my_func(a, b, ccc=c, ddd=d)

print(res)

resl = dak.map_partitions(my_func, aa, bb, ccc=cc, ddd=dd)
```

lgray:
please go ahead and add it!

douglasdavis:
looks good!

PR #295, opened 2023-06-21
feat: add `Array.__awkward_function__` method

douglasdavis:
Adds support for using `ak.XXX` functions with `dask_awkward.Array` objects.

### DIFF
--- a/src/dask_awkward/lib/core.py
+++ b/src/dask_awkward/lib/core.py
@@ -1197,16 +1197,18 @@ class Array(DaskMethodsMixin, NDArrayOperatorsMixin):
         """Clear the divisions of a Dask Awkward Collection."""
         self._divisions = (None,) * (self.npartitions + 1)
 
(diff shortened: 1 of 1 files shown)


src/dask_awkward/lib/core.py
```diff
@@ -1197,6 +1197,17 @@ def clear_divisions(self) -> None:
         """Clear the divisions of a Dask Awkward Collection."""
         self._divisions = (None,) * (self.npartitions + 1)
 
+    def __awkward_function__(self, func, args, kwargs):
+        import dask_awkward
+
+        if any(isinstance(arg, ak.Array) for arg in args):
+            raise TypeError("cannot mix awkward.Array and dask_awkward.Array")
```
douglasdavis, line 1204: This is currently too simple. `args` can be many different things (i.e. a dict containing `ak.Array` or `dak.Array` objects); so this check will need to be more robust

PR #298, opened 2023-06-23
feat: add `drop_none` method

iasonkrom:
From slack discussions, `drop_none` was never put into `dask-awkward` as it was added recently in `awkward`. Starting a PR to add this method.


douglasdavis (commented):
let's add the axis=None check and this should be good to go. Looks like drop_none never changes dimensions of an array :+1:

PR #304, opened 2023-06-29
feat: do partition compat. checks with `PartitionCompatibility` sum type

douglasdavis:
TL;DR

- existing `dak.compatible_partitions` gets a new argument (`how_strict`) which provides some user-definable passing criteria. (existing uses of the function are unchanged, just a new argument where the default value lines up with current behavior).
- new function `dak.partition_compatibility` along with new sum type `PartitionCompatibility` can be used for more flexibility.

---

This PR introduces a `partition_compatibility` function that returns a sum type (the new `PartitionCompatibility` enum) that can be `YES`, `NO`, or `MAYBE`. This will allow downstream code to use `MAYBE` as desired. In dask-awkward we will simply use `NO` to raise exceptions.

Any place that was original checking

```python
if not dak.compatible_partitions(a, b):
    raise IncompatiblePartitions(...)
```
can now write:
```python
if dak.partition_compatibility(a, b) == dak.PartitionCompatibility.NO:
    raise IncompatiblePartitions(...)
```
One can also use `MAYBE` to go in another direction or use [pattern matching](https://peps.python.org/pep-0636/) with 3.10

the existing `compatible_partitions` function now gets a new named argument: `how_strict` which defaults to `PartitionCompatibility.MAYBE` (or `1`) (the existing behavior before this PR), setting it to `PartitionCompatibility.YES` (or `2`) will cause the function to return `False` if we are not _absolutely_ sure of the compatibility with via known `divisions`

### DIFF
--- a/docs/api-inspect.rst
+++ b/docs/api-inspect.rst
@@ -6,5 +6,7 @@ Inspection
 .. autosummary::
    :toctree: generated/
 
+   partition_compatibility
+   PartitionCompatibility
    necessary_columns
    sample
--- a/src/dask_awkward/lib/__init__.py
+++ b/src/dask_awkward/lib/__init__.py
@@ -1,6 +1,11 @@
-from dask_awkward.lib.core import Array, Record, Scalar
+from dask_awkward.lib.core import Array, PartitionCompatibility, Record, Scalar
 from dask_awkward.lib.core import _type as type
-from dask_awkward.lib.core import map_partitions, typetracer_from_form
+from dask_awkward.lib.core import (
+    compatible_partitions,
+    map_partitions,
+    partition_compatibility,
+    typetracer_from_form,
+)
 from dask_awkward.lib.describe import fields
 from dask_awkward.lib.inspect import necessary_columns, sample
 from dask_awkward.lib.io.io import (
--- a/src/dask_awkward/lib/core.py
+++ b/src/dask_awkward/lib/core.py
@@ -2117,64 +2117,6 @@ def typetracer_array(a: ak.Array | Array) -> ak.Array:
         raise TypeError(msg)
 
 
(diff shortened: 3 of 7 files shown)


douglasdavis:
@lgray @iasonkrom would definitely welcome your thoughts. I know coffea is using what this PR deprecates

iasonkrom:
I like the idea. I can play around with it a bit over the weekend I think. I believe only my recent PR used what this PR deprecates so it should be fine if this doesn’t change anything in dask_awkward functionality, right @lgray?

lgray:
I like the pattern matching direction a bit more, but another idea occurs to me, a kwarg that's "strictness".

Somehow the whole idea of trinary statements is bugging me, but to keep it simpler we can have binary statements with a knob for strictness.

```python3
if dak.compatible_partitions(a, b, how_strict=some_integer_value)
    raise
```

Then we can define what the levels of strictness are much like the compression levels in compression algorithms.

Somehow that jives with how I think about this a bit better. What do you think?

lgray:
I think what was bugging me here is that there are different kinds of `compatible_partition` failures that loosely fit into an idea of strictness. Really they're categories and there's a clear set of core compatibility requirements, followed by various categories of failure.

I think a pattern matching implementation may make the most sense?

iasonkrom:
I actually also think that it may be better if `PartitionCompatibility` is implement as "strictness". I find this simpler to have a kwarg instead of doing `dak.partition_compatibility(a, b) == dak.PartitionCompatibility.NO`. However, my opinion shouldn't matter and more experienced people should decide

douglasdavis:
I like the adding a `how_strict` argument for the existing `compatible_parititions` functions. This PR now doesn't change any existing behavior, but adds more flexiblity

PR #311, opened 2023-07-06
feat: support for `dask.array` collections in `map_partitions`

douglasdavis:
- add support for dask.array.Array in map_partitions
- add test


douglasdavis:
This PR also removes "custom" meta determination in `__array_func__`; it now relies on the logic in `map_partitions` to determine the new collection's meta.

PR #317, opened 2023-07-12
Revert `unproject_layout`'s action; use `shape_touched` in column optimization.

douglasdavis:
Should fix #314 and #313.

This reverts the action that unproject_layout and plugs the hole left behind by using the `shape_touched` attribute of the typetracer report. The only negative side-effect we have is converting to `dask.array.Array` is failing in tests because of a mismatch between a regular numpy array and a masked numpy array.

Issue #314, opened 2023-07-07
dak.ones_like / zeros_like fails


lgray:
Caught another one trying to figure out why `dak.count(events.event, axis=0)` is so slow.
Was attempting to do `dak.sum(dak.ones_like(events.event), axis=0)` but ran what's reproduced below.
Same error for zeros_like, since it's coming from full_like.

```python3
import uproot
import awkward as ak
import dask_awkward as dak

tree = uproot.open({"https://github.com/CoffeaTeam/coffea/raw/master/tests/samples/nano_dy.root": "Events"})
dtree = uproot.dask(tree)
    
print(ak.ones_like(tree["event"].array()))

print(dak.ones_like(dtree.event).compute())
```

fails with:
```
Traceback (most recent call last):
  File "/Users/lgray/miniforge3/envs/coffea-dev/lib/python3.8/site-packages/awkward/_dispatch.py", line 43, in dispatch
    next(gen_or_result)
  File "/Users/lgray/miniforge3/envs/coffea-dev/lib/python3.8/site-packages/awkward/operations/ak_ones_like.py", line 38, in ones_like
    return _impl(array, highlevel, behavior, dtype, including_unknown)
  File "/Users/lgray/miniforge3/envs/coffea-dev/lib/python3.8/site-packages/awkward/operations/ak_ones_like.py", line 42, in _impl
    return ak.operations.ak_full_like._impl(
  File "/Users/lgray/miniforge3/envs/coffea-dev/lib/python3.8/site-packages/awkward/operations/ak_full_like.py", line 205, in _impl
    out = ak._do.recursively_apply(layout, action, behavior)
  File "/Users/lgray/miniforge3/envs/coffea-dev/lib/python3.8/site-packages/awkward/_do.py", line 35, in recursively_apply
    return layout._recursively_apply(
  File "/Users/lgray/miniforge3/envs/coffea-dev/lib/python3.8/site-packages/awkward/contents/numpyarray.py", line 1258, in _recursively_apply
    result = action(
  File "/Users/lgray/miniforge3/envs/coffea-dev/lib/python3.8/site-packages/awkward/operations/ak_full_like.py", line 119, in action
    nplike.ones_like(original, dtype=dtype),
  File "/Users/lgray/miniforge3/envs/coffea-dev/lib/python3.8/site-packages/awkward/_nplikes/array_module.py", line 87, in ones_like
    assert not isinstance(x, PlaceholderArray)
AssertionError

[...]

Issue #313, opened 2023-07-07
Computes that require shape only fail after form rehydration (count, zeros_like, ones_like, ...)


lgray:
This is in uproot 5.0.10, dask-awkward 2023.7.0, and awkward 2.3.1.
I'm not sure where to put this one, since the actual error is in awkward, but I'm putting it here since it's in a dak function.

@agoose77 

```python3
import uproot
import awkward as ak
import dask_awkward as dak

tree = uproot.open({"https://github.com/CoffeaTeam/coffea/raw/master/tests/samples/nano_dy.root": "Events"})
dtree = uproot.dask(tree)
    
print(ak.count(tree["event"].array()))

print(dak.count(dtree.event).compute())
```

fails with:
```
Traceback (most recent call last):
  File "count_repro.py", line 10, in <module>
    print(dak.count(dtree.event).compute())
  File "/Users/lgray/miniforge3/envs/coffea-dev/lib/python3.8/site-packages/dask/base.py", line 314, in compute
    (result,) = compute(self, traverse=False, **kwargs)
  File "/Users/lgray/miniforge3/envs/coffea-dev/lib/python3.8/site-packages/dask/base.py", line 599, in compute
    results = schedule(dsk, keys, **kwargs)
  File "/Users/lgray/miniforge3/envs/coffea-dev/lib/python3.8/site-packages/dask/threaded.py", line 89, in get
    results = get_async(
  File "/Users/lgray/miniforge3/envs/coffea-dev/lib/python3.8/site-packages/dask/local.py", line 511, in get_async
    raise_exception(exc, tb)
  File "/Users/lgray/miniforge3/envs/coffea-dev/lib/python3.8/site-packages/dask/local.py", line 319, in reraise
    raise exc
  File "/Users/lgray/miniforge3/envs/coffea-dev/lib/python3.8/site-packages/dask/local.py", line 224, in execute_task
    result = _execute_task(task, data)
  File "/Users/lgray/miniforge3/envs/coffea-dev/lib/python3.8/site-packages/dask/core.py", line 119, in _execute_task
    return func(*(_execute_task(a, cache) for a in args))
  File "/Users/lgray/miniforge3/envs/coffea-dev/lib/python3.8/site-packages/dask/optimization.py", line 990, in __call__
    return core.get(self.dsk, self.outkey, dict(zip(self.inkeys, args)))
[...]

PR #319, opened 2023-07-13
fix: make `unproject_layout` do nothing (just return original layout)

douglasdavis:
This reverts our form hydration feature originally added in https://github.com/dask-contrib/dask-awkward/pull/184. It appears to have triggered #313 and #314. We need more testing of the feature before it's ready.

This PR is an alternative to #317. In that PR we not only revert unproject_layout's action, but we also introduce more touching in the column optimization to make up for what unproject_layout was trying to accomplish. Unfortunately using `shape_touched` from the typetracer report causing over-touching of data.

cc @lgray


lgray:
ran tests on coffea and it looks like we're good

PR #327, opened 2023-07-21
draft: unproject_layout work

douglasdavis:


douglasdavis:
ref #313

lgray:
any blockers on getting this merged + release?

agoose77:
In theory, we're back where we were supposed to be in July: this PR will fix the column optimisation, but we expect it to have some problems where nested contents have lengths that need to be determined. I think that's probably a better error to encounter, and we anticipate it. 

I'm working towards that fix; right now, I'm addressing `axis=0` concatenation and something related for tiled. So, I vote to merge this and see what breaks.

PR #330, opened 2023-07-25
Fix error when using from_awkward on an empty array

jrueb:
This change fixes the ValueError that is raised if one tries to use from_awkward on an empty awkward array. For example by executing  `dak.from_awkward(ak.Array([]), 1)`.


douglasdavis:
thanks @jrueb! this looks good- I added another small case to the test (a case where the empty array actually has a form). I'll merge when CI is green :+1:

PR #341, opened 2023-08-16
feat: configuration can handle list of optimizations to run

douglasdavis:
@martindurant - would like your take here. I've been tinkering with tests for the JSON column optimization and I'd like to able to do some graph surgery in the tests to get at results from the input layer (that is, get directly at the result of the call of `ak.from_parquet` or `ak.from_json`), but if the layer chaining occurs it's a bit more difficult to do that.

With this PR no functionality changes, just more config flexibility


martindurant:
Sure, that's a fine way to do it, and agree than in principle it's a good idea to be able to turn specific parts of the optimization to run.

PR #349, opened 2023-08-25
refactor: use `ak.typetracer.touch_data`

agoose77:
This is not yet released in `awkward`.


douglasdavis:
awkward-main job passing :+1:

PR #351, opened 2023-08-25
refactor: use `ak.typetracer.XX` high-level functions

agoose77:
Depends upon https://github.com/scikit-hep/awkward/pull/2671

### DIFF
--- a/src/dask_awkward/__init__.py
+++ b/src/dask_awkward/__init__.py
@@ -12,7 +12,7 @@ from dask_awkward.lib.core import _type as type
 from dask_awkward.lib.core import (
     compatible_partitions,
     map_partitions,
-    partition_compatibility
+    partition_compatibility,
 )
 from dask_awkward.lib.describe import fields
 from dask_awkward.lib.inspect import necessary_columns, sample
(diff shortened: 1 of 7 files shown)


agoose77:
@douglasdavis I've removed part of an existing test that invokes `dak.from_awkward` upon an Awkward Array backed by a typetracer. I *think* this should be an error; typetracers should be an internal detail of dask-awkward. Do you agree with this? If so, I will keep the change and update the PR to add a test for the array backend in `from_awkward`.

douglasdavis (commented):
@agoose77 that test was meant to address an issue with calling `from_awkward` on an empty array (that test was written with this PR: https://github.com/dask-contrib/dask-awkward/pull/330); `ak.Array(some_form.length_zero_array(highlevel=False))` was being used just to create an empty awkward array (so not using a typetracer, but an empty high level array). was the test failing for after the changes in this PR?

agoose77:
@douglasdavis gah, I'm doing too much refactoring work in parallel — I changed the function of this test in a previous commit. Good spot, let me recover the original intention.

PR #354, opened 2023-08-31
feat: add custom pickle implementation

### LINKED ISSUE #344: Layout changes due to pickling of partitions are not tracked during typetracing
Right now, the act of pickling an Awkward Array leads to `ak.to_packed`, which projects out indexed nodes (amongst other things). This means that due to transfer of results between tasks, the form of an array could be different according to whether its predecessor tasks were computed on the same node. 

Off the top of my head, I envisage this being a problem for a layout containing e.g. `IndexedArray` nodes that contain only placeholder `index` and `content`. These would be projected (recursively) during `to_packed`, and ultimately throw an error.

The safest solution is a layout-preserving project, that retains all the nodes but introduces trivial indexes for e.g. `IndexedArray`. I particularly don't want to have to start worrying about `PlaceholderArray` in any other places besides the `nplikes` right now, meaning this function should manually perform things like `IndexedArray.project()` for indexed nodes containing placeholders.

However, it might be safe to use a simplifying `to_packed` that is placeholder-aware, iff. we're willing to drop the requirement that the resulting form matches the typetracing form. I can't yet predict if that will be a problem.

Pinging @jpivarski in case he's already thought about this, but I'm going to think on it.

agoose77:
Fixes #344 by overriding the default Awkward `__reduce__` implementations.

Requires https://github.com/scikit-hep/awkward/pull/2682

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -27,7 +27,7 @@ classifiers = [
   "Topic :: Scientific/Engineering",
 ]
 dependencies = [
-  "awkward >=2.2.4",
+  "awkward >=2.4.0",
   "dask >=2023.04.0",
 ]
 dynamic = ["version"]
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -66,6 +66,9 @@ test = [
 [project.entry-points."dask.sizeof"]
 awkward = "dask_awkward.sizeof:register"
 
+[project.entry-points."awkward.pickle.reduce"]
+dask_awkward = "dask_awkward.pickle:plugin"
+
 [tool.hatch.version]
 source = "vcs"
 path = "src/dask_awkward/__init__.py"
--- a/src/dask_awkward/pickle.py
+++ b/src/dask_awkward/pickle.py
@@ -1,0 +1,66 @@
+from __future__ import annotations
+
+__all__ = ("plugin",)
+
+import pickle
+
+import awkward as ak
+
+
+def pickle_record(record: ak.Record, protocol: int) -> tuple:
+    layout = ak.to_layout(record, allow_record=True)
+    form, length, container = ak.operations.to_buffers(
+        layout.array,
+        buffer_key="{form_key}-{attribute}",
+        form_key="node{id}",
+        byteorder="<",
+    )
+
+    # For pickle >= 5, we can avoid copying the buffers
+    if protocol >= 5:
+        container = {k: pickle.PickleBuffer(v) for k, v in container.items()}
+
+    if record.behavior is ak.behavior:
+        behavior = None
+    else:
+        behavior = record.behavior
+
+    return (
+        object.__new__,
+        (ak.Record,),
+        (form.to_dict(), length, container, behavior, layout.at),
+    )
+
+
+def pickle_array(array: ak.Array, protocol: int) -> tuple:
+    layout = ak.to_layout(array, allow_record=True)
+    form, length, container = ak.operations.to_buffers(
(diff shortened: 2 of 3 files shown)


src/dask_awkward/pickle.py
```diff
@@ -0,0 +1,66 @@
...
+    if array.behavior is ak.behavior:
+        behavior = None
+    else:
+        behavior = array.behavior
+
+    return (
+        object.__new__,
+        (ak.Array,),
+        (form.to_dict(), length, container, behavior),
+    )
```
agoose77, line 57: For now this is just `to_buffers` with _no_ packing. In future we can implement a form-preserving pack operation, perhaps in Awkward Array

jpivarski: No packing is not a bad way to start for serialization in Dask (as long as it's not the default for pickle outside of Dask).

src/dask_awkward/pickle.py
```diff
@@ -0,0 +1,66 @@
...
+    form, length, container = ak.operations.to_buffers(
+        layout.array,
+        buffer_key="{form_key}-{attribute}",
+        form_key="node{id}",
+        byteorder="<",
+    )
+
+    # For pickle >= 5, we can avoid copying the buffers
+    if protocol >= 5:
+        container = {k: pickle.PickleBuffer(v) for k, v in container.items()}
```
martindurant, line 21: We certainly need a test to ensure this does the right thing

src/dask_awkward/pickle.py
```diff
@@ -0,0 +1,66 @@
...
+    )
+
+
+def plugin(obj, protocol: int) -> tuple | NotImplemented:
+    if isinstance(obj, ak.Record):
+        return pickle_record(obj, protocol)
+    elif isinstance(obj, ak.Array):
+        return pickle_array(obj, protocol)
+    else:
+        return NotImplemented
```
martindurant, line 66: Return rather than raise?

pyproject.toml
```diff
@@ -66,6 +66,9 @@ test = [
 [project.entry-points."dask.sizeof"]
 awkward = "dask_awkward.sizeof:register"
 
+[project.entry-points."awkward.pickle.reduce"]
+dask_awkward = "dask_awkward.pickle:plugin"
```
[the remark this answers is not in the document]
agoose77: Yep!

src/dask_awkward/pickle.py
agoose77: Return is intentional: the sentinel object is there to permit awkward to use this mechanism for more types in a backwards compatible way. 🙂

agoose77:
@martindurant are you OK for me to merge this? :)

tests/test_pickle.py
```diff
@@ -0,0 +1,19 @@
...
+import pickle
+
+import awkward as ak
+
+
+def test_pickle_ak_array():
+    buffers = []
+    array = ak.Array([[1, 2, 3], [4, 5, 6], [7, 8, 9], [10, 11, 12]])[[0, 2]]
+    next_array = pickle.loads(
+        pickle.dumps(array, protocol=5, buffer_callback=buffers.append), buffers=buffers
```
martindurant, line 12: This is a nice concise way to test pickle 5!
Can we add a line that confirms that `buffer` did in fact get some content?

That's all I have, otherwise good to go.

agoose77:
Let me also test records.

PR #357, opened 2023-09-05
feat: add `bytes_with_sample` for creating bytes reading instructions

douglasdavis:
Pulling this out of the JSON PR (#94) because we'll also use it for `from_text` (prototype here: https://github.com/douglasdavis/dask-awkward/pull/7)

### DIFF
--- a/src/dask_awkward/lib/io/io.py
+++ b/src/dask_awkward/lib/io/io.py
@@ -11,6 +11,7 @@ from awkward.types.numpytype import primitive_to_dtype
 from dask.base import flatten, tokenize
 from dask.highlevelgraph import HighLevelGraph
 from dask.utils import funcname, is_integer, parse_bytes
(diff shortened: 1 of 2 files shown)


src/dask_awkward/lib/io/io.py
```diff
@@ -572,3 +575,121 @@ def from_map(
...
+                length = []
+
+                # figure out offsets, spreading around spare bytes
+                while size - place > (blocksize1 * 2) - 1:
+                    place += blocksize1
+                    off.append(int(place))
+                    length.append(off[-1] - off[-2])
+                length.append(size - off[-1])
+
+                if not_zero:
```
douglasdavis, line 649: @martindurant this `not_zero` argument is the one thing from this copied upstream-dask-code that I don't totally understand; do you know when this is relevant?

PR #358, opened 2023-09-05
feat: add a general `from_text` input function

douglasdavis:


src/dask_awkward/lib/io/io.py
```diff
@@ -727,3 +728,70 @@ def _bytes_with_sample(
...
+        delimiter,
+        False,
+        blocksize,
+        sample_size,
+    )
+
+    rfind = sample_bytes.rfind(delimiter)
+    if rfind > 0:
+        sample_bytes = sample_bytes[:rfind]
+    meta = typetracer_array(_string_array_from_bytestring(sample_bytes, delimiter))
```
martindurant, line 789: Isn't meta always array[string] ? Even if all input files are empty, this layout is a sufficient description. I think we should just never get a sample (sample_size=None) and don't bother finding a delimiter here.

douglasdavis: Ah yes of course, thanks!

Issue #363, opened 2023-09-13
Filling histograms twice fails in 2023.9.1 only


iasonkrom:
```python
import uproot
import hist.dask as hda
import dask_awkward as dak
from distributed import Client

events = uproot.dask({"../../coffea_dev/coffea/tests/samples/nano_dy.root": "Events"})

h = hda.Hist.new.Variable([0, 10, 20, 30, 40, 80, 120, 200], name="x").Double()

p1 = events.Electron_pt[:20]
p2 = events.Electron_pt[20:]

h.fill(dak.flatten(p1))
h.fill(dak.flatten(p2))

h.compute(scheduler="processes")
```
This fails with `Client()` or `"processes"` scheduler .
With `"processes"` it gives: 
```
File ~/fun/egamma_dev/venv/lib/python3.10/site-packages/dask_awkward/pickle.py:46, in <dictcomp>()
     44 # For pickle >= 5, we can avoid copying the buffers
     45 if protocol >= 5:
---> 46     container = {k: pickle.PickleBuffer(v) for k, v in container.items()}
     48 if array.behavior is ak.behavior:
     49     behavior = None

TypeError: a bytes-like object is required, not 'PlaceholderArray'
```
while with `Client()` it gives: 
```
File ~/fun/egamma_dev/venv/lib/python3.10/site-packages/awkward/_nplikes/shape.py:77, in __gt__()
     76 def __gt__(self, other):
---> 77     raise TypeError("cannot order unknown lengths")

TypeError: cannot order unknown lengths
```

douglasdavis:
@agoose77 since you worked on the pickle stuff I'm going ahead and pinging you here :) I'll still take a look at investigating at some point today

iasonkrom:
@douglasdavis @agoose77 Also I would like to note that I've been encountering some strange behavior in 2023.9.1 in general.
For instance this:
```python
events = NanoEventsFactory.from_root(
    {"root_files/Egamma1.root": "Events"},
    permit_dask=True,
    chunks_per_file=10,
).events()

def filter_events(events, pt):
    good_events = events[dak.num(events.Electron) >= 2]
    abs_eta = abs(events.Electron.eta)
    pass_eta_ebeegap = (abs_eta < 1.4442) | (abs_eta > 1.566)
    pass_tight_id = events.Electron.cutBased == 4
    pass_pt = events.Electron.pt > pt
    pass_eta = abs_eta <= 2.5
[...]

PR #366, opened 2023-09-17
fix: pickle placeholder

agoose77:


src/dask_awkward/lib/optimize.py
```diff
@@ -419,23 +420,31 @@ def _get_column_reports(dsk: HighLevelGraph) -> dict[str, Any]:
...
+            # List offsets are tagged as {key}.{LIST_KEY}. This routine resolve
+            # _columns_, so we use a wildcard to indicate that we want to load
+            # *any* child column of this list. If the list contains no records,
+            # then we load
+            if column.endswith(LIST_KEY):
+                list_parent_path = column[: -(len(LIST_KEY) + 1)].rstrip(".")
+                if list_parent_path not in necessary_columns:
+                    necessary_columns.append(f"{list_parent_path}.*")
+            else:
+                necessary_columns.append(column)
```
agoose77, line 445: @douglasdavis this is where I was initially thinking that we need to do something for `shape_touched`. However, it might prove fiddly.

Naively, I thought about `A.*` if we want the shape of `A.B`. However, this wouldn't work for the following type of `A`
```
1 * {
    x: var * int64,
    y: var * int64
}
```

Computing `A.y` wouldn't give any information about the length of the `NumpyArray` in `A.x`.

agoose77: Just to follow up (I don't expect you to have seen this over the weekend) — this is the blocker for @lgray's demo this week AFAICT.

douglasdavis (approved):
Things look good to me and tests are passing, just had that one question on the inline comment

src/dask_awkward/lib/optimize.py
agoose77: This PR fixes the pickling problem, but there's still the task of loading lengths for buffers with `shape_touched`. I'll move this to Slack :)

iasonkrom:
It doesn't do it for me.

The reproducer in https://github.com/dask-contrib/dask-awkward/issues/363:
```python
import dask
import uproot
import dask_awkward as dak
events = uproot.dask({"/path/to/coffea/tests/samples/nano_dy.root": "Events"})
p1 = events.Electron_pt[:20]
p2 = events.Electron_pt[20:]
dask.compute(p1, p2, scheduler="processes")
```
fails with `TypeError: internal_error: the _UnknownLength class should never be directly instantiated`, and my pytest looks like this:
```python
FAILED tests/test_tnp.py::test_local_compute[False-processes] - TypeError: internal_error: the _UnknownLength class should never be directly instantiated
FAILED tests/test_tnp.py::test_local_compute[True-processes] - TypeError: internal_error: the _UnknownLength class should never be directly instantiated
FAILED tests/test_tnp.py::test_distributed_compute[False] - TypeError: cannot order unknown lengths
FAILED tests/test_tnp.py::test_distributed_compute[True] - TypeError: cannot order unknown lengths
```
```python
import dask
import uproot
import dask_awkward as dak
from distributed import Client
events = uproot.dask({"/path/to/coffea/tests/samples/nano_dy.root": "Events"})
p1 = events.Electron_pt[:20]
p2 = events.Electron_pt[20:]
with Client():
    dask.compute(p1, p2)
```
This snippet wil give you this: `TypeError: cannot order unknown lengths` (same as the pytest)

agoose77:
@iasonkrom you need the latest Awkward for this (HEAD)

iasonkrom:
Yeah my bad, however, 

first snippet now gives:
```python
TypeError: PlaceholderArray supports only trivial slices, not int

This error occurred while calling

    ak.from_buffers(
        {'class': 'RecordArray', 'fields': ['run', 'luminosityBlock', 'event'...
        40
        {'node1-data': <awkward._nplikes.placeholder.PlaceholderArray object ...
        highlevel = False
        buffer_key = '{form_key}-{attribute}'
        byteorder = '<'
    )
```
and the second snippet still the same error:
```python
  File "/Users/iason/fun/egamma_dev/egamma-tnp/test2.py", line 10, in <module>
    dask.compute(p1, p2)
  File "/Users/iason/fun/egamma_dev/venv/lib/python3.10/site-packages/dask/base.py", line 628, in compute
  File "/Users/iason/fun/egamma_dev/venv/lib/python3.10/site-packages/distributed/spill.py", line 181, in __setitem__
    super().__setitem__(key, value)
  File "/Users/iason/fun/egamma_dev/venv/lib/python3.10/site-packages/zict/buffer.py", line 193, in __setitem__
    self.fast[key] = value
  File "/Users/iason/fun/egamma_dev/venv/lib/python3.10/site-packages/zict/lru.py", line 136, in __setitem__
    self.set_noevict(key, value)
  File "/Users/iason/fun/egamma_dev/venv/lib/python3.10/site-packages/zict/common.py", line 127, in wrapper
    return func(*args, **kwargs)
  File "/Users/iason/fun/egamma_dev/venv/lib/python3.10/site-packages/zict/lru.py", line 163, in set_noevict
    if weight > self.n:
  File "/Users/iason/fun/egamma_dev/venv/lib/python3.10/site-packages/awkward/_nplikes/shape.py", line 67, in __gt__
    raise TypeError("cannot order unknown lengths")
TypeError: cannot order unknown lengths
```

agoose77:
@iasonkrom could you use https://github.com/scikit-hep/awkward/pull/2714 instead of HEAD for awkward? I might have misremembered with PR you should use :)

douglasdavis:
I'll put out a release with this fix in 👍

PR #367, opened 2023-09-18
misc: `str` module typing

douglasdavis:
Add missing annotations in `str` module

### DIFF
--- a/src/dask_awkward/lib/str.py
+++ b/src/dask_awkward/lib/str.py
@@ -1,15 +1,21 @@
 from __future__ import annotations
 
 import functools
+from collections.abc import Callable
+from typing import Any, TypeVar
 
 import awkward.operations.str as akstr
+from typing_extensions import ParamSpec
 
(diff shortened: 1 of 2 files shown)


src/dask_awkward/lib/str.py
```diff
@@ -597,10 +603,10 @@ def replace_substring(
 @always_highlevel
 def replace_substring_regex(
     array: Array,
-    pattern,
-    replacement,
+    pattern: str,
+    replacement: str | bytes,
     *,
-    max_replacements=None,
+    max_replacements: int | None = None,
```
martindurant, line 609: some formatters prefer Optional[int]. I don't know if there's any consequence of choosing one over the other.

douglasdavis: with `from __future__ import annotations` the current ["blessed" syntax](https://stackoverflow.com/questions/69440494/python-3-10-optionaltype-or-type-none) is `foo | None` over `Optional[foo]`. Some interesting arguments in that SO thread, of course!

PR #368, opened 2023-09-18
draft: feat: use shape touched information in necessary columns optimization

douglasdavis:


douglasdavis:
@agoose77 so far this is just the start of some code to be able to pass `necessary_shape_columns` around the dask-awkward code base inline with how we pass around `necessary_columns`. I think the major part of getting this to work will be in the `__call__` methods; Once we've created an array with the necessary _data_ columns, we can then (perhaps in `unproject_layout`, because we call `unproject_layout` in the `__call__` methods, or in something preceeding/following `unproject_layout`) create `PlaceholderArray`s with known length for the columns that are necessary _shape_ columns; we'll know which one's they are by tracking it as an attribute of the callable class.

Does this sound reasonable?

src/dask_awkward/lib/optimize.py
```diff
@@ -123,7 +123,10 @@ def optimize_columns(dsk: HighLevelGraph) -> HighLevelGraph:
     for name, neccols in layer_to_necessary_columns.items():
         meta = layers[name]._meta
         neccols = _prune_wildcards(neccols, meta)
-        layers[name] = layers[name].project_columns(neccols)
+        layers[name] = layers[name].project_columns(
+            neccols,
+            necessary_shape_columns=None,
```
douglasdavis, line 128: this `None` can be replaced by `TypeTracerArray.shape_touched`

douglasdavis:
cc @lgray; if my comment above is the route we take then there will be some required changes in uproot (I'd be happy to help/do it- it should be simple- the function classes like `upoot.dask.UprootRead` will need their `project_columns` implementations updated)

lgray:
that shouldn't be a big issue!

PR #371, opened 2023-09-21
feat(draft): `dask_awkward.to_rdataframe`

### LINKED ISSUE #370: Feature request: support for ak.to_rdataframe
Hello,

for the purposes of using C++ functions with Awkward arrays as arguments (called from the Python environment) it seems the recommendation is to use RDataFrame as the "hook" (and with RDataFrame then it is then trivial to call C++ methods). This is demonstrated (e.g.) [here](https://github.com/ianna/PyHep2022/tree/main) and described in [this talk](https://indico.fnal.gov/event/23628/contributions/240305/attachments/154908/201609/Awkward%20Arrays%20to%20RDataFrame%20and%20back-10.05.2022.pdf).

For Dask-awkward, the relevant method doesn't seem to exist:

`
AttributeError: module 'dask_awkward' has no attribute 'to_rdataframe'
`

Would it be possible for this feature to be added in the long term? In the meantime, is there a method of calling C++ functions from Dask-awkward?

Thanks!

James Catmore (ATLAS)

douglasdavis:
Fix #370

Uses map partitions and dask.delayed to support converting dictionaries of dask-awkward `Array` collections to RDataFrame.

Since awkward's `to_rdataframe` does some loading of ROOT/cppyy related things this requires importing `awkward._connect.rdataframe.to_rdataframe` before it will work. For distributed workloads I'm guessing one would need to have a worker plugin that does that same import.

### DIFF
--- a/src/dask_awkward/lib/io/io.py
+++ b/src/dask_awkward/lib/io/io.py
@@ -1,9 +1,9 @@
 from __future__ import annotations
 
 import math
-from collections.abc import Iterable
+from collections.abc import Callable, Iterable, Mapping, Sequence
 from dataclasses import dataclass
-from typing import TYPE_CHECKING, Any, Callable, Mapping, Protocol
+from typing import TYPE_CHECKING, Any, Protocol
 
 import awkward as ak
 import numpy as np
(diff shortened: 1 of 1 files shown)


martindurant:
> For distributed workloads I'm guessing one would need to have a worker plugin that does that same import.

You could have a custom pickle implementation that imports before rehydrating the object

jpivarski:
Before getting too deep into this, see https://github.com/dask-contrib/dask-awkward/issues/370#issuecomment-1730016956. I think we should be trying to make this connection at the level of cppyy, rather than RDataFrame. There's a 1:1 relationship between how cppyy can be used and how Numba can be used, but RDataFrame does a lot more.

E.g. there's a global setting (default?)

```c++
ROOT::EnableImplicitMT();
```

that distributes processing of every item in the array to all available threads, which (at best) will oversubscribe CPU cores when there are Dask workers on all cores.

Issue #260, opened 2023-05-11
Several functions (including dak.max) incorrectly raise awkward._errors.FieldNotFoundError


jrueb:
Please consider the following
```python
from coffea.nanoevents import NanoEventsFactory, NanoAODSchema
import dask_awkward as dak

events = NanoEventsFactory.from_root(
    "https://github.com/CoffeaTeam/coffea/raw/master/tests/samples/nano_dy.root",
    metadata={"dataset": "DoubleMu"},
    schemaclass=NanoAODSchema,
    permit_dask=True,
).events()
pt = events["Muon"]["pt"]
print(dak.max(pt).compute())
```
This will raise an error `awkward._errors.FieldNotFoundError: no field 'pt' in record with 1 fields`.
This error should not happen. `print(ak.max(pt.compute()))` works fine.
I've also observed the error with `dak.min` and `dak.is_none`.

dask_awkward version 2023.5.0
awkward version 2.2.0
Coffea version 2023.4.0.rc3

lgray:
@jrueb it would help if you can reproduce this without coffea to pin down where the error is. Usually stuff like this ends up being related to typetracers, which is actually down in awkward.

You can see an example of making some fake data to play with, using a form, here:
https://github.com/scikit-hep/awkward/issues/2416

Then you can work up to dask_awkward and a file if there's no problem in awkward itself, etc.

lgray:
Related to #256

agoose77:
I think this can be closed now?

jrueb:
> I think this can be closed now?

It seems there was a conclusion in https://github.com/scikit-hep/awkward/issues/2463, which I can't fully follow. If this is a wontfix, how would I have to adjust my code in order to fix this?

lgray:
This can't be closed now because the fix is getting unproject_layout to work correctly, which we've had to revert for now because it doesn't. We are definitely fixing this, just taking longer and with more churn than thought at first. I would mark the underlying issue a priority for August @agoose77 @douglasdavis.

douglasdavis:
Yeah this isn't something where we can suggest an adjustment to the failing code (except to turn off column optimization, which is absolutely undesireable). @lgray summarized the situation -- I'm currently working on figuring out the bugs in `unproject_layout`/"form rehydration"

douglasdavis:
Main issue being tracked in #313

PR #379, opened 2023-10-03
feat: implement `shape_touched` optimisation

### LINKED ISSUE #313: Computes that require shape only fail after form rehydration (count, zeros_like, ones_like, ...)

agoose77:
Wider ecosystem tasks required:
- [x] Update dask-histogram `mock` implementation
- [x] Update uproot5 `_dask` layer
- [ ] Update coffea's form remapping

This PR will hopefully allow us to close #313 and #260, by handling `shape_touched` in a proper fashion.

> **Warning**
> Requires scikit-hep/uproot5#966, CoffeaTeam/coffea#900, and awkward `main`

The high-level overview of this PR is as follows:
- Move the column optimisation mechanism to the input layer `io_func` implementations rather than in the `AwkwardInputLayer`.
- Operate at the level of "buffers" instead of "columns" for column optimisation
- Add helper functions to annotate forms and determine which buffers should be touched to compute the shape of another buffer


douglasdavis:
awesome, @agoose77! let me know if you'd like me to start going through this now while it's still a draft, or if I should wait a bit

agoose77:
@douglasdavis yes - it's not far off I think. There's a failing ROOT test that I need to fix, but I'll get to that shortly :)

src/dask_awkward/lib/utils.py
```diff
@@ -0,0 +1,135 @@
...
+        # NumPy like array is easy
+        elif form.is_numpy or form.is_unknown:
+            pass
+
+        # Anything else grab the content and keep recursing
+        else:
+            impl(form.content, f"{key}.content")
+
+    # Perform a "deep" copy without preserving references
+    form = ak.from_dict(ak.to_dict(form))
```
douglasdavis, line 133: this line isn't working for me with `awkward@main` installed; should it be:?
```suggestion
    form = ak.forms.from_dict(form.to_dict())
```

tests/test_core.py
```diff
@@ -332,10 +332,14 @@ def test_record_str(daa: Array) -> None:
 
 
 def test_record_to_delayed(daa: Array) -> None:
+    dask.config.set({"awkward.raise-failed-meta": True})
+    dask.config.set({"awkward.optimization.on-fail": "raise"})
```
douglasdavis, line 336: setting the dask config like this is global; should we use the context manager option?:
```python
def test():
    with dask.config.set(
        {
            "awkward.raise-failed-meta": True,
            "awkward.optimization.on-fail": "raise",
        }
    ):
        # ... rest of test
```

src/dask_awkward/layers/layers.py
```diff
@@ -24,15 +23,20 @@ def from_blockwise(cls, layer: Blockwise) -> AwkwardBlockwiseLayer:
         ob.__dict__.update(layer.__dict__)
         return ob
 
-    def mock(self) -> tuple[AwkwardBlockwiseLayer, Any | None]:
+    def mock(self) -> AwkwardBlockwiseLayer:
```
douglasdavis, line 26: The `AwkwardInputLayer` class inherits from `AwkwardBlockwiseLayer` and has a different `mock` method return type. Since `state` appears to usually be a `dict`; should the `mock` methods that were originally returning `None` just return an empty dict?

src/dask_awkward/layers/layers.py
```diff
@@ -213,11 +187,11 @@ def __init__(
         self.fn = fn
         super().__init__(mapping, **kwargs)
 
-    def mock(self) -> tuple[MaterializedLayer, Any | None]:
+    def mock(self) -> MaterializedLayer:
         mapping = copy.copy(self.mapping)
         if not mapping:
             # no partitions at all
-            return self, None
```
douglasdavis, line 220: Here's a line where I'm wondering if we should return `self, {}` to keep the `mock` interface something that returns a two element tuple

douglasdavis (changes requested):
Thanks for all of this work, @agoose77 ! I still have more to read (especially what's in the `lib/utils.py` module), but just wanted to throw up a few initial questions

src/dask_awkward/layers/layers.py
agoose77: Ooh! I missed this. Hmm.

tests/test_core.py
agoose77: In fact, this was just for local testing ... I should remove this.

src/dask_awkward/layers/layers.py
agoose77: I've taken a slightly different tack in e0827e9 — add a new `prepare_for_projection` method that `mock` can call.

agoose77:
@douglasdavis @lgray this PR _removes_ `dak.necessary_columns`. The reason for this is that these buffer keys are more input-layer implementation specific than columns, which correspond to form fields.

I can see there being some debugging benefit to e.g. `dak.necessary_buffers`, do you feel strongly on this? (I'm assuming yes, and will restore this).

douglasdavis:
> @douglasdavis @lgray this PR _removes_ `dak.necessary_columns`. The reason for this is that these buffer keys are more input-layer implementation specific than columns, which correspond to form fields.
> 
> I can see there being some debugging benefit to e.g. `dak.necessary_buffers`, do you feel strongly on this? (I'm assuming yes, and will restore this).

It's something we should probably still have, and generating the list should probably be something the `ImplementsIOFunctionWithProjection` supports, and `dak.necessary_columns` can just call that interface

agoose77:
> It's something we should probably still have, and generating the list should probably be something the `ImplementsIOFunctionWithProjection` supports, and `dak.necessary_columns` can just call that interface


OK, this was the push I needed. I've added a `dak.report_necessary_buffers`, which returns a mapping of layer names to namedtuples (or None, if the layer doesn't support projection / isn't an input layer).

These keys are still scoped to the individual layers, but we can choose sensible values. The parquet, JSON, and unremapped Uproot IO layers use `dask.lib.utils.form_with_unique_keys`, which uses a simple `.content`, `union#0`, `record.field` syntax (plus the standard attributes). Remapped Uproot layers use the remapped keys, which might be e.g. coffea's syntax.

I think that's fine; we could add a mechanism to the projection interface to "prettify" these buffer names into e.g. columns or otherwise human readable strings, but I envisage this being more useful for testing than interrogation.

douglasdavis:
It looks like we have a bit of a regression in the interpretation of what are actually necessary pieces of data.

Using this utility function I've written, we can inspect how we're going to actually call awkward's Parquet reader:

```python
from __future__ import annotations

import dask

import dask_awkward as dak
from dask_awkward.lib.optimize import optimize


def find_and_breakdown_input_layer(
    collection: dak.Array,
    layer_name_prefix: str,
) -> tuple[Callable, Any]:
    """Get first partition input layer callable and argument

    This function does a little surgery on the task graph to grab the
    input layer (partition 0), extract the callable at that layer, and
    also grab the argument that will be passed to that callable. We
    then return that callable and the argument for further inspection.

    We optimize the graph with _only_ the columns optimization. The
    layer chaining optimization makes it a little harder to perform
    the task graph surgery to isolate the input layer callable.

    Parameteters
    ------------
    collection : dask_awkward.Array
        dask-awkward Array collection of interest
    layer_name_prefix : str
        Prefix used to filter layers and determine the input layer.
        For example, a JSON input layer will start with "from-json"
        and a Parquet input layer will start with "from-parquet".

    Returns
    -------
    Callable
        The callable at the input layer
    str
        The string argument to the callable

    """
    with dask.config.set({"awkward.optimization.which": ["columns"]}):
        optimized_hlg = optimize(collection.dask, [])
        layers = list(optimized_hlg.layers)
        layer_name = [name for name in layers if name.startswith(layer_name_prefix)][0]
        sgc, arg = optimized_hlg[(layer_name, 0)]
        fn = sgc.dsk[layer_name][0]

    return fn, arg
```

Take for example a graph that is going to call `ones_like` on a single field in a record array:

```python
a = dak.from_parquet("s3://ddavistemp/hpq/", storage_options={"anon": True})
ones = dak.ones_like(a.muons.charge)
```

We can run our little utility function on this collection to see how it's input layer will be optimized:

```python
fn, arg = find_and_breakdown_input_layer(ones, "from-parquet")
print(fn)
```

If I run this code with the latest release of dask-awkward (2023.9.3) we see (notice the field that is actually grabbed is `pt`, while we asked for `charge`, but we know that since we're calling `ones_like` we don't actually need the data, just the offsets):

```python
FromParquetFn(
  form=RecordForm([ListOffsetForm('i32', RecordForm([NumpyForm('float64')], ['pt']))], ['muons'])
  listsep=list.item
  unnamed_root=False
  columns=['muons.list.item.pt']
  behavior=None
  max_gap=64000
  max_block=256000000
  footer_sample_size=1000000
  generate_bitmasks=False
)
```
We'll read only `muons.list.item.pt` from the Parquet files.

If I install this PR's branch we see:

```python
FromParquetFn(
  form=RecordForm([ListOffsetForm('i32', RecordForm([NumpyForm('float64'), NumpyForm('float64'), NumpyForm('float64'), NumpyForm('float64'), NumpyForm('int64')], ['pt', 'eta', 'phi', 'mass', 'charge']))], ['muons'])
  listsep=list.item
  unnamed_root=False
  columns=['muons.list.item.pt', 'muons.list.item.eta', 'muons.list.item.phi', 'muons.list.item.mass', 'muons.list.item.charge']
  behavior=None
  max_gap=64000
  max_block=256000000
  footer_sample_size=1000000
  generate_bitmasks=False
)
```
We're reading the entire `muons` record.

So I think we need to add some logic to "pick" a field. This has come up in previous discussions. In the latest released dask-awkward we just pick the first field when this case comes up; but we can be probably be more intelligent and pick the most efficient option based on type

src/dask_awkward/layers/layers.py
```diff
@@ -31,20 +33,14 @@ def mock(self) -> AwkwardBlockwiseLayer:
...
-    def __repr__(self) -> str:
-        return "Awkward" + super().__repr__()
-
     def __getstate__(self) -> dict:
-        d = self.__dict__.copy()
-        import pickle
+        # Indicator that this layer has been serialised
+        state = self.__dict__.copy()
+        state["has_been_unpickled"] = True
+        return state
```
agoose77, line 40: I added this to detect when a layer has been unpickled, in order to disable column optimisation. Does this align with your expectations?

douglasdavis: I'll ping @martindurant since he wrote the original `__getstate__`

douglasdavis: I can't think of why we'd want to disable optimization if a layer has been pickled. Has this come up in discussion before?

agoose77: I don't know _why_ the `_meta` was stored in the first place — it doesn't seem like we were reading it anywhere. My interpretation was that we don't want to run column optimisation if we evaluate the graph on the scheduler because it would already have been optimised. I don't know how one would even enter into the column optimisation routines to begin with, though, so this interpretation is likely flawed.

src/dask_awkward/lib/io/columnar.py
```diff
@@ -86,55 +87,69 @@ def project(
...
+                # Ensure we get the "actual" last form key
+                for last_record_form_key in record_form_keys_with_contents:
+                    ...
+
+            # First see if any child is already included
+            for any_child_form_key in form_key_to_child_form_keys[last_record_form_key]:
+                any_child_path = form_key_to_path[any_child_form_key]
+                if any_child_path in paths:
+                    break
+            # Otherwise, add the last child
```
agoose77: @douglasdavis what do you think of this optimisation?

This is the simplest "find a leaf" that I could come up with. I originally played around with computing a heuristic for the mean buffer size of the path to the leaf, but I think it's probably not worth the complexity; any index nodes will immediately throw off the estimate. It might be reasonable to assume that most input sources won't have any index nodes, but it is a lot more work.

src/dask_awkward/layers/layers.py
douglasdavis: [`self._meta`](https://github.com/dask-contrib/dask-awkward/blob/5981337650231e5bde00bc0cf9c8b9358e86404c/src/dask_awkward/layers/layers.py#L76) is getting assigned right now in `AwkwardInputLayer` (which inherits from `AwkwardBlockwiseLayer`); we were using it to instantiate a new form, followed by a new typetracer array to run the typetracer only graph after a call of the `mock` method. I think Martin wrote the code to pop the _meta attribute upon serialization because TypetracerArray wasn't serializable. I think with the refactoring you've done we no longer need to worry about `_meta` being a part of the layer, correct?

agoose77: `_meta` can still be within the layer as an attribute of e.g. the `io_func`, but not directly in `self._meta` for the `AwkwardInputLayer`.

agoose77:
@douglasdavis I added support for a new feature - columnar sources. IO functions that are "columnar" have a mechanism of interpreting the column projection in terms of "columns". 

JSON, Parquet, and uproot support this; though, in the latter case, ROOT keys are only ever "top-level" columns.

src/dask_awkward/lib/io/columnar.py
```diff
@@ -0,0 +1,141 @@
...
+
+    def project(
+        self: T,
+        report: TypeTracerReport,
+        state: FormStructure,
+    ) -> T:
+        if not self.use_optimization:
+            return self
+
+        return self.project_columns(self.necessary_columns(report, state))
```
douglasdavis, line 141: Looks like we're relying on both the JSON and Parquet reader function classes to have a `project_columns` method, should we enforce that in this mixin?:

```suggestion
        return self.project_columns(self.necessary_columns(report, state))
        
    @abc.abstractmethod
    def project_columns(columns: Sequence[str]) -> Self:
        ...
```

src/dask_awkward/__init__.py
```diff
@@ -16,7 +16,11 @@
     partition_compatibility,
 )
 from dask_awkward.lib.describe import fields
-from dask_awkward.lib.inspect import necessary_columns, sample
```
douglasdavis, line 19: I think it makes sense to (for now) keep a `dak.necessary_columns` that just calls `report_necessary_columns`; we should deprecate it, but this has been in for a little while now and it might be best to keep it around for a while. We've never had a formal deprecation process in dask-awkward, but now might be a time to start!

douglasdavis (changes requested):
This is looking pretty great! Added a couple more small comments. At this point I'm wondering how much @lgray may have tortured the work here up til this point?

src/dask_awkward/lib/io/columnar.py
douglasdavis: OK this seems reasonable to me; maybe one day we can revisit trying to find an int leaf over e.g. a string leaf, but this should be on par with what we were already doing, so sounds good to me!

src/dask_awkward/lib/io/columnar.py
agoose77: I'm torn on this. We also require behavior` and `form` attributes. I think it's conventional to not add `abc` methods to mixins,

I've made a new PR that treats `self` as a protocol implementer.

douglasdavis: Ah yes this is a mixin 🤦 my mind was in interface mode, ignore!

douglasdavis:
looks like that uproot test grabbing the file over http is causing CI problems, it's really small so let's go ahead and commit it to the repository

agoose77:
I *think* the CI was struggling because an older (PyPI) uproot was being pulled in; this *works* but doesn't support buffer optimisation, so it reads the entire file. This is why we haven't seen the error before!

douglasdavis:
should we increase the `awkward` version req to `>=2.4.5`?

docs/api/inspect.rst
```diff
@@ -8,7 +8,7 @@ Inspection
 
    partition_compatibility
    PartitionCompatibility
-   necessary_columns
+   report_necessary_buffers
```
agoose77, line 11:
```suggestion
   report_necessary_buffers
   report_necessary_columns
```

douglasdavis:
awesome! thanks @agoose77

PR #383, opened 2023-10-09
fix: harden typing in the code base; add pre-commit check; improve docs

douglasdavis:
- [x] fix various existing typing errors
- [x] require mypy check with pre-commit
- [x] add developer docs explaining that while typing is not a requirement in dask-awkward, if it's used in a signature we want to make sure it's checked

we've always had type hints throughout dask-awkward but it's never been formalized, this PR aims to clarify typing in the project.

### DIFF
--- a/src/dask_awkward/layers/layers.py
+++ b/src/dask_awkward/layers/layers.py
@@ -44,7 +44,7 @@ class AwkwardBlockwiseLayer(Blockwise):
 
 
 class ImplementsIOFunction(Protocol):
-    def __call__(self, *args, **kwargs):
+    def __call__(self, *args: Any, **kwargs: Any) -> AwkwardArray:
         ...
 
 
--- a/src/dask_awkward/layers/layers.py
+++ b/src/dask_awkward/layers/layers.py
@@ -44,7 +44,7 @@ class AwkwardBlockwiseLayer(Blockwise):
 
 
 class ImplementsIOFunction(Protocol):
-    def __call__(self, *args, **kwargs) -> AwkwardArray:
+    def __call__(self, *args, **kwargs):
         ...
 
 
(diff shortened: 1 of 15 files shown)


src/dask_awkward/layers/layers.py
```diff
@@ -220,7 +220,7 @@ def project(
         self,
         report: TypeTracerReport,
         state: T,
-    ):
+    ) -> AwkwardInputLayer:
```
agoose77, line 223: Eek! My bad.

src/dask_awkward/lib/utils.py
```diff
@@ -125,12 +129,12 @@ def render_buffer_key(form: Form, form_key: str, attribute: str) -> str:
     return f"{form_key}-{attribute}"
 
 
-def parse_buffer_key(buffer_key: str) -> tuple[str, str]:
+def parse_buffer_key(buffer_key: str) -> list[str]:
     return buffer_key.rsplit("-", maxsplit=1)
```
agoose77, line 133:
```suggestion
def parse_buffer_key(buffer_key: str) -> tuple[str, str]:
    head, tail = buffer_key.rsplit("-", maxsplit=1)
    return head, tail
```

src/dask_awkward/layers/layers.py
```diff
@@ -44,7 +44,7 @@ def __repr__(self) -> str:
 
 
 class ImplementsIOFunction(Protocol):
-    def __call__(self, *args, **kwargs) -> AwkwardArray:
+    def __call__(self, *args, **kwargs):
```
agoose77, line 47: Why drop the return?

douglasdavis: Since `*args` and `**kwargs` are untyped we get a mypy error; it's something I'd like to revisit if we can better hint args and kwargs here, I just hadn't thought about it yet, but I supposed we could put `Any` for now

PR #385, opened 2023-10-10
fix: declare pyarrow optional for 3.12

agoose77:
I noticed that we were failing in CI due to Arrow being installed from sdist on Python 3.12.

In future, perhaps these large binary dependencies should not be installed from source as a rule. That might be a forward-looking thing to do, we also need to just ensure we don't install PyArrow here, because it's not yet available on 3.12.

This PR changes our CI to avoid pulling in `pyarrow` via `dask`, and also ensures that our own `pyproject.toml` just skips `pyarrow` and `aiohttp` if we are on 3.12

### DIFF
--- a/.github/workflows/coverage.yml
+++ b/.github/workflows/coverage.yml
@@ -32,7 +32,7 @@ jobs:
       run: |
         pip install wheel
         pip install pytest-cov
-        pip install dask[complete]
+        pip install dask[array,dataframe,distributed,diagnostics]
         pip install -q --no-cache-dir -e .[complete,test]
         pytest --cov=dask_awkward --cov-report=xml
     - name: Upload Coverage to Codecov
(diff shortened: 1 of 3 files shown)


agoose77:
@douglasdavis would you mind taking a look at our dependence upon pyarrow? I assume we want it to be optional here, even in our tests if it's not installed?

douglasdavis:
> @douglasdavis would you mind taking a look at our dependence upon pyarrow?

@agoose77, sure! as of right now the sample datasets used for a lot of tests are instantiated in a pytest fixture via a `dak.from_parquet` call; which is why `pyarrow` is a hard dependency to run the test suite

> I assume we want it to be optional here, even in our tests if it's not installed?

I think this would be a good idea in the long run. A short task would be to `pytest.importorskip("pyarrow")` but this would lead to many tests getting skipped. A longer but better task would be to transition away from using small parquet sample datasets for running the tests to small-something-else. Originally I wrote small JSON samples, but since parquet supported column optimization I transitioned to using Parquet samples just to get a little more implicit testing of column optimization.

I'm happy to spend some time thinking and working on this

PR #388, opened 2023-10-16
ci: bump Python use of 3.10 to 3.11 in CI; drop Python 3.8

douglasdavis:


agoose77:
As a maintainer, I'm in favour of this, However, Scientific Python is more aggressive with Python versions than e.g. Awkward Array is at the moment. @lgray are there any experiment restrictions on Python versions that would warrant us reverting this PR?

lgray:
python 3.8 is all of our present images, is in common CMSSW releases, and allowed by coffea. Because of this I was planning to follow the standard python 3.8 end of life later in 2024.

We do have python 3.10 images coming out now, and I'm relatively happy so long as a functioning version of dask-awkward is available python 3.8... I'd prefer to have bugfixes available there for a while though.

lgray:
Actually - yes @douglasdavis please revert the part of this where we drop python 3.8. We will certainly find bugs as we release this for more users to try, and there's going to be a good portion of them effectively stuck with python 3.8 for one reason or another.

douglasdavis:
OK sounds good, I mistakenly thought awkward-array followed Scientific Python versioning, apologies!

PR #395, opened 2023-10-24
feat: deprecate `opt_touch_all` from `map_partitions`.

douglasdavis:
The `opt_touch_all` argument to `map_partitions` was added to make layers manually touchable (that is, tell the optimization code to not rely on the typetracer graph for touching in that layer, just touch all awkward data at that point). This allowed us to workaround operations where automatic typetracer touching wasn't working. 

It was only used with `dak.zip`, but with the work on upstream typetracer and the optimization refactor here in dask-awkward, I think we're in a place now where we don't need it.

It's currently used in coffea, see: https://github.com/CoffeaTeam/coffea/issues/915

### DIFF
--- a/src/dask_awkward/lib/core.py
+++ b/src/dask_awkward/lib/core.py
@@ -1572,7 +1572,6 @@ def partitionwise_layer(
     func: Callable,
     name: str,
     *args: Any,
-    opt_touch_all: bool = False,
     **kwargs: Any,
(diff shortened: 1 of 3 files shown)


agoose77:
@lgray can we merge this and bump coffea?

agoose77:
@lgray has given this the green light!

PR #400, opened 2023-10-31
feat: support returning empty arrays in `from_map` IO function calls based on user declared exceptions

douglasdavis:
Discussed on Slack, this is a way to give `from_map` callers the ability to provide a list of "allowed" exceptions that should just return an empty array. It requires the `io_func` to have a `form` attribute such that a correctly formed empty array can be instantiated. cc @lgray 

To use this new feature there are two requirements:

1. Write a `mock_empty` method (that takes a single argument that should be passed to `ak.to_backend`) in the class function passed to `from_map`. (The `ColumnarProjectionMixin` class in the `lib.io.columnar` module provides an example implementation).
2. pass values to the `empty_on_raise` and `empty_backend` arguments of `from_map`. `tests/test_io.py` has example `from_map` calls that includes values for these arguments

### DIFF
--- a/src/dask_awkward/layers/layers.py
+++ b/src/dask_awkward/layers/layers.py
@@ -93,7 +93,7 @@ class IOFunctionWithMocking(ImplementsMocking, ImplementsIOFunction):
         assert self._meta is not None
         return self._meta
 
-    def mock_empty(self, backend: str = "cpu") -> AwkwardArray:
+    def mock_empty(self, backend: BackendT = "cpu") -> AwkwardArray:
         import awkward as ak
 
         if backend not in ("cpu", "jax", "cuda"):
--- a/src/dask_awkward/layers/layers.py
+++ b/src/dask_awkward/layers/layers.py
@@ -55,6 +55,9 @@ class ImplementsMocking(Protocol):
     def mock(self) -> AwkwardArray:
         ...
 
+    def mock_empty(self, backend: str) -> AwkwardArray:
+        ...
+
 
 class ImplementsProjection(ImplementsMocking, Protocol[T]):
     def prepare_for_projection(self) -> tuple[AwkwardArray, TypeTracerReport, T]:
@@ -86,13 +89,26 @@ class IOFunctionWithMocking(ImplementsMocking, ImplementsIOFunction):
         assert self._meta is not None
         return self._meta
 
+    def mock_empty(self, backend: str = "cpu") -> AwkwardArray:
+        import awkward as ak
+
+        if backend not in ("cpu", "jax", "cuda"):
+            raise ValueError(
+                f"backend must be one of 'cpu', 'jax', or 'cuda', received {backend}"
+            )
+        return ak.to_backend(
+            self._meta.layout.form.length_zero_array(highlevel=False),
+            backend=backend,
+            highlevel=True,
+        )
+
 
 def io_func_implements_projection(func: ImplementsIOFunction) -> bool:
     return hasattr(func, "prepare_for_projection")
 
 
 def io_func_implements_mocking(func: ImplementsIOFunction) -> bool:
-    return hasattr(func, "mock")
+    return hasattr(func, "mock") and hasattr(func, "mock_empty")
 
 
 def io_func_implements_columnar(func: ImplementsIOFunction) -> bool:
--- a/src/dask_awkward/layers/layers.py
+++ b/src/dask_awkward/layers/layers.py
@@ -2,17 +2,21 @@ from __future__ import annotations
 
 import copy
 from collections.abc import Callable, Mapping
-from typing import TYPE_CHECKING, Any, Protocol, TypeVar, cast
+from typing import TYPE_CHECKING, Any, Literal, Protocol, TypeVar, Union, cast
 
 from dask.blockwise import Blockwise, BlockwiseDepDict, blockwise_token
 from dask.highlevelgraph import MaterializedLayer
 from dask.layers import DataFrameTreeReduction
+from typing_extensions import TypeAlias
 
 from dask_awkward.utils import LazyInputsDict
 
 if TYPE_CHECKING:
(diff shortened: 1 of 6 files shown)


douglasdavis:
This currently lacks the ability to keep track of what arguments fail.

src/dask_awkward/lib/io/io.py
```diff
@@ -460,6 +464,21 @@ def __call__(self, packed_arg):
...
+    fn: Callable[P, Array],
+    allowed_exceptions: tuple[type[BaseException], ...],
+) -> Callable[P, Array]:
+    @functools.wraps(fn)
+    def wrapped(*args, **kwargs):
+        try:
+            result = fn(*args, **kwargs)
+            return result
+        except allowed_exceptions:
+            return ak.Array(fn.form.length_zero_array(highlevel=False))
```
agoose77, line 477: We also have the existing `.mock` interface: https://github.com/douglasdavis/dask-awkward/blob/a368106dd57c4a355022db42ab9f97f853e2f566/src/dask_awkward/layers/layers.py#L54-L56

I wonder if either we should extend this with another method `mock_empty()`, or whether we should require only `mock` and take the form from the mocked array: `fn.mock().layout.form.length_zero_array()`

One thing to be mindful of is that we should allow the implementer to choose the backend.

douglasdavis: Ah yes I like the idea of reusing mock for this task, thanks for peaking at this

douglasdavis: Since `mock` has been around for a while and has been a typetracer-only part of the code, perhaps it is best to go with a new `mock_empty(backend: str)` method

lgray: Concerning the backend configurability: we should probably forbid the implementer to inject the typetracer backend as for this purpose it just doesn't make any sense?

douglasdavis:
Alright I've reworked the PR to use a new `mock_empty` method which is designed to be restricted to one of `("cpu", "jax", "cuda")`

douglasdavis:
We also need a way to communicate the backend to `return_empty_on_raise`

src/dask_awkward/layers/layers.py
```diff
@@ -86,13 +93,26 @@ def mock(self) -> AwkwardArray:
         assert self._meta is not None
         return self._meta
 
+    def mock_empty(self, backend: BackendT = "cpu") -> AwkwardArray:
+        import awkward as ak
+
+        if backend not in ("cpu", "jax", "cuda"):
```
lgray, line 99: FWIW, despite this being a bit of an abuse of a private interface you could make a set out of the keys of https://github.com/scikit-hep/awkward/blob/main/src/awkward/_backends/dispatch.py#L24 
and substract "typetracer" from the set and use that here. Should remove and problems coming from more backends over time.

@agoose77 may have issue with this though given it being hacky.

src/dask_awkward/lib/io/io.py
```diff
@@ -565,21 +582,20 @@ def from_map(
     if io_func_implements_mocking(func):
         io_func = func
         array_meta = cast(ImplementsMocking, func).mock()
-    # If we know the meta, we can spoof mocking
+    # # If we know the meta, we can spoof mocking
     elif meta is not None:
-        io_func = IOFunctionWithMocking(meta, func)
+        # io_func = IOFunctionWithMocking(meta, func)
+        io_func = func
```
douglasdavis, line 588: @agoose77 - what was the need for this part of the `from_map` code (wrapping `io_func` in `IOFunctionWithMocking`). I ran into a problem here when writing a test for the new feature of this PR. If `mock` wasn't defined but `meta` was, If I tried to use `self.form` in the `io_func` that I defined, I lost the ability to get the `form` attr from my `io_func` because it was now wrapped by `IOFunctionWithMocking` which only has two attrs: `_io_func` and and `_meta`. I removed this problem by utilizing `mock` (as you suggested!) as part of this PR, but it was still a confusing thing to run into to see the `io_func` that I wrote get covered up by something else.

I'm also thinking it's unnecessary to bolt on an available `mock` method if it's not already there. If consuming code wants to support necessary-data-projection that consuming code will definitely need an `io_func` with a `mock` method

agoose77: This was written from first-principles (and to support the existing API), rather than as a use-case.

I think it addresses the case that there are _two_ input layers; one that supports column optimisation, and the other which does not. By allowing `meta` to be used during `mock`, we can avoid computing the non-optimising input during optimisation of the graph. 

This might be useful for users who cannot selectively read but do know the type of their data up-front.

As such, I think it's an important feature

douglasdavis: Thanks for the explanation! That does make sense w.r.t. multiple (mixed source) input layers. It would be quite strict to require all io funcs to be function classes (for example `read_text` just uses a standalone function and derives a meta, so it would use this feature if combined with another optimizable input layer). Seems like things should stay as-is

src/dask_awkward/lib/io/io.py
```diff
@@ -460,15 +457,31 @@ def __call__(self, packed_arg):
...
+    backend: BackendT,
+) -> Callable:
+    @functools.wraps(fn)
+    def wrapped(*args, **kwargs):
+        try:
+            return fn(*args, **kwargs)
+        except allowed_exceptions:
+            return fn.mock_empty(backend)
+
+    return wrapped
```
agoose77, line 472: We discussed adding a report mechanism here, so that the partitions can return a `Report` object (that perhaps the IO func can populate).

This would mean at a high level that we don't build an `dak.Array` from the `io_func` applications; rather, we need an unpacking step so that we build two subgraphs that depend upon the `io_func` result.

lgray:
Great! This gives enough to prototype something that's a good analogue of what we have presently in coffea.

PR #402, opened 2023-11-02
fix: use duck-typing check in place of `isinstance(..., np.ndarray)`

douglasdavis:
Followup to #401

### DIFF
--- a/src/dask_awkward/lib/core.py
+++ b/src/dask_awkward/lib/core.py
@@ -534,10 +534,21 @@ def new_record_object(dsk: HighLevelGraph, name: str, *, meta: Any) -> Record:
     return Record(dsk, name, meta)
 
 
+def _is_numpy_or_cupy_like(arr: Any) -> bool:
+    return (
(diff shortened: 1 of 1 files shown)


src/dask_awkward/lib/core.py
```diff
@@ -534,10 +534,20 @@ def new_record_object(dsk: HighLevelGraph, name: str, *, meta: Any) -> Record:
...
+        and hasattr(arr, "dtype")
+    )
+
+
 def _finalize_array(results: Sequence[Any]) -> Any:
     # special cases for length 1 results
     if len(results) == 1:
-        if isinstance(results[0], (int, ak.Array, np.ndarray)):
+        np_like = _is_numpy_or_cupy_like(results[0])
+        if isinstance(results[0], (int, ak.Array)) or np_like:  # type: ignore[unreachable]
```
douglasdavis, line 550: I have no idea why mypy is complaining that the second half of this `or` is unreachable. It's absolutely reachable! `results[0]` is an `Any`.

PR #405, opened 2023-11-08
feat!: error if the caller invokes np.XXX on an array

### LINKED ISSUE #404: bug: reducers from numpy no longer seem to work?
Tests with this sort of syntax were passing until  dask-awkward 2023.11.0.

@agoose77 @douglasdavis 

```python3
import numpy as np
import uproot

events = uproot.dask("https://github.com/CoffeaTeam/coffea/raw/master/tests/samples/nano_dy.root:Events")

np.all(events.Muon_pt > 30)
```

it now fails with:
```
  File "<__array_function__ internals>", line 200, in all
  File "/Users/lgray/miniforge3/envs/coffea-dev/lib/python3.8/site-packages/numpy/core/fromnumeric.py", line 2515, in all
    return _wrapreduction(a, np.logical_and, 'all', axis, None, out,
  File "/Users/lgray/miniforge3/envs/coffea-dev/lib/python3.8/site-packages/numpy/core/fromnumeric.py", line 86, in _wrapreduction
    return ufunc.reduce(obj, axis, dtype, out, **passkwargs)
  File "/Users/lgray/miniforge3/envs/coffea-dev/lib/python3.8/site-packages/dask_awkward/lib/core.py", line 1380, in __array_ufunc__
    raise RuntimeError("Array ufunc supports only method == '__call__'")
RuntimeError: Array ufunc supports only method == '__call__'
```

This is also true of np.sum, np.any, etc.

agoose77:
Closes #404 with an error message.

### DIFF
--- a/src/dask_awkward/lib/core.py
+++ b/src/dask_awkward/lib/core.py
@@ -1357,6 +1357,13 @@ class Array(DaskMethodsMixin, NDArrayOperatorsMixin):
         """Clear the divisions of a Dask Awkward Collection."""
         self._divisions = (None,) * (self.npartitions + 1)
 
+    def __array_function__(self, func, types, args, kwargs):
+        raise NotImplementedError(
+            "overloads of NumPy functions are not supported in dask-awkward. "
+            "Instead, use the Awkward Array equivalent, e.g. `ak.all` instead "
+            "of `np.all`"
+        )
+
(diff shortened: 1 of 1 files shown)


douglasdavis (approved):
Since this wasn't supported to begin I'd give this a green light. However, it looks like you may be cooking something else, @agoose77? (based on some chatter in Slack)

PR #409, opened 2023-11-11
feat: add new dask-behavior protocol

agoose77:
This PR is a suggestion for replacing `_dask_array_` with an extension to the descriptor protocol, namely `_dask_get`, `_dask_set`, etc.

The motivations for this are to avoid reserving a name within the function signatures, keep property descriptors looking like property descriptors, and to avoid the need to compute the function signature.

Unfortunately, if the fallback case of using `map_partitions` were implemented naively, we'd need to support a collection type that represents an `Any` type. As this isn't feasible, this PR uses a heuristic (as we already do) to determine whether the descriptor represents a method (in which case, delaying the call to map_partitions). The heuristic that I chose is to test whether the resolved attribute is `callable`.

If we are interested in this PR, then I'd probably want to add a fallback for coffea, unless they want to sync up again with coffea's latest release.

@lgray @douglasdavis

### DIFF
--- a/src/dask_awkward/lib/core.py
+++ b/src/dask_awkward/lib/core.py
@@ -69,7 +69,20 @@ T = TypeVar("T")
 log = logging.getLogger(__name__)
 
 
-def make_dask_descriptor(func: Callable) -> Callable[[T, type[T], Array], Any]:
+def _make_dask_descriptor(func: Callable) -> Callable[[T, type[T], Array], Any]:
+    """Adapt a function accepting a `dask_array` into a dask-awkward descriptor
+    that invokes and returns the user function when invoked.
+
+    Parameters
+    ----------
+    func : Callable dask-awkward descriptor body
+
+    Returns
+    -------
+    Callable
+        The callable dask-awkward descriptor
+    """
+
     def descriptor(instance: T, owner: type[T], dask_array: Array) -> Any:
         impl = func.__get__(instance, owner)
         return impl(dask_array)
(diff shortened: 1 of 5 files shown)


src/dask_awkward/lib/core.py
```diff
@@ -68,6 +68,56 @@
...
+
+
+F = TypeVar("F", bound=Callable)
+G = TypeVar("G", bound=Callable)
+
+
+class dask_property(property):
+    _dask_get: Callable | None = None
+    _dask_set: Callable | None = None
+    _dask_del: Callable | None = None
```
agoose77, line 97: I defined `__selattr__` and `__delattr__` although these wouldn't make sense in a functional API. We could come up with a `dak.setattr` and `dak.delattr` function that always returns the input array, i.e. `_dask_set` actually returns self, but I doubt we actually need `__delattr__` support.

lgray:
I'll need to poke at this in coffea to make sure it covers what's needed.

douglasdavis:
I'm fine with making a change like this (and I'll add that beyond behaviors "just working", I'd consider @lgray's opinion a more important one than mine w.r.t. the ergonomics)

lgray:
should I wait until all this is green to give it a check?

agoose77:
@lgray should be ready to go now!

lgray:
OK - there's something I need to make work first (the multi-return stuff that @iasonkrom found) and then I'll give this a proper looking over.

lgray:
https://github.com/dask-contrib/dask-awkward/actions/runs/6853762042/job/18635289824?pr=409#step:5:692 a problem?

agoose77:
> https://github.com/dask-contrib/dask-awkward/actions/runs/6853762042/job/18635289824?pr=409#step:5:692 a problem?

I also noticed this! It's unrelated (main): https://github.com/dask-contrib/dask-awkward/pull/413#issuecomment-1808725620

agoose77:
@lgray any thoughts on this? :)

agoose77:
@lgray I fixed a bug that caused your regression. Would you be happy to test this PR again?

agoose77:
This passes in coffea, so we can look to merging. @lgray would you be happy to pin `>=` to the latest dask-awkward?

lgray:
Yes - I've already pinned that PR to the next release of dask-awkward. Please go ahead and merge / release!

agoose77:
@douglasdavis I already pinged you for review but realised I didn't add context: this is good to go, are you happy to sign off on it?

douglasdavis (approved):
Seems like a nice API improvement to me!

PR #410, opened 2023-11-11
refactor: enforce more type hints

agoose77:
This is a small PR to enforce some more type hints

### DIFF
--- a/src/dask_awkward/lib/structure.py
+++ b/src/dask_awkward/lib/structure.py
@@ -2,7 +2,7 @@ from __future__ import annotations
 
 import builtins
 import warnings
-from collections.abc import Sequence
+from collections.abc import Iterable, Mapping, Sequence
 from numbers import Number
 from typing import TYPE_CHECKING, Any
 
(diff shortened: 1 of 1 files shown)


src/dask_awkward/lib/structure.py
```diff
@@ -319,10 +314,8 @@ def combinations(
 
 
 @borrow_docstring(ak.copy)
-def copy(array):
-    raise DaskAwkwardNotImplemented(
-        "This function is not necessary in the context of dask-awkward."
-    )
+def copy(array: Array) -> Array:
+    return array
```
douglasdavis, line 319: This could be a little confusing, no? Folks might expect to get a new object

agoose77: This change should have been louder and/or had a review from you, sorry.

My rationale is that we want `ak.XXX` to work on both `ak` and `dak` `Array` objects. Thus, we should still support the function. However, I now realise that with the way that coffea (ab)uses `behavior`, and now with `attrs`, we'll need to perform a shall copy. Let me make a new PR.

Otherwise, are you happy with this rationale?

PR #413, opened 2023-11-13
fix: `dak.copy` should copy metadata

agoose77:
My recent change to `dak.copy` made it an identity. Whilst that's useful _at the buffer level_ (dask-awkward has only pure functions at the user level) we still want to copy the metadata. This will become _more_ important with `Array.attrs`.


douglasdavis (approved):
LGTM!

PR #415, opened 2023-11-14
feat: add computable report when graceful-read-failure is active in `from_map`

douglasdavis:
This changes the `from_map` interface to return two collections when `empty_on_raise` and `empty_backend` are both not `None`. (When those arguments are both `None` we just return a single `Array`, which has been the existing behavior until this PR, hopefully the `@overload`s show this in a clear way)

- The first returned `Array` collection is what we've always had: the lazy awkward array populated by data on disk.
- The second returned `Array` is a record array, one element per partition, that represents a report about failed reads.

After banging my head thinking of ways to handle reports inside the graph I realized it's possible for us to just make it an awkward array (instead of some new object, the rabbit hole I originally went down). I still need to check if this breaks any optimizations, but my intuition is that it shouldn't. I still wouldn't bet the farm :)

cc @lgray @agoose77 for your thoughts. This PR demonstrates the possibility, and I think it fits in nicely with the existing `uproot.iterate(..., report=True)` two-item return pattern.

### DIFF
--- a/src/dask_awkward/layers/layers.py
+++ b/src/dask_awkward/layers/layers.py
@@ -107,13 +107,13 @@ class IOFunctionWithMocking(ImplementsMocking, ImplementsIOFunction):
         )
 
 
-def io_func_reor_wrapped(func: ImplementsIOFunction) -> bool:
-    return hasattr(func, "__reor_wrapped__")
+def io_func_rer_wrapped(func: ImplementsIOFunction) -> bool:
+    return hasattr(func, "__rer_wrapped__")
 
 
 def maybe_unwrap(func: Callable) -> Callable:
-    if io_func_reor_wrapped(func):
-        return func.__reor_wrapped__  # type: ignore
+    if io_func_rer_wrapped(func):
+        return func.__rer_wrapped__  # type: ignore
     return func
 
 
@@ -242,7 +242,7 @@ class AwkwardInputLayer(AwkwardBlockwiseLayer):
             ImplementsProjection, fn
         ).prepare_for_projection()
 
-        if io_func_reor_wrapped(self.io_func):
+        if io_func_rer_wrapped(self.io_func):
             new_return = (new_meta_array, type(new_meta_array)([]))
         else:
             new_return = new_meta_array
@@ -267,16 +267,8 @@ class AwkwardInputLayer(AwkwardBlockwiseLayer):
         fn = maybe_unwrap(self.io_func)
         io_func = cast(ImplementsProjection, fn).project(report=report, state=state)
 
-        if io_func_reor_wrapped(self.io_func):
-            from dask_awkward.lib.io.io import return_empty_on_raise
-
-            io_func = return_empty_on_raise(
-                io_func,
-                allowed_exceptions=self.io_func.allowed_exceptions,  # type: ignore
-                backend=self.io_func.backend,  # type: ignore
-                success_callback=self.io_func.success_callback,  # type: ignore
-                failure_callback=self.io_func.failure_callback,  # type: ignore
-            )
+        if io_func_rer_wrapped(self.io_func):
+            io_func = self.io_func.recreate(io_func)  # type: ignore
 
         return AwkwardInputLayer(
             name=self.name,
--- a/src/dask_awkward/layers/layers.py
+++ b/src/dask_awkward/layers/layers.py
@@ -4,6 +4,7 @@ import copy
 from collections.abc import Callable, Mapping
 from typing import TYPE_CHECKING, Any, Literal, Protocol, TypeVar, Union, cast
 
+import awkward as ak
 from dask.blockwise import Blockwise, BlockwiseDepDict, blockwise_token
 from dask.highlevelgraph import MaterializedLayer
 from dask.layers import DataFrameTreeReduction
@@ -107,13 +108,13 @@ class IOFunctionWithMocking(ImplementsMocking, ImplementsIOFunction):
         )
 
 
-def io_func_rer_wrapped(func: ImplementsIOFunction) -> bool:
-    return hasattr(func, "__rer_wrapped__")
+def io_func_empty_on_error_wrapped(func: ImplementsIOFunction) -> bool:
+    return hasattr(func, "_empty_on_error_wrapped")
 
 
 def maybe_unwrap(func: Callable) -> Callable:
-    if io_func_rer_wrapped(func):
-        return func.__rer_wrapped__  # type: ignore
+    if io_func_empty_on_error_wrapped(func):
+        return func._empty_on_error_wrapped  # type: ignore
     return func
 
 
@@ -242,8 +243,8 @@ class AwkwardInputLayer(AwkwardBlockwiseLayer):
             ImplementsProjection, fn
         ).prepare_for_projection()
 
-        if io_func_rer_wrapped(self.io_func):
-            new_return = (new_meta_array, type(new_meta_array)([]))
+        if io_func_empty_on_error_wrapped(self.io_func):
+            new_return = (new_meta_array, ak.from_iter([]))
         else:
             new_return = new_meta_array
 
@@ -267,7 +268,7 @@ class AwkwardInputLayer(AwkwardBlockwiseLayer):
         fn = maybe_unwrap(self.io_func)
         io_func = cast(ImplementsProjection, fn).project(report=report, state=state)
 
-        if io_func_rer_wrapped(self.io_func):
+        if io_func_empty_on_error_wrapped(self.io_func):
             io_func = self.io_func.recreate(io_func)  # type: ignore
 
         return AwkwardInputLayer(
--- a/src/dask_awkward/layers/layers.py
+++ b/src/dask_awkward/layers/layers.py
@@ -113,7 +113,7 @@ def io_func_reor_wrapped(func: ImplementsIOFunction) -> bool:
 
 def maybe_unwrap(func: Callable) -> Callable:
     if io_func_reor_wrapped(func):
-        return func.__reor_wrapped__
+        return func.__reor_wrapped__  # type: ignore
     return func
 
 
@@ -272,10 +272,10 @@ class AwkwardInputLayer(AwkwardBlockwiseLayer):
 
             io_func = return_empty_on_raise(
                 io_func,
-                allowed_exceptions=self.io_func.allowed_exceptions,
-                backend=self.io_func.backend,
-                success_callback=self.io_func.success_callback,
-                failure_callback=self.io_func.failure_callback,
+                allowed_exceptions=self.io_func.allowed_exceptions,  # type: ignore
+                backend=self.io_func.backend,  # type: ignore
+                success_callback=self.io_func.success_callback,  # type: ignore
+                failure_callback=self.io_func.failure_callback,  # type: ignore
             )
 
         return AwkwardInputLayer(
--- a/src/dask_awkward/layers/layers.py
+++ b/src/dask_awkward/layers/layers.py
@@ -107,20 +107,30 @@ class IOFunctionWithMocking(ImplementsMocking, ImplementsIOFunction):
         )
 
 
+def io_func_reor_wrapped(func: ImplementsIOFunction) -> bool:
+    return hasattr(func, "__reor_wrapped__")
+
+
+def maybe_unwrap(func: Callable) -> Callable:
+    if io_func_reor_wrapped(func):
+        return func.__reor_wrapped__
+    return func
+
+
 def io_func_implements_projection(func: ImplementsIOFunction) -> bool:
-    return hasattr(func, "prepare_for_projection")
+    return hasattr(maybe_unwrap(func), "prepare_for_projection")
 
 
 def io_func_implements_mocking(func: ImplementsIOFunction) -> bool:
-    return hasattr(func, "mock")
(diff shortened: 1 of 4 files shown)


src/dask_awkward/lib/io/io.py
```diff
@@ -639,6 +687,11 @@ def from_map(
     else:
         result = new_array_object(hlg, name, meta=array_meta, npartitions=len(inputs))
 
+    if empty_on_raise and empty_backend:
+        res = result.map_partitions(first, meta=array_meta, output_divisions=1)
+        rep = result.map_partitions(second, meta=empty_typetracer())
```
douglasdavis, line 692: Here `res` is the actual interesting `Array` collection. The `array_meta` object is the meta we'd normally expect without going through the process of dual returns with the report. We can manually pass that meta object to `map_partitions` to override its default behavior of automatically determining a new meta, we also make sure to preserve divisions if they're known.

And `rep` is the new report `Array` collection, it doesn't matter what the meta or divisions are, this should be a pretty small object once computed. If/when we converge on a proper schema for a report array perhaps we can pass in its typetracer version for absolute correctness here

douglasdavis:
I just gave the schema a little better definition, here it is in JSONSchema:
```json
{
    "type": "object",
    "properties": {
        "args": {"type": "array", "items": {"type": "string"}},
        "kwargs": {
            "type": "array",
            "items": {"type": "array", "items": {"type": "string"}},
        },
        "exception": {"type": "string"},
        "message": {"type": "string"},
    },
}
```
We report back the function arguments, keyword arguments, the exception name, and the message associated with it.
- `args` is a list of strings
- `kwargs` is a list of list of strings, the inner list is always length two: `[key, value]`
- `exception` is a list of strings
- `message` is a list of strings

The report is always the same length as there are number of partitions. So if we have five partitions and two failures, the computed report will end up looking something like (making up some arguments and a single keyword argument):

```
[
    {'args': [], 'kwargs': [], 'exception': '', 'message': ''},   # <-- succeeded
    {'args': [], 'kwargs': [], 'exception': '', 'message': ''},   # <-- succeeded
    {'args': ["/path/to/file2", "True", "2"], 'kwargs': [["kwarg1", "foo"]], 'exception': 'RuntimeError', 'message': 'Failed to read.'},
    {'args': [], 'kwargs': [], 'exception': '', 'message': ''},   # <-- succeeded
    {'args': ["/path/to/file4", "True", "4"], 'kwargs': [["kwarg1", "bar"]], 'exception': 'RuntimeError', 'message': 'Failed to read.'},
]

**Comment by lgray** [comment]
Oh - this is *very* good. I'll try to test it asap.

**Comment by lgray** [comment]
Will this scale OK when there are 200k partitions?

**Comment by douglasdavis** [comment]
> Will this scale OK when there are 200k partitions?

I imagine it would. I think a 200k element array isn't too crazy, especially when most elements will be records without much data.

An alternative approach to shrink the report would be to actually have the elements of the "report array" that represent successful reads to just be None instead of the thin record. Something like:

```python
[
    None,
    None,
    {'args': ["/path/to/file2", "True", "2"], 'kwargs': [["kwarg1", "foo"]], 'exception': 'RuntimeError', 'message': 'Failed to read.'},
    None,
    {'args': ["/path/to/file4", "True", "4"], 'kwargs': [["kwarg1", "bar"]], 'exception': 'RuntimeError', 'message': 'Failed to read.'},
]
```

instead of what I wrote in my above comment.

> If, instead of the partition number, it would just report the start and stop rows it would be more immediately useful to users I think

Seems like we may want to add the ability to have consumers of this API be able to pass in a callback function that accepts the `*args` and `**kwargs` (with domain knowledge) and construct something to add to the report. As it stands `dask-awkward` can only blindly accept the args and kwargs.

douglasdavis:
Ah I may have misunderstood, if the divisions are known it would be possible for dask-awkward to combine the known divisions with the failed partitions.

lgray:
Yeah I like the idea of being able to control the form of what gets returned, with some decent presets.

You understood my question the first time correctly! If there are (somehow 200k) errors are we going to cause *further* errors or hassle for the user. :-D Seems like not so much, but I wanted to double check with you.

douglasdavis:
Great, I'm working on making what gets reported customizable, with what I think is a sane default.

lgray:
Oh I see what you mean.

Yeah I mean for the return:
```
[
    None,
    None,
    {'args': ["/path/to/file2", "True", "2"], 'kwargs': [["kwarg1", "foo"]], 'exception': 'RuntimeError', 'message': 'Failed to read.'},
    None,
    {'args': ["/path/to/file4", "True", "4"], 'kwargs': [["kwarg1", "bar"]], 'exception': 'RuntimeError', 'message': 'Failed to read.'},
]
```

something like:
```
[
    None,
    None,
    {'args': ["/path/to/file2", "True", "200-300"], 'kwargs': [["kwarg1", "foo"]], 'exception': 'RuntimeError', 'message': 'Failed to read.'},
    None,
    {'args': ["/path/to/file4", "True", "400-500"], 'kwargs': [["kwarg1", "bar"]], 'exception': 'RuntimeError', 'message': 'Failed to read.'},
]
```

Would be more immediately useful.

douglasdavis:
Ah in my little example that was just some made up set of arguments that were getting passed to the actual read function. So in the uproot case it would be the arguments passed to the callable class [_UprootOpenAndRead](https://github.com/scikit-hep/uproot5/blob/4112159408cc2cb7b1c359761a683b8ec1288c98/src/uproot/_dask.py#L1047), for example.

I've just pushed a commit that makes the report construction customizable but with the default that we've already been discussing.

To customize, the consumer of this API needs to define two functions: a success function and a failure function. The success function just takes `f(*args, **kwargs)`, while the failure function takes `f(exception, *args, **kwargs)`. Both must return an awkward array. The default versions for each are pretty easy to interpret.

`from_map` now has four new arguments -- I'm not loving it but I'm also having a hard time picturing an alternative

douglasdavis:
For reporting the rows, I think one would just need to combine the knowledge of 

```python
def find_bad_rows(report, divisions):
    bad_rows = []
    for i, element in enumerate(report):
        # if element shows bad
        if element.bad:
            bad_rows.append((divisions[i], divisions[i+1]))

report_collection, array_collection = uproot.dask(..., report=True)
report, array = dask.compute(report_collection, array_collection)
bad_rows = find_bad_rows(report, array_collection.divisions)
```

Or, if the *args and **kwargs have any knowledge of eventual rows, that can be embedded in the consumer defined success/failure functions.

lgray:
Just kind of a cosmetic thing but can we have the return be:
```python3
array, report = uproot.dask(...)
```

that would match typetracer with report, etc. as far as consistency goes.

douglasdavis:
Of course! this was just some shorthand -- I'm sure the PR to uproot will be be more thought out :)

douglasdavis:
As this PR stands our column optimization breaks because the `io_func` is getting wrapped with our `return_empty_on_raise` function. The optimization checks to see if something is projectable fails because of the wrapping. I'm trying to solve this now.

lgray:
awesome! A parquet-based one would be cool too but is not as high-priority.

src/dask_awkward/layers/layers.py
```diff
@@ -225,14 +237,20 @@ def prepare_for_projection(self) -> tuple[AwkwardInputLayer, TypeTracerReport, T
...
         """
         assert self.is_projectable
+        fn = maybe_unwrap(self.io_func)
         new_meta_array, report, state = cast(
-            ImplementsProjection, self.io_func
+            ImplementsProjection, fn
         ).prepare_for_projection()
 
+        if io_func_rer_wrapped(self.io_func):
+            new_return = (new_meta_array, type(new_meta_array)([]))
```
agoose77, line 246: Is this just creating a blank `ak.Array`?
```suggestion
            new_return = (new_meta_array, ak.from_iter([]))
```

src/dask_awkward/lib/io/io.py
```diff
@@ -504,29 +496,104 @@ def __call__(self, packed_arg):
...
-    backend: BackendT,
-) -> Callable:
-    @functools.wraps(fn)
-    def wrapped(*args, **kwargs):
+def default_report_success(*args: Any, **kwargs: Any) -> ak.Array:
+    return ak.Array(
+        [
+            {
+                "args": [],
+                "kwargs": [],
```
agoose77, line 504: We probably want to ensure that this matches the type of the failure. Given that we know this ahead of time, we can express the default report as a form and use `ak.Array(form.length_zero_array(highlevel=False))`

src/dask_awkward/lib/io/io.py
```diff
@@ -504,29 +496,104 @@ def __call__(self, packed_arg):
...
+    func: Callable,
+    *iterables: Iterable,
+    empty_on_raise: tuple[type[BaseException], ...],
+    empty_backend: BackendT,
+    args: tuple[Any, ...] | None = None,
+    label: str | None = None,
+    token: str | None = None,
+    divisions: tuple[int, ...] | tuple[None, ...] | None = None,
+    meta: ak.Array | None = None,
+    empty_success_callback: Callable[..., ak.Array] = default_report_success,
```
agoose77, line 592: How do you feel about `on_success` instead of `empty_success_callback`

src/dask_awkward/lib/io/io.py
agoose77: Can we not specify `meta` and let dask-awkward figure it out using the graph?

src/dask_awkward/layers/layers.py
douglasdavis: Yes it is, for historical reasons awkward wasn't imported in this module, but I think it would be fine now.

src/dask_awkward/layers/layers.py
```diff
@@ -107,20 +107,30 @@ def mock_empty(self, backend: BackendT = "cpu") -> AwkwardArray:
...
+
+def maybe_unwrap(func: Callable) -> Callable:
+    if io_func_rer_wrapped(func):
+        return func.__rer_wrapped__  # type: ignore
+    return func
+
+
 def io_func_implements_projection(func: ImplementsIOFunction) -> bool:
-    return hasattr(func, "prepare_for_projection")
+    return hasattr(maybe_unwrap(func), "prepare_for_projection")
```
agoose77, line 121: I'd be tempted to remove the unwrapping here, and require that the caller unwrap the IO function. That way it's a bit more explicit?

src/dask_awkward/lib/io/io.py
```diff
@@ -504,29 +496,104 @@ def __call__(self, packed_arg):
...
+                "args": [str(a) for a in args],
+                "kwargs": [[k, str(v)] for k, v in kwargs.items()],
+                "exception": str(type(exception).__name__),
+                "message": str(exception),
+            },
+        ],
+    )
+
+
+class return_empty_on_raise:
```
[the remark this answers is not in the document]
douglasdavis: Yes sounds like a good idea!

src/dask_awkward/layers/layers.py
```diff
@@ -107,20 +107,30 @@ def mock_empty(self, backend: BackendT = "cpu") -> AwkwardArray:
         )
 
 
+def io_func_rer_wrapped(func: ImplementsIOFunction) -> bool:
+    return hasattr(func, "__rer_wrapped__")
```
agoose77, line 111: It looks like this is inspired by `__wrapped__`. For our naming, perhaps `_raise_on_error_wrapped` might be better?

agoose77:
@douglasdavis looks good! I made some comments.

In general, do you think it's possible / sensible to move the `raise_on_error` machinery to additional protocols that `io_func` implements? I.e. rather than wrapping the IO function, we explicitly ask it if it supports these features? My hunch is that will be slightly simpler, and we could still provide a convenience adaptor that does this for plain-callable IO functions.

Hopefully this review doesn't feel like a lot of nitpicks and lastminute.com suggestions.

douglasdavis:
@agoose77 thanks a lot for the comments. 

> In general, do you think it's possible / sensible to move the raise_on_error machinery to additional protocols that io_func implements? I.e. rather than wrapping the IO function, we explicitly ask it if it supports these features? My hunch is that will be slightly simpler, and we could still provide a convenience adaptor that does this for plain-callable IO functions.

I have thought a bit about this, especially with the `from_map` arguments starting to blow up a bit, and with the specific changes which had to be made to handle the new complexities of wrapped io_funcs. I'll keep ruminating on it a bit.

> Hopefully this review doesn't feel like a lot of nitpicks and lastminute.com suggestions. I'm glad you've thought about this, because it's important to get right, and I'm just adding some without-much-context thoughts.

Not at all!

src/dask_awkward/lib/io/io.py
```diff
@@ -496,29 +499,147 @@ def __call__(self, packed_arg):
...
+def on_failure_default(
+    exception: type[BaseException],
+    *args: Any,
+    **kwargs: Any,
+) -> ak.Array:
+    return ak.Array(
+        [
+            {
+                "args": [str(a) for a in args],
+                "kwargs": [[k, str(v)] for k, v in kwargs.items()],
```
agoose77, line 551: Perhaps this should be `repr`?

douglasdavis: Yea that makes more sense 👍

PR #418, opened 2023-11-15
feat!: harden dtype handling for scalars

douglasdavis:
This PR hardens treatment of meta/dtypes for Scalar collections. It changes the `Scalar.__init__` and `new_scalar_object` function, which should really only find use inside of dask-awkward, but it's still possible this can break things. 

The PR makes sure that all Scalars have a meta and a dtype, stemming from conversation with @agoose77 on Slack, so I'd be happy your feedback here :)


src/dask_awkward/lib/core.py
```diff
@@ -85,14 +85,22 @@ def __init__(
...
         known_value: Any | None = None,
     ) -> None:
         if not isinstance(dsk, HighLevelGraph):
             dsk = HighLevelGraph.from_collections(name, dsk, dependencies=())  # type: ignore
         self._dask: HighLevelGraph = dsk
         self._name: str = name
-        self._meta: Any = self._check_meta(meta)
+        if meta is not None and dtype is None:
+            self._meta = self._check_meta(meta)
+            self._dtype = np.dtype(self._meta.type.content.primitive)
```
agoose77, line 98: If the meta always has a `NumpyType` type, we could also just use `self._meta.layout.dtype`. I think that might be slightly better, because it uses Awkward's dtype→primitive conversion.

agoose77, line 98:
```suggestion
            self._dtype = self._meta.layout.dtype
```

agoose77:
OK, looks good (though I made a change)! It seems then that scalars are now always implemented as an `ak.Array` of length 1. I assume that this is so that `_meta` from dask-awkward collections is always `None` or an `ak.Array` for simplicity?

douglasdavis:
>  _meta from dask-awkward collections is always None or an ak.Array for simplicity?

Yes that's what I'm going for here!

PR #421, opened 2023-11-29
feat!: change default for `raise_failed_meta`

agoose77:
@alexander-held reported seeing an obscure warning the other day. I think at this stage we're mature enough to want to see error reports from typetracer failures. This PR changes the default and tries to hide the obscure any mention of typetracer.

We should eventually remove the `length_zero_array` fallback.

### DIFF
--- a/src/dask_awkward/awkward.yaml
+++ b/src/dask_awkward/awkward.yaml
@@ -2,7 +2,7 @@ awkward:
 
   # This option causes graph creation to fail if the metadata of a
   # dask-awkward Array collection cannot be determined.
-  raise-failed-meta: False
+  raise-failed-meta: True
 
   # This option is for cases where new array collections are created
   # with unknown metadata. The default setting is to compute the first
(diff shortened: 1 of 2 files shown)


agoose77:
@lgray for your awareness!

lgray:
I was going to add an override in coffea for this. I approve of this change, it is much more clear what to do if it fails IMHO.

lgray:
I would suggest we also raise on column optimization failures by default? If the column optimization fails (which is usually a genuine error anyway) there tends to be an enormous performance penalty. I think the iteration cycle for coding would be faster if it just fails rather than continuing onwards reading the whole file.

douglasdavis:
I think this is a good change. Looks like the failure is related to typetracer and a datetime dtype

agoose77:
Looks like the dtype problem arises from w

agoose77:
@lgray yes, a separate PR is a good idea. I think we probably should look to that now that column optimisation is in a better place.

PR #425, opened 2023-12-01
fix: support logical concatenation at `axis=0`

### LINKED ISSUE #305: `ak.concatenate(..., axis=0)` should not touch any data
We can (and technically do) implement this as a growth in the number of partitions.

agoose77:
> [!WARNING]
> Depends upon scikit-hep/awkward#2862, scikit-hep/awkward#2860
 
Fixes #305.

### DIFF
--- a/src/dask_awkward/lib/operations.py
+++ b/src/dask_awkward/lib/operations.py
@@ -1,9 +1,13 @@
 from __future__ import annotations
 
 from collections.abc import Mapping
-from typing import Any
+from typing import TYPE_CHECKING, Any
 
 import awkward as ak
+from awkward.operations.ak_concatenate import (
+    enforce_concatenated_form as enforce_layout_to_concatenated_form,
+)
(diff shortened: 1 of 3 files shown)


src/dask_awkward/lib/operations.py
```diff
@@ -42,28 +59,75 @@ def concatenate(
...
+                previous_layer_names=[a.name for a in arrays],
+                fn=_concatenate_axis_0_meta,
+            )
+        else:
+            g = {
+                (name, i): k
+                for i, k in enumerate(
+                    k for collection in arrays for k in collection.__dask_keys__()
+                )
+            }
```
douglasdavis, line 120: I did not know this level of list/dict comprehension existed 🤯

douglasdavis (approved):
I'm still absorbing the new "enforcement" stuff in upstream awkward, but the way the graphs are getting constructed makes sense to me so I think this is good to go

PR #430, opened 2023-12-02
fix: pass `keys` to column optimisation

### LINKED ISSUE #426: Column optimization breaks in case of computing function of records in addition to exploded records?
```python3
import dask
from dask_awkward.lib.testutils import assert_eq

from coffea.nanoevents import NanoEventsFactory

with dask.config.set({"awkward.optimization.enabled": True}):
    eagerevents = NanoEventsFactory.from_root(
        {"tests/samples/nano_dy.root": "Events"},
        delayed=False,
    ).events()

    daskevents = NanoEventsFactory.from_root(
        {"tests/samples/nano_dy.root": "Events"},
        delayed=True,
    ).events()

    mval_eager, (a_eager, b_eager) = eagerevents.Muon.metric_table(
        eagerevents.Jet, return_combinations=True
    )
    mval_dask, (a_dask, b_dask) = daskevents.Muon.metric_table(
        daskevents.Jet, return_combinations=True
    )

    mval_dask, (a_dask, b_dask) = dask.compute(mval_dask, (a_dask, b_dask))

    assert_eq(mval_eager, mval_dask)
    assert_eq(a_eager, a_dask)
    assert_eq(b_eager[field], b_dask[field])
```

results in:
```
/Users/lgray/coffea-dev/coffea/src/coffea/nanoevents/schemas/nanoaod.py:243: RuntimeWarning: Missing cross-reference index for FatJet_genJetAK8Idx => GenJetAK8
  warnings.warn(
/Users/lgray/coffea-dev/coffea/src/coffea/nanoevents/schemas/nanoaod.py:243: RuntimeWarning: Missing cross-reference index for FatJet_genJetAK8Idx => GenJetAK8
  warnings.warn(
  File "iason.py", line 26, in <module>
    assert_eq(a_eager, a_dask)
  File "/Users/lgray/coffea-dev/dask-awkward/src/dask_awkward/lib/testutils.py", line 37, in assert_eq
    assert_eq_arrays(
  File "/Users/lgray/coffea-dev/dask-awkward/src/dask_awkward/lib/testutils.py", line 67, in assert_eq_arrays
    b_tt = typetracer_array(b)
  File "/Users/lgray/coffea-dev/dask-awkward/src/dask_awkward/lib/core.py", line 2390, in typetracer_array
    a.layout.to_typetracer(forget_length=True),
  File "/Users/lgray/miniforge3/envs/coffea-dev/lib/python3.8/site-packages/awkward/contents/content.py", line 275, in to_typetracer
    return self._to_typetracer(forget_length)
  File "/Users/lgray/miniforge3/envs/coffea-dev/lib/python3.8/site-packages/awkward/contents/listoffsetarray.py", line 230, in _to_typetracer
    self._content._to_typetracer(forget_length),
  File "/Users/lgray/miniforge3/envs/coffea-dev/lib/python3.8/site-packages/awkward/contents/listoffsetarray.py", line 230, in _to_typetracer
    self._content._to_typetracer(forget_length),
  File "/Users/lgray/miniforge3/envs/coffea-dev/lib/python3.8/site-packages/awkward/contents/indexedarray.py", line 243, in _to_typetracer
    self._content._to_typetracer(forget_length),
  File "/Users/lgray/miniforge3/envs/coffea-dev/lib/python3.8/site-packages/awkward/contents/recordarray.py", line 370, in _to_typetracer
    contents = [x._to_typetracer(forget_length) for x in self._contents]
  File "/Users/lgray/miniforge3/envs/coffea-dev/lib/python3.8/site-packages/awkward/contents/recordarray.py", line 370, in <listcomp>
    contents = [x._to_typetracer(forget_length) for x in self._contents]
  File "/Users/lgray/miniforge3/envs/coffea-dev/lib/python3.8/site-packages/awkward/contents/numpyarray.py", line 225, in _to_typetracer
    data = self._raw(backend.nplike)
  File "/Users/lgray/miniforge3/envs/coffea-dev/lib/python3.8/site-packages/awkward/contents/numpyarray.py", line 199, in _raw
    return to_nplike(self.data, nplike, from_nplike=self._backend.nplike)
  File "/Users/lgray/miniforge3/envs/coffea-dev/lib/python3.8/site-packages/awkward/_nplikes/__init__.py", line 41, in to_nplike
    return nplike.asarray(array)
  File "/Users/lgray/miniforge3/envs/coffea-dev/lib/python3.8/site-packages/awkward/_nplikes/typetracer.py", line 658, in asarray
    assert not isinstance(obj, PlaceholderArray)
AssertionError
```

And we shouldn't be running into `PlaceholderArray`s after column optimization!
If you flip column optimization to False there are no more placeholder arrays, but an assertion is throw because of an error I need to track down in coffea (but there's no exception from within awkward array).

You'll need to use `coffea@master` to have this produce the appropriate error.

agoose77:
@douglasdavis I think this fixes #426.

However, I changed the type hints and made a subtle change to appease the type checker. I'm not convinced it's correct and would appreciate your input. I'll tag you at the edit site.

In our tests we assume that we can pass in `keys=[]`. Is that supposed to be the way that `dask.optimize` is used?

### DIFF
--- a/src/dask_awkward/lib/inspect.py
+++ b/src/dask_awkward/lib/inspect.py
@@ -4,7 +4,6 @@ from typing import TYPE_CHECKING, Any, NamedTuple
 
 import numpy as np
 from dask.base import unpack_collections
-from dask.highlevelgraph import HighLevelGraph
 
 from dask_awkward.layers import AwkwardInputLayer
 
@@ -81,8 +80,9 @@ def report_necessary_buffers(
 
     name_to_necessary_buffers: dict[str, NecessaryBuffers | None] = {}
     for obj in collections:
-        dsk = obj if isinstance(obj, HighLevelGraph) else obj.dask
-        projection_data = o._prepare_buffer_projection(dsk)
+        dsk = obj.__dask_graph__()
+        keys = obj.__dask_keys__()
+        projection_data = o._prepare_buffer_projection(dsk, keys)
 
         # If the projection failed, or there are no input layers
         if projection_data is None:
@@ -178,8 +178,9 @@ def report_necessary_columns(
 
     name_to_necessary_columns: dict[str, frozenset | None] = {}
     for obj in collections:
-        dsk = obj if isinstance(obj, HighLevelGraph) else obj.dask
-        projection_data = o._prepare_buffer_projection(dsk)
+        dsk = obj.__dask_graph__()
+        keys = obj.__dask_keys__()
+        projection_data = o._prepare_buffer_projection(dsk, keys)
 
         # If the projection failed, or there are no input layers
         if projection_data is None:
--- a/src/dask_awkward/lib/optimize.py
+++ b/src/dask_awkward/lib/optimize.py
@@ -287,48 +287,48 @@ def rewrite_layer_chains(dsk: HighLevelGraph, keys: Sequence[Key]) -> HighLevelG
     dependents = dsk.dependents
     all_layers = set(dsk.layers)
     while all_layers:
-        lay = all_layers.pop()
-        val = dsk.layers[lay]
-        if not isinstance(val, AwkwardBlockwiseLayer):
+        layer_key = all_layers.pop()
+        layer = dsk.layers[layer_key]
+        if not isinstance(layer, AwkwardBlockwiseLayer):
             # shortcut to avoid making comparisons
-            layers[lay] = val  # passthrough unchanged
+            layers[layer_key] = layer  # passthrough unchanged
             continue
-        children = dependents[lay]
-        chain = [lay]
-        lay0 = lay
+        children = dependents[layer_key]
+        chain = [layer_key]
+        current_layer_key = layer_key
         while (
             len(children) == 1
-            and dsk.dependencies[list(children)[0]] == {lay}
+            and dsk.dependencies[list(children)[0]] == {current_layer_key}
             and isinstance(dsk.layers[list(children)[0]], AwkwardBlockwiseLayer)
-            and len(dsk.layers[lay]) == len(dsk.layers[list(children)[0]])
-            and lay not in required_layers
+            and len(dsk.layers[current_layer_key]) == len(dsk.layers[list(children)[0]])
+            and current_layer_key not in required_layers
         ):
             # walk forwards
-            lay = list(children)[0]
-            chain.append(lay)
-            all_layers.remove(lay)
-            children = dependents[lay]
-        lay = lay0
-        parents = dsk.dependencies[lay]
+            current_layer_key = list(children)[0]
+            chain.append(current_layer_key)
+            all_layers.remove(current_layer_key)
+            children = dependents[current_layer_key]
+
+        parents = dsk.dependencies[layer_key]
         while (
             len(parents) == 1
-            and dependents[list(parents)[0]] == {lay}
+            and dependents[list(parents)[0]] == {layer_key}
             and isinstance(dsk.layers[list(parents)[0]], AwkwardBlockwiseLayer)
-            and len(dsk.layers[lay]) == len(dsk.layers[list(parents)[0]])
+            and len(dsk.layers[layer_key]) == len(dsk.layers[list(parents)[0]])
             and list(parents)[0] not in required_layers
         ):
             # walk backwards
-            lay = list(parents)[0]
-            chain.insert(0, lay)
-            all_layers.remove(lay)
-            parents = dsk.dependencies[lay]
+            layer_key = list(parents)[0]
+            chain.insert(0, layer_key)
+            all_layers.remove(layer_key)
+            parents = dsk.dependencies[layer_key]
         if len(chain) > 1:
             chains.append(chain)
             layers[chain[-1]] = copy.copy(
                 dsk.layers[chain[-1]]
             )  # shallow copy to be mutated
         else:
-            layers[lay] = val  # passthrough unchanged
+            layers[layer_key] = layer  # passthrough unchanged
 
     # do rewrite
     for chain in chains:
--- a/src/dask_awkward/lib/optimize.py
+++ b/src/dask_awkward/lib/optimize.py
@@ -3,7 +3,7 @@ from __future__ import annotations
 import copy
 import logging
 import warnings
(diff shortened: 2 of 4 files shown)


src/dask_awkward/lib/inspect.py
```diff
@@ -81,8 +80,9 @@ def report_necessary_buffers(
 
     name_to_necessary_buffers: dict[str, NecessaryBuffers | None] = {}
     for obj in collections:
-        dsk = obj if isinstance(obj, HighLevelGraph) else obj.dask
-        projection_data = o._prepare_buffer_projection(dsk)
+        dsk = obj.__dask_graph__()
+        keys = obj.__dask_keys__()
+        projection_data = o._prepare_buffer_projection(dsk, keys)
```
agoose77, line 86: @douglasdavis first uncertainty - I'm pretty sure that anything in `collections` has to be something that has a `__dask_graph__` attribute, so I'm assuming here that we can also ask for `__dask_keys__` too.

src/dask_awkward/lib/optimize.py
```diff
@@ -77,15 +78,15 @@ def optimize(
...
+            dsk = optimize_columns(dsk, keys)
         if "layer-chains" in which:
             dsk = rewrite_layer_chains(dsk, keys)
 
     return dsk
 
 
 def _prepare_buffer_projection(
-    dsk: HighLevelGraph,
+    dsk: HighLevelGraph, keys: Sequence[Key]
```
agoose77, line 89: I *believe* the keys from `__dask_optimize__` are always sequence of `Key`? So, I think we can impose that constraint on our own APIs. LMK if not and you'd prefer to support single keys too. Need to figure out how to do that, because the `Key` type is so general.

src/dask_awkward/lib/optimize.py
```diff
@@ -292,7 +281,7 @@ def rewrite_layer_chains(dsk: HighLevelGraph, keys: Any) -> HighLevelGraph:
     chains = []
     deps = copy.copy(dsk.dependencies)
 
-    required_layers = {k[0] for k in keys}
+    required_layers = {k[0] for k in keys if isinstance(k, tuple)}
```
agoose77, line 284: @douglasdavis this is a big change. I am not sure, given Dask's type hints, whether we can safely assume that a key is always a tuple, can we? But it seems like this code does. I have too little domain specific knowledge here to know, so I encoded the assertion in an isinstance (rather than an assert) i.e. this is not final or correct.

martindurant: Dask keys are NOT in general tuples - strings are common. However, for anything blockwise based (including all of dask-awkward), the pattern is (name: str, index: int). @douglasdavis can comment on whether this is also true for scalars.

In dask.core:
```
def iskey(key: object) -> bool:
    """Return True if the given object is a potential dask key; False otherwise."""
    typ = type(key)
    if typ is tuple:
        return all(iskey(i) for i in cast(tuple, key))
    return typ in {bytes, int, float, str}
```

agoose77: Hmm, OK! I wonder at what point can / should we encode this in our typing and function signatures?

src/dask_awkward/lib/inspect.py
douglasdavis: Yes `__dask_keys__` is part of the [collection interface](https://docs.dask.org/en/stable/_modules/dask/typing.html#DaskCollection.__dask_keys__)

src/dask_awkward/lib/optimize.py
douglasdavis: In this part of the code it does make sense for keys to always be a tuple in dask-awkward world, but when using dask-awkward and dask-histogram together, and since dask-histogram is dask-awkward-optimization-aware, it should be possible to run into a case where we have a string key (dask-histogram collections use a simple string as the collection key). Do you think this check is necessary?

agoose77: @douglasdavis if this is @ me, then I don't think _this_ check is right, but we probably need some kind of switch to determine whether `k[0]` will throw (i.e. if the key is not a sequence). I also imagine that `k[0]` is not intended to pull the first character of the key if it's a string. So, perhaps better
```suggestion
    required_layers = {k[0] if (isinstance(k, tuple) and ...) else k for k in keys if isinstance(k, tuple)}
```
I'm spit-balling here, though.

douglasdavis: Ah! but this is here just for the purpose of awkward blockwise layers, so we do always expect to see `(name, i)`. So the isinstance tuple check here makes sense and perhaps we should rename the variable to be `required_blockwise_layers`

src/dask_awkward/lib/optimize.py
douglasdavis: I'm not sure if this is an assumption we can make. See dask.dataframe's [https://github.com/dask/dask/blob/bb9aa197b851726caa47250a3a12598a34249008/dask/dataframe/optimize.py#L15-L18](implementation). We can always coerce keys into a sequence -- which is what dataframe does

src/dask_awkward/lib/optimize.py
```diff
@@ -117,26 +118,14 @@ def _prepare_buffer_projection(
...
-    # a dependent, it must be the end of a graph. These are the things
-    # we need to compute for; we only use a single partition (the
-    # first). for a single collection `.compute()` this list will just
-    # be length 1; but if we are using `dask.compute` to pass in
-    # multiple collections to be computed simultaneously, this list
-    # will increase in length.
-    leaf_layers_keys = [
-        (k, 0) for k, v in dsk.dependents.items() if isinstance(v, set) and len(v) == 0
-    ]
-
```
douglasdavis, line 131: If we drop this line we lose our only-optimize-on-the-first-partition strategy which makes sure the typetracer-only graph only executes on the minimal necessary graph.

agoose77: Ah, yes! I was working on this quite late, and this seems obvious in the light of day. I wonder whether we can determine which of the passed-in keys are collections that we can slice like this?

douglasdavis:
Alright this is getting close. It fixes #426 which was the goal. Just need to get a test in that doesn't depend on coffea.

PR #437, opened 2023-12-13
fix: pickle array attributes

### LINKED ISSUE #427: Serialise attrs in pickling an Array

agoose77:
This PR should have been done earlier (it fell off my To-Do list!). It introduces a versioned unpickling scheme (through dedicated unpickling functions) that ensures we also serialise the `attrs` and `behavior` dictionaries. As per the semantics of `attrs`, we drop any prefixed keys.

@lgray it occurs to me that inter-worker transfer would break arrays which depend upon `@foo` keys existing in the attrs dict. I can't imagine that we would want that. Maybe we need to discuss the issue further, though this PR can merge as-is.

Closes #427

### DIFF
--- a/src/dask_awkward/pickle.py
+++ b/src/dask_awkward/pickle.py
@@ -1,5 +1,6 @@
 from __future__ import annotations
 
+from collections.abc import Mapping
 from typing import Any
 
 __all__ = ("plugin",)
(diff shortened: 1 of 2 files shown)


lgray:
For `@foo` it's setup by construction, and in fact exploited, in coffea that those attrs disappear on serialization. 

For more general use I see the issue but I tend to consider `@foo` client-side only attributes.

agoose77:
@lgray I think we've hithertoo thought about these attributes disappearing during serialisation of the graph, but not of the serialisation that occurs when a result is moved between workers during graph execution. As I understand it, should Dask decide to do this, the high-level behaviour would break because these attributes would no longer exist.

agoose77:
My view is also that we should always drop attributes at serialisation time. My question is whether a chained operation that doesn't get fused and runs on multiple nodes would ever need the things you're storing in `attrs` under `@`-prefixed keys. If it's _only_ needed at graph building time, then we're fine!

Otherwise something more magical, e.g. `behavior`-overloads for pickling, would be needed.

lgray:
In my case what you're mentioning never happens. Everything I need `@foo` for is during task-graph creation as opposed to execution. Though, for sure I can imagine people doing much more clever things where what you're saying may happen... But then they should just use a regular attr, no?

PR #444, opened 2024-01-08
docs: add IO notebook

douglasdavis:
Add a 3 part jupyter notebook diving into some IO details.


docs/examples/io-tutorial/notebook.ipynb
martindurant, line 1: I find that maybe this could be split into two:
- the basic to/from commands and how to use them, with a little detail no what the options mean and that the user expects optimisation to happen (can we make a dataset big enough to show a timing or memory difference?)
- a detailed "intermediate" one, the only one that mentions layers and protocols

btw.: should we have a from_avro? https://awkward-array.org/doc/main/reference/generated/ak.from_avro_file.html (does not allow schema input/selection; does not work on chunks within avro files)

PR #448, opened 2024-01-11
fix: return proxy of attrs, behavior; dak.from_awkward correctly sets attrs

### LINKED ISSUE #447: bug: attrs do not survive compute()
```python3
import awkward
import dask_awkward

x = dask_awkward.from_awkward(awkward.Array([0]), 1)
x.attrs["test"] = "test"

print(x.attrs)

print(x.compute().attrs)
```

yields:
```
{'test': 'test'}
{}
```

agoose77:
Closes #447 by making `.attrs` and `.behavior` read-only. This addresses a confusing part of the API: currently these are writeable, but the underlying compute does not share the results of mutating attrs or behavior.

If we need to, we can add a `dask_awkward.with_attrs` to add a mechanism of mutating these objects. I'm not sure what the ramifications of this PR will be — we may internally be expecting to pass around the true attrs dicts, which would be broken by this change.


lgray:
So I can only add attrs to an awkward array and then make a dask-awkward array from that?

This probably breaks nanoevents?

lgray:
yep:
```
self = <coffea.nanoevents.factory.NanoEventsFactory object at 0x12a0f1330>

    def events(self):
        """Build events"""
        if self._is_dask:
            events = self._mapping(form_mapping=self._schema)
            report = None
            if isinstance(events, tuple):
                events, report = events
>           events.attrs["@original_array"] = events
E           TypeError: 'mappingproxy' object does not support item assignment
```

agoose77:
You can still override attrs of a dask_awkward array, but it's harder to do. You just copy the `meta` and patch the `.attrs`. The intention of this PR is that the user-facing `.attrs` should not be writeable.

lgray:
Ah, indeed I can see a few work arounds. I'll confirm one in an hour or so.

lgray:
Yep - all good - simple work around found.

douglasdavis:
based on conversation this morning it seemed that this is good to go?

martindurant:
(test coverage would be nice)

lgray:
@douglasdavis yes this is fine from my end

lgray:
@martindurant let me know if this is sufficient for tests, otherwise I think this one is good to go. Thanks in advance for the review.

tests/test_core.py
```diff
@@ -891,6 +891,24 @@ def test_shape_only_ops(fn: Callable, tmp_path_factory: pytest.TempPathFactory)
         result.compute()
 
 
+def test_assign_behavior() -> None:
+    x = ak.Array([{"a": 1, "b": 2}, {"a": 3, "b": 4}], behavior={}, attrs={})
+    dx = dak.from_awkward(x, 3)
```
martindurant, line 896: Should test that `dx`'s behavior and attrs are the ones given (would have been `None`)

lgray: See other changes. Unless we don't want to take the array's original attrs if none given.

lgray: OK this exposed a few places where attrs weren't being propagated correctly (in particular the whole from-json interface). That is fixed now.

lgray:
@martindurant fixing a few things has led to scope creep of this PR... :shrug: Anyway it's ready for another look.

martindurant (approved):

PR #449, opened 2024-01-12
feat: add support for `Delayed` in partitionwise layer creation.

douglasdavis:
Broadcasting will occur equivalently at each partition because the Delayed object is *not* partitioned (as shown in the test added in this PR).

### DIFF
--- a/src/dask_awkward/lib/core.py
+++ b/src/dask_awkward/lib/core.py
@@ -2029,13 +2029,22 @@ def map_partitions(
         dependencies=flat_deps,
     )
 
+    dak_arrays = tuple(filter(lambda x: isinstance(x, Array), flat_deps))
+    if len(dak_arrays) == 0:
+        raise TypeError(
+            "at least one argument passed to map_partitions "
+            "should be a dask_awkward.Array collection."
+        )
+    in_npartitions = dak_arrays[0].npartitions
+    in_divisions = dak_arrays[0].divisions
+    if partition_compatibility(*dak_arrays) == PartitionCompatibility.NO:
+        raise IncompatiblePartitions("map_partitions", *dak_arrays)
+
     if output_divisions is not None:
         if output_divisions == 1:
             new_divisions = flat_deps[0].divisions
         else:
-            new_divisions = tuple(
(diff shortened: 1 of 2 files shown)


lgray:
Cool, I'll give it a try in the morning!

lgray:
I'm getting an odd error when I try to print out the dask awkward array that's made from something using delayed:
```
  File "/Users/lgray/coffea-dev/coffea/tests/test_lookup_tools.py", line 177, in test_correctionlib
    print(test_out_dak)
  File "/Users/lgray/coffea-dev/dask-awkward/src/dask_awkward/lib/core.py", line 998, in __str__
    f"npartitions={self.npartitions}"
  File "/Users/lgray/coffea-dev/dask-awkward/src/dask_awkward/lib/core.py", line 1083, in npartitions
    return len(self.divisions) - 1
  File "/Users/lgray/miniforge3/envs/ewkcoffea-debug/lib/python3.10/site-packages/dask/delayed.py", line 635, in __len__
    raise TypeError("Delayed objects of unspecified length have no len()")
TypeError: Delayed objects of unspecified length have no len()
```

I haven't traced it to understand why the output dak array things it's a dask.delayed?

agoose77:
It's probably that something's constructed `Array` with the `divisions` argument taken from `delayed.divisions`, i.e. in a previous frame.

douglasdavis:
Is it possible to get a repro for that `TypeError`?

`divisions` should always be a tuple (specifically `tuple[int, ...] | tuple[None, ...]`) on a `dak.Array`

lgray:
I'll see if I can get one - I'm trying to make sure first I'm not bamboozling myself.

lgray:
somehow array.divisions has, itself, become a delayed object!?

douglasdavis:
Just to jot this down information down in a more permanent place: we discovered during slack conversation that if a `Delayed` object was the first thing passed to `map_partitions`, we broke an assumption in the code that the first argument would be a `dak.Array`. The `map_partitions` implementation is not invariant w.r.t. order of function arguments, but at least one `dak.Array` is required.

martindurant:
> we broke an assumption in the code that the first argument would be a dak.Array

easy to fix, right?

PR #455, opened 2024-01-24
fix: remove over aggressive multiargument `getitem` to `map_partitions`

### LINKED ISSUE #454: Calling `len` on a `dak.Array` can report the wrong length
Eager arrays:

```python
>>> import uproot, skhep_testdata
>>> a = uproot.open({skhep_testdata.data_path("uproot-HZZ.root"): "events"}).arrays()
>>> len(a)
2421
>>> b = a["Muon_Px", 0]
>>> len(b)
2
```

Lazy arrays:

```python
>>> a = uproot.dask({skhep_testdata.data_path("uproot-HZZ.root"): "events"})
>>> len(a)
2421
>>> b = a["Muon_Px", 0]
>>> len(b)
2421
```

With a slice, one can easily get into the situation in which the length is not known without reading data. `len` can't return an `unknown_length`, but it could be made to raise an exception. Even the length of the source array (`a` in the above) is not always known: setting the `files` to more than one file and `open_files=False` in [uproot.dask](https://uproot.readthedocs.io/en/latest/uproot._dask.dask.html) will create a source array whose length is not known.

Should `dak.Array.__len__` keep track of when it truly knows the length and when it doesn't? A second-best option would be for it to always raise a

```
ValueError: length of dak.Array is not known before computation
```

douglasdavis:
Fix #454

### DIFF
--- a/src/dask_awkward/lib/core.py
+++ b/src/dask_awkward/lib/core.py
@@ -47,6 +47,7 @@ from dask_awkward.lib.optimize import all_optimizations
 from dask_awkward.utils import (
     DaskAwkwardNotImplemented,
     IncompatiblePartitions,
+    field_access_to_front,
     first,
     hyphenize,
     is_empty_slice,
@@ -1396,17 +1397,15 @@ class Array(DaskMethodsMixin, NDArrayOperatorsMixin):
         )
 
     def _getitem_tuple(self, where):
+        where = field_access_to_front(where)
         if isinstance(where[0], int):
             return self._getitem_outer_int(where)
 
-        elif isinstance(where[0], str):
+        elif isinstance(where[0], (str, list)):
             first, rest = where[0], where[1:]
             if rest:
                 return self[first][rest]
             return self[first]
-
-        elif isinstance(where[0], list):
-            return self._getitem_outer_str_or_list(where)
 
         elif isinstance(where[0], slice) and is_empty_slice(where[0]):
             return self._getitem_trivial_map_partitions(where)
--- a/src/dask_awkward/lib/core.py
+++ b/src/dask_awkward/lib/core.py
@@ -1400,7 +1400,10 @@ class Array(DaskMethodsMixin, NDArrayOperatorsMixin):
             return self._getitem_outer_int(where)
 
         elif isinstance(where[0], str):
-            return self._getitem_outer_str_or_list(where)
+            first, rest = where[0], where[1:]
+            if rest:
+                return self[first][rest]
+            return self[first]
 
         elif isinstance(where[0], list):
             return self._getitem_outer_str_or_list(where)
--- a/src/dask_awkward/lib/core.py
+++ b/src/dask_awkward/lib/core.py
@@ -975,7 +975,7 @@ class Array(DaskMethodsMixin, NDArrayOperatorsMixin):
 
     def __len__(self) -> int:
         if not self.known_divisions:
-            raise NotImplementedError(
+            raise TypeError(
                 "Cannot determine length of collection with unknown partition sizes without executing the graph.\n"
                 "Use `dask_awkward.num(..., axis=0)` if you want a lazy Scalar of the length.\n"
                 "If you want to eagerly compute the partition sizes to have the ability to call `len` on the collection"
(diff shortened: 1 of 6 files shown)


src/dask_awkward/lib/core.py
```diff
@@ -1400,7 +1400,10 @@ def _getitem_tuple(self, where):
             return self._getitem_outer_int(where)
 
         elif isinstance(where[0], str):
-            return self._getitem_outer_str_or_list(where)
+            first, rest = where[0], where[1:]
```
martindurant, line 1403: @jpivarski at some point talked about the possibility of reordering these elements to shuffle "fields" before "indexes", which would be more general than this, which only account for "more than one". I think in this context, "fields" is always str and "indexes" can be slice, int, ints or bools. Of course, this is also complicated by ellipsis, but I think we already accounted for that before calling here.

src/dask_awkward/utils.py
```diff
@@ -149,3 +149,23 @@ def second(seq: Iterable[T]) -> T:
...
+
+def field_access_like(entry: Any) -> bool:
+    if isinstance(entry, str):
+        return True
+    if isinstance(entry, (list, tuple)) and all(isinstance(x, str) for x in entry):
+        return True
+    return False
+
+
+def field_access_to_front(seq: Sequence[Any]) -> tuple[Any, ...]:
```
douglasdavis, line 163: @martindurant this is the function we talked a bit about yesterday.

src/dask_awkward/lib/core.py
```diff
@@ -1396,14 +1397,15 @@ def _getitem_slice_on_zero(self, where):
         )
 
     def _getitem_tuple(self, where):
+        where = field_access_to_front(where)
```
douglasdavis, line 1400: @martindurant (2/2) and here we call it for tuple getitems

douglasdavis:
```
FAILED tests/test_getitem.py::test_layered_string
```
failing because it's each element of the tuple is a list[str] which is doing field selection without diving one layer into the column structure. so the `first, rest = where[0], where[1:]` logic breaks, so another case that needs to be accounted for

src/dask_awkward/utils.py
martindurant: If you invert it:
```python
def field_access_like(entry: object) -> bool:
    if isinstance(entry, str):
        return False
    if isinstance(entry, (list, tuple)) and all(isinstance(x, str) for x in entry):
        return False
    return True
```

Then it's simply:
```
sorted(seq, key=field_access_like)
```
because it will put "field-likes" first but otherwise not change ordering.

martindurant: (well, this would be a list version)

douglasdavis:
@lgray this is the PR that would be worth taking for a spin to make sure nothing bad happens w.r.t. your worries from this morning's meeting.

src/dask_awkward/utils.py
```diff
@@ -149,3 +149,60 @@ def second(seq: Iterable[T]) -> T:
...
+    """
+    new_seq: list[Any] = []
+    n_front = 0
+    for entry in seq:
+        if field_access_like(entry):
+            new_seq.insert(n_front, entry)
+            n_front += 1
+        else:
+            new_seq.append(entry)
+    return tuple(new_seq), n_front
```
martindurant, line 208: Can't we use sorted and/or itertools.groupby?
```suggestion
    reversed(seq, key=lambda x: not field_access_like(x))
```

douglasdavis: We could but the `n_front` variable is useful in this case for knowing _where_ we stop having field-access like elements in the list.

lgray:
Standard checks look OK for data touching, please go ahead!

src/dask_awkward/utils.py
martindurant: That you could get from groupby (or map, filter, any...)
```suggestion
    return sorted(seq, key=field_access_like), sum(map(field_access_like)) # assuming field_access_like is inverted
```

douglasdavis:
On my end I think this is in a good place, but I'll let @martindurant wrap this one up.

PR #463, opened 2024-01-30
chore: let website-based release trigger workflow

lgray:


martindurant:
Does this work given `if: startsWith(github.ref, 'refs/tags')` on the upload?

lgray:
Same workflow operates fine over in dask-histogram.

PR #465, opened 2024-02-01
fix: optimize even if we have to make the HLG

lgray:
@martindurant this seems like a bug to me, so I am removing it.


martindurant (approved):
Yes, I suppose you are right. The original thinking was, I think, "well it's not awkward/blockwise, so no opts will help"; but of course there can be elements like that further up the layer stack.

PR #470, opened 2024-02-12
fix: projected AwkwardInputLayers have __dask_tokenize__

lgray:
Another possible fix to #468, perhaps the simplest.

@agoose77 @martindurant @jpivarski

### DIFF
--- a/src/dask_awkward/layers/layers.py
+++ b/src/dask_awkward/layers/layers.py
@@ -117,6 +117,19 @@ def io_func_implements_report(func: ImplementsIOFunction) -> bool:
     return hasattr(func, "return_report")
 
 
+class AwkwardTokenizable:
+
+    def __init__(self, ret_func, parent_name):
+        self.parent_name = parent_name
+        self.ret_func = ret_func
+
+    def __dask_tokenize__(self):
+        return (AwkwardTokenizable, self.parent_name)
+
+    def __call__(self, *_, **__):
+        return self.ret_func
+
+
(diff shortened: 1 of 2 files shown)


src/dask_awkward/layers/layers.py
```diff
@@ -117,6 +117,19 @@ def io_func_implements_report(func: ImplementsIOFunction) -> bool:
...
 
 
+class AwkwardTokenizable:
+
+    def __init__(self, ret_func, parent_name):
+        self.parent_name = parent_name
+        self.ret_func = ret_func
+
+    def __dask_tokenize__(self):
+        return (AwkwardTokenizable, self.parent_name)
```
martindurant, line 127: May as well 
```suggestion
        return ("AwkwardTokenizable", self.parent_name)
```
or other literal we make just once

src/dask_awkward/layers/layers.py
```diff
@@ -117,6 +117,19 @@ def io_func_implements_report(func: ImplementsIOFunction) -> bool:
...
+
+    def __call__(self, *_, **__):
+        return self.ret_func
```
martindurant, line 130:
```suggestion
        return self.ret_func()
```
?

lgray: the original code says otherwise `io_func=lambda *_, **__: new_return,`

lgray: right that is a value not a function :D

Issue #472, opened 2024-02-16
CuPy buffers error triggered from dask_awkward.from_parquet


NJManganelli:
A file with a perhaps ususual (or bad) parquet schema is triggering an awkward error when computed on via dask_awkward.from_parquet.

```
# all imports included in the event it takes more than just dak to reproduce
import coffea
import numpy as np
import numba as nb
import awkward as ak
import dask_awkward as dak
import dask
import uproot
import hist
from hist.dask import Hist
import trino
import matplotlib.pyplot as plt
import s3fs
import fsspec
from minio import Minio
from coffea.nanoevents import NanoEventsFactory, BaseSchema


joined = dak.from_parquet("join_out_cupy_error_triggered.parquet")
ak.num(joined.electrons, axis=1).compute()
```

results in the following
(coffea: 2024.2.1  ak: 2.6.1  uproot: 5.2.2  hist: 2.7.2  dask-awkward: 2024.2.0  numpy: 1.26.4  numba: 0.59.0  trino: 0.327.0 as package versions)

```
---------------------------------------------------------------------------
AssertionError                            Traceback (most recent call last)
Cell In[32], line 2
      1 joined3 = dak.from_parquet("join_out_cupy_error_triggered.parquet")
----> 2 ak.num(joined3.electrons, axis=1).compute()

File [/usr/local/lib/python3.11/site-packages/dask/base.py:379](http://localhost:8190/usr/local/lib/python3.11/site-packages/dask/base.py#line=378), in DaskMethodsMixin.compute(self, **kwargs)
    355 def compute(self, **kwargs):
    356     """Compute this dask collection
    357 
    358     This turns a lazy Dask collection into its in-memory equivalent.
   (...)
    377     dask.compute
    378     """
--> 379     (result,) = compute(self, traverse=False, **kwargs)
    380     return result

File [/usr/local/lib/python3.11/site-packages/dask/base.py:665](http://localhost:8190/usr/local/lib/python3.11/site-packages/dask/base.py#line=664), in compute(traverse, optimize_graph, scheduler, get, *args, **kwargs)
    662     postcomputes.append(x.__dask_postcompute__())
    664 with shorten_traceback():
--> 665     results = schedule(dsk, keys, **kwargs)
    667 return repack([f(r, *a) for r, (f, a) in zip(results, postcomputes)])

[...]

PR #473, opened 2024-02-16
fix: build proper mask in `unproject_layout`

agoose77:
Fixes a bug reported by @NJManganelli in which an un-needed mask was not specialised to the appropriate `valid_when` value.

@NJManganelli can we vendor your reproducer parquet file for our tests?


agoose77:
TODO: move `unproject_layout` to Awkward.

agoose77:
@NJManganelli does this fix #472 as well? I assume not.

NJManganelli:
@agoose77 

Indeed this fixes the issue from the slack convo; issue 472 remains and still raises the same CuPy buffers error message

Cheers!

NJManganelli:
And yes, feel free to add that file into the test cauldron!

PR #477, opened 2024-02-22
perf: private map partitions

lgray:
This PR splits `map_parititions` into itself and an unsafe version `_map_partitions` which does not make the effort to sanitize input functions or data.

This avoids creating an ArgsKwargsPackedFunction when it is not necessary to do so in routine operations like getitem and fancy indexing.

Saves O(1)s in a complex analysis.

### DIFF
--- a/src/dask_awkward/lib/core.py
+++ b/src/dask_awkward/lib/core.py
@@ -1909,7 +1909,13 @@ def _map_partitions(
     **kwargs: Any,
 ) -> Array:
     """Map a callable across all partitions of any number of collections.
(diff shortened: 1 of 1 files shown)


martindurant:
I see nothing wrong with this at all. Where do we document (for ourselves) when we can call this internal method?

lgray:
@martindurant added a helpful description.

lgray:
anything blocking this one?

martindurant (approved):

martindurant:
Let's push this one in, and I'll deal with the caching PR

PR #478, opened 2024-03-01
feat: more obvious names for report products

lgray:
Call the report output `report` and the from-uproot part `from-uproot` (or whatever input file format it happens to be) but with a different token key.

### DIFF
--- a/src/dask_awkward/lib/io/io.py
+++ b/src/dask_awkward/lib/io/io.py
@@ -643,7 +643,9 @@ def from_map(
             res = result.map_partitions(
                 first, meta=array_meta, label=label, output_divisions=1
             )
-            rep = result.map_partitions(second, meta=empty_typetracer(), label=f"{label}-report")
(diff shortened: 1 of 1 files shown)


src/dask_awkward/lib/io/io.py
```diff
@@ -640,8 +640,10 @@ def from_map(
 
     if io_func_implements_report(io_func):
         if cast(ImplementsReport, io_func).return_report:
-            res = result.map_partitions(first, meta=array_meta, output_divisions=1)
-            rep = result.map_partitions(second, meta=empty_typetracer())
+            res = result.map_partitions(
+                first, meta=array_meta, label=label, output_divisions=1
+            )
+            rep = result.map_partitions(second, meta=empty_typetracer(), label="report")
```
martindurant, line 646: Perhaps `f"{label}-report"`?

martindurant (approved):

PR #479, opened 2024-03-01
feat: use tree reduction to aggregate i/o report

lgray:
This lets us scale past ~30k input partitions without hitting a recursion error from ak.concatenate.

@nsmith- give this a try if you have time, should only need a client-side update.

### DIFF
--- a/src/dask_awkward/lib/io/io.py
+++ b/src/dask_awkward/lib/io/io.py
@@ -643,7 +643,9 @@ def from_map(
             res = result.map_partitions(
                 first, meta=array_meta, label=label, output_divisions=1
             )
-            rep = result.map_partitions(second, meta=empty_typetracer(), label=f"{label}-report")
+            rep = result.map_partitions(
+                second, meta=empty_typetracer(), label=f"{label}-report"
+            )
             return res, rep
 
     return result
--- a/src/dask_awkward/lib/io/io.py
+++ b/src/dask_awkward/lib/io/io.py
@@ -643,7 +643,7 @@ def from_map(
             res = result.map_partitions(
                 first, meta=array_meta, label=label, output_divisions=1
             )
-            rep = result.map_partitions(second, meta=empty_typetracer(), label="report")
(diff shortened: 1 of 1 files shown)


lgray:
Making this a separate PR from the nice label names since it brings additional functionality.

nsmith-:
I think the root issue is that in the array finalizer
https://github.com/dask-contrib/dask-awkward/blob/3ad9b49ed942edc69edd55745e1c1a424e912c23/src/dask_awkward/lib/core.py#L826-L828
there is a scaling issue with `ak.concatenate`. Probably this should be fixed in awkward itself. Looking into it.

lgray:
It would be good to fix the root cause for sure - however the tree reduce here also comes with some useful side effects.. Less cluster memory use and less of a penalty / more resiliency to losing intermediate data products that eventually get returned to the user.

nsmith-:
Yeah one can certainly question the wisdom of holding and collecting N separate report futures from the cluster, when N is easily O(100k)

nsmith-:
I can confirm that this patch works.

nsmith-:
Yes, I saw that the tree reductions for both the main data processing and the IO report aggregation were happening in parallel.

src/dask_awkward/lib/io/io.py
```diff
@@ -640,8 +642,48 @@ def from_map(
...
+            rep_part = result.map_partitions(
+                second, meta=empty_typetracer(), label=f"{label}-partitioned-report"
+            )
+
+            rep_trl = AwkwardTreeReductionLayer(
+                name=rep_trl_name,
+                name_input=rep_part.name,
+                npartitions_input=rep_part.npartitions,
+                concat_func=concat_fn,
+                tree_node_func=lambda x: x,
```
martindurant, line 670: We should just define `identity()`. Pretty sure dask has one.

martindurant (approved):
One nit

src/dask_awkward/lib/io/io.py
lgray: I'll get this in a housekeeping PR in just a sec.

PR #482, opened 2024-03-08
docs: dask-awkward needs a logo

jpivarski:
I'm making a poster for a conference, and so everything needs a logo.

This is based on the [Dask horizontal logo](https://docs.dask.org/en/stable/logos.html), and has a very subtle nod to the [Awkward Array logo](https://github.com/scikit-hep/awkward/tree/main/docs-img/logo).


martindurant:
Simple can be good :)
Since I am epically bad at these things, you can have any image you like.

lgray:
I might suggest offsetting one of the dask logo color squares too if you are able. Otherwise LGTM!

jpivarski:
Maybe, but there's only three squares, so misaligning one of them wouldn't look unintentional enough. I tried making the middle one poke through, but then it just looks too busy. Often, logos are seen from a distance, so they can't rely on fine detail for recognition. (It's hard to make useful favicon.ico files!)

I think I'll just merge as-is. Thanks for your comments!

PR #483, opened 2024-03-11
feat: support slicing with int16/int8 index arrays

veprbl:


lgray:
@veprbl could you please add corresponding test variants? Thanks!

martindurant:
Would perhaps checking `dtype.kind in "bui"` (also including unsigned here, which ought to be allowed) be an easier check than listing all those types? Or do they perhaps have a common superclass?

jpivarski:
```python
            if issubclass(
                dtype, (np.bool_, bool, np.int64, np.int32, np.int16, np.int8, int)
            ):
```

The `np.int64, np.int32, np.int16, np.int8` group has a common superclass in NumPy called `np.signedinteger`. (If you also want the unsigned cases, there's `np.integer`.) NumPy booleans (`np.bool_`) are not subclasses of integers the way that Python booleans are. The Python `bool` and `int` do not share any superclasses with NumPy booleans and numbers, apart from `object`, so they would still need to be checked.

A better runtime type test for "anything like an integer" is `numbers.Integral`, since that's a protocol, rather than relying on explicit inheritance. It will catch all integers in any library, including unsigned, including Python builtins and NumPy numbers. Since Python `bool` is an `int`, it catches `bool`, but _not_ `np.bool_`.

So assuming that you're okay with unsigned integers, the above can be replaced by

```python
            if issubclass(dtype, (np.bool_, numbers.Integral))
```

As a side note, I think it's confusing to name an `np.dtype.type` as "`dtype`", since this distinction between NumPy types described as dtype objects (instances of `np.dtype`) and NumPy types described as types (classes, instances of `type`) is already confusing. It would be best to not mix names.

martindurant:
> I think it's confusing to name an np.dtype.type as "dtype", since this distinction between NumPy types described as dtype objects 

Could it be phrased as `isinstance` in this case, or would my `.kind` suggestion be an easier-to-read alternative?

veprbl:
I've implemented based on nice suggestion from @jpivarski and a test as requested by @lgray .

The commits won't show up right now, but you can see them on the branch https://github.com/dask-contrib/dask-awkward/compare/main...veprbl:dask-awkward:patch-2?expand=1

lgray:
@veprbl just needed a close/reopen to kick it back into action

martindurant:
Looks like we must add `dask-expr` to the awkward-main job

martindurant:
Looks like we nee to add dask-expr to https://github.com/dask-contrib/dask-awkward/blob/main/pyproject.toml#L40 , the "test" deps, or at least to https://github.com/dask-contrib/dask-awkward/blob/main/.github/workflows/awkward-main.yml#L31

Not sure which is most appropriate yet. Dask being installable but refusing to work without dask-expr is ... strange.

lgray:
@martindurant it's only the dataframe test that is failing, FWIW.

lgray:
And it is a new requirement in the `dask[dataframe]` extras install. https://github.com/dask/dask/blob/2024.3.0/pyproject.toml#L58

PR #486, opened 2024-03-15
feat: turn off ak's typetracer _new checks during optimize

martindurant:
Will have no effect until https://github.com/scikit-hep/awkward/pull/3054 is out

@lgray , any other places we need this optimisation?

### DIFF
--- a/src/dask_awkward/lib/inspect.py
+++ b/src/dask_awkward/lib/inspect.py
@@ -177,36 +177,39 @@ def report_necessary_columns(
     seen_names = set()
 
     name_to_necessary_columns: dict[str, frozenset | None] = {}
-    for obj in collections:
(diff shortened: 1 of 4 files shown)


lgray:
During `dak.necessary_columns` as well. Though that has some slowness that's kind of strange but different, it hasn't been high priority. This will help, of course.

jpivarski:
If you specify the lines of code that should be avoided, I'll add them to this PR (with only one global switch for all).

lgray:
Yeah I haven't fully thought through or experienced the user-side lifecycle of `necessary_columns`... I have no idea if people use it a lot directly and if it needs to be snappy.

martindurant:
OK, will merge this now, and we can add more blocks or set a global state in the future, if we decide.

PR #491, opened 2024-03-28
ENH: One-pass column optimization

martindurant:


src/dask_awkward/lib/io/columnar.py
```diff
@@ -74,105 +55,3 @@ def mock_empty(self: S, backend: BackendT = "cpu") -> AwkwardArray:
...
-        paths = set()
-        wildcard_form_key = set()
-        for buffer_key in data_buffers:
-            form_key, attribute = parse_buffer_key(buffer_key)
-            if attribute in METADATA_ATTRIBUTES:
-                wildcard_form_key.add(form_key)
-            else:
-                paths.add(form_key_to_path[form_key])
-
-        # Select the most appropriate column for each wildcard
```
agoose77, line 136: @martindurant is the plan to restore e.g. wildcard matching in this PR?

martindurant: I don't think that happens in this PR, this one's just about auto-selecting required columns. We might need more work for the case where the shape of something is required but not the data.

agoose77: I'm only surface-glancing, but I noted that this logic has been removed and didn't see anywhere that it was recovered. Might have missed something!

martindurant: You are right, this should live after checking shape_touched_in, more complex than I had hoped.

martindurant: OK< I have some shonky code in optimize.optimize_columns to deal with shapes. I am probably still missing some touch-buffers on the outputs, but most things are in line. 

It's all pretty ugly and in the latest commit I seem to have introduced some typetracer->numpy coersion.

src/dask_awkward/lib/optimize.py
```diff
@@ -111,10 +111,30 @@ def optimize_columns(dsk: HighLevelGraph, keys: Sequence[Key]) -> HighLevelGraph
...
+
+def _buf_to_col(s):
+    return (
+        s[2:]
+        .replace(".content", "")
+        .replace("-offsets", "")
+        .replace("-data", "")
+        .replace("-index", "")
+        .replace("-mask", "")
+    )
```
agoose77, line 131: I don't have time *right now* to look at the full PR here, but is this assumption about buffer name mappings enforced anywhere? Right now the design is such that input sources can use a non-canonical buffer naming scheme.

For coffea I think we made this possible so that Coffea could directly pass in their own buffer names with semantic meaning.

jpivarski: Do we have access to the call to `ak.from_buffers` here? That function takes a callable that constructs the buffer names from the `form_key`s and these roles.

The roles are fixed words like this, and Coffea only has the power to change the `form_key`, not the role. This is almost the complete set of such names; only `-tags` (for UnionArray) is missing.

agoose77: @jpivarski I started writing a reply here, but I think I should just aim for explicit verbosity, and hope we end up on the same page.

@martindurant this might be useful too, not sure.

When writing the current column optimisation API for dask-awkward, I had the following intentions:
- be generic enough that coffea doesn't have to write super weird code
- make few assumptions about optimisation, let the IO func do the work

Crucially, the optimisation strategies differ between Parquet, ROOT, and remapped-ROOT:
- Parquet cannot read the metadata arrays (so we have a pick-a-child strategy)
- ROOT cannot read the metadata arrays (so we have a read-the-whole-branch strategy)
- Remapped-ROOT can read individual buffers (as a single branch may correspond to an offsets buffer

To handle all of this complexity, I made column optimisation in Dask Awkward a thin contract; the strategy for optimisation remains an implementation-specific feature. We implement an optimisation strategy for "column" (pick-one-child) formats, but not for the ROOT/Remapped-ROOT cases.
The contract imposed upon IO functions wishing to implement a strategy is:

- `IOFunc.prepare_for_projection()` returns:
  - `array` - an Awkward `Array` that has buffers already named
  - `report` - a `TypeTracerReport`
  - `state` - additional state required later by `project()`
- `IOFunc.project(report, state)` returns:
  - `layer` - the projected layer


As such, Dask Awkward doesn't need to know about buffer / form keys in its optimisation machinery.

I think where possible we should keep this "black-box" approach, and have the IO-func perform the strategy for optimisation.

agoose77:
@jpivarski I started writing a reply here, but I think I should just aim for explicit verbosity, and hope we end up on the same page.

@martindurant this might be useful too, not sure.

When writing the current column optimisation API for dask-awkward, I had the following intentions:
- be generic enough that coffea doesn't have to write super weird code
- make few assumptions about optimisation, let the IO func do the work

Crucially, the optimisation strategies differ between Parquet, ROOT, and remapped-ROOT:
- Parquet cannot read the metadata arrays (so we have a pick-a-child strategy)
- ROOT cannot read the metadata arrays (so we have a read-the-whole-branch strategy)
- Remapped-ROOT can read individual buffers (as a single branch may correspond to an offsets buffer

To handle all of this complexity, I made column optimisation in Dask Awkward a thin contract; the strategy for optimisation remains an implementation-specific feature. We implement an optimisation strategy for "column" (pick-one-child) formats, but not for the ROOT/Remapped-ROOT cases.
The contract imposed upon IO functions wishing to implement a strategy is:

- `IOFunc.prepare_for_projection()` returns:
  - `array` - an Awkward `Array` that has buffers already named
  - `report` - a `TypeTracerReport`
  - `state` - additional state required later by `project()`
- `IOFunc.project(report, state)` returns:
  - `layer` - the projected layer


As such, Dask Awkward doesn't need to know about buffer / form keys in its optimisation machinery.

I think where possible we should keep this "black-box" approach, and have the IO-func perform the strategy for optimisation.

martindurant:
I think you are effectively saying we should move the text manipulation to the IOFunction superclass so it can be overridden - which I am fine with. I don't otherwise see the consequence of what you are saying and why you need extra methods. As far as I could tell, the old technique was al about making layers that could go through the "no consequence" run of the graph at optimization time. From my point of view, all IO layers should implement project(), but with the option that it does nothing. (Or, alternatively, a meaningful project() could be what defines an IO Layer, since this is the only thing we need them for).

Your engagement is well appreciated! I hope you can help push this through, as it has turned out much harder than I had initially thought, and the implicit contracts in the various protocols are exactly what confused me the most. Now, I believe the only issue with the current tests is some cached state: either the explicit caches, or a missing commit().

agoose77:
> as it has turned out much harder than I had initially thought

I have a lot of sympathy for this, it's a rabbit hole.

> the implicit contracts in the various protocols are exactly what confused me the most

Yes, in writing my reply I came to realise that the uproot code is really adding to the complexity of thinking about this, even though it's effectively separate. Having to consider coffea-via-uproot on awkward-dask is hard.

> I hope you can help push this through

Will be glad too.

agoose77:
I'm also mindful that we want to be clear about what "column" means; I think dask-awkward should/can only reason about _buffers_, and it's the IO-func's job to figure out whether that means "columns" (i.e. coffea reports the TTree branches for `necessary_columns`, whereas Parquet reports the attribute-path to the leaf).

agoose77:
@martindurant also, I'd like to apologise for how long it took to get that detailed reply onto paper; my personal life has had quite a lot going on recently (new house among them!) that means I've had much less spare time than anticipated.

martindurant:
OK, so agreed: we can maintain the buffer names while optimising, and only normalise to the IO-specific column names within the IOFunc's project() method, so that different backends can handle that differently. Also I can get rid of the "@." part, which I copied from the previous optimisation code and I don't think it does anything.

@jpivarski , this is a place where "."-containing field names might be a problem:
- when the optimisation says that you need the offsets for some record field, we look for any column beginning with that schema path. So if there is a field "field" and another "field.name" which is NOT a sub-field of the first, we might erroneously select it.

martindurant:
> we can maintain the buffer names while optimising, and only normalise to the IO-specific column names within the IOFunc's project() method

done

agoose77:
I'll hopefully take a look at this before our meeting tomorrow :)

martindurant:
Ping @agoose77 , do you think you have time to look at this?

agoose77:
@martindurant one question that came to me a while ago (only just actioning now) is - why do we store the report on the layer? Can we not just retrieve it from the `meta`? We don't expose an `ak.` function to do that yet, but it's basically:
```python
assert ak.backend(array) == "typetracer"
form, length, container = ak.to_buffers(array)
reports = {b.report for b in container.values() if b is not None}
```

martindurant:
If we only need to store the meta on the layer, fine. But we cannot access the meta of the collection (dak.Array) at optimize time, since dask only passes the combined graph of all the collections that were passed to compute().

lgray:
I'll probably need this test to function:
```
FAILED tests/test_inspect.py::test_basic_root_works - AttributeError: '_UprootRead' object has no attribute 'form'
```

To be able to run any of my tests.

martindurant:
Oh, right - fair enough. I'll look at what patch root needs. In this codebase, of course we only ensure parquet and json.

martindurant:
tests/test_inspect.py::test_basic_root_works is fixed with https://github.com/scikit-hep/uproot5/pull/1225

The two remaining failures are:
- getting an exception report is causing touching of columns (execution reports should drop all column touching reports?)
- a behaviour/attribute gets lost, presumably it is set on a meta that gets overwritten

src/dask_awkward/lib/core.py
```diff
@@ -518,6 +523,7 @@ def f(self, other):
                 meta = op(self._meta, other._meta)
             else:
                 meta = op(self._meta, other)
+            [_.commit(name) for _ in self.report]
```
agoose77, line 526: Let's lift this into a utility function e.g. 
```python
def commit_to_reports(name, reports):
    for report in reports:
        report.commit(name)
```

src/dask_awkward/lib/core.py
```diff
@@ -398,6 +399,10 @@ def name(self) -> str:
     def key(self) -> Key:
         return (self._name, 0)
 
+    @property
+    def report(self):
+        return getattr(self._meta, "_report", set())
```
agoose77, line 404: Can we lift `_report` from the meta to the Layer? I know that it's more state to keep in sync, but I don't love the act of tagging non-reserved attributes onto an Awkward Array, and this will better separate the stateful part of typetracer from the stateless part.

src/dask_awkward/lib/io/io.py
```diff
@@ -620,16 +625,34 @@ def from_map(
...
-    # If we know the meta, we can spoof mocking
-    elif meta is not None:
-        io_func = IOFunctionWithMocking(meta, func)
-        array_meta = meta
+        array_meta, report = typetracer_with_report(
+            form_with_unique_keys(io_func.form, "@"),
+            highlevel=True,
+            behavior=io_func.behavior,
+            buffer_key=render_buffer_key,
+        )
```
agoose77, line 638: This is the big change:

Control over buffer names now solely rests with dask-awkward, rather than the provider of the IO function e.g. uproot. 

This means that dask-awkward generates buffer names and gives them to the IO-function, meaning that the IO-functions now need to understand how to parse these keys instead of dask-awkward asking the IO-functions to parse them. This is all semantics; this is a trivial operation and we can define public functions in dask-awkward to do this.

Defining our column-optimisation interface at the Form level still ensures unique names, but it now requires that the IO-function understands how to map from the form to the buffers it works.[^vm] A nice perk of this change is that because dask-awkward now controls buffer names, we can ensure that they are globally unique (i.e. between different input layers) e.g. by embedding the layer ID. This might prove useful.

[^vm]: To understand the consequence of this, I need to more intimately remind myself how coffea currently remaps keys (where keys are VM instructions) to that which it presents to uproot.

src/dask_awkward/lib/optimize.py
```diff
@@ -178,70 +102,80 @@ def optimize_columns(dsk: HighLevelGraph, keys: Sequence[Key]) -> HighLevelGraph
...
+        new_lay = lay.project(cols)
+        dsk2[k] = new_lay
+
+    return HighLevelGraph(dsk2, dsk.dependencies)
+
+
+def _optimize_columns(dsk, all_layers):
+    for k, lay in dsk.copy().items():
+        if not isinstance(lay, AwkwardInputLayer) or not hasattr(
+            lay.io_func, "_column_report"
```
agoose77, line 133: Just like with `_report`, I'd like to formalise this as part of the API. This would become a part of the column-projection interface.

src/dask_awkward/lib/optimize.py
```diff
@@ -178,70 +102,80 @@ def optimize_columns(dsk: HighLevelGraph, keys: Sequence[Key]) -> HighLevelGraph
...
+    for k, _, cols in _optimize_columns(dsk, all_layers):
+        first = (_buf_to_col(s) for s in cols)
+        # remove root "offsets", which appears when computing divisions
+        second = {s for s in first if s and s != "offsets"}
+        # remove columns that have sub-columns also in the set
+        for c in second.copy():
+            if "." in c:
+                parent = c.rsplit(".", 1)[0]
+                second.discard(parent)
+        out[k] = sorted(second)
```
agoose77, line 177: This logic assumes that buffer names are `.` delimited and are hierarchical, i.e. `foo.bar` is below `foo`. This isn't true for form remapping, which can move arbitrary buffers around. 

That is to say, the real operation we want to implement is `necessary_buffers` which happens to be `necessary_columns` for the Parquet/JSON schemes.

agoose77 (commented):
This is an initial-thoughts review. I'll propose some concrete changes where I can in the next few days. The main things to tackle in my view are:
1. Re-implement `necessary_columns` in terms of the IO function
  - "columns" in my view (and hence the existing implementation) represents the IO-function's definition of a column, i.e. it's different between uproot, and Parquet/JSON. 
  - We should ask the IO-function to translate buffer names to column names e.g. https://github.com/scikit-hep/uproot5/blob/734700ef1f822338b03a7573df484909b317b2c2/src/uproot/_dask.py#L1077
2. Understand how coffea & uproot need to change to match this new dask-awkward-driven buffer mapping

There's probably more to think about, but as I say this is just an initial thought. I recognise that the coupling between coffea and uproot is confusing, so I'll be sure to start with that.

agoose77:
I don't seem to be able to push to the repo, so here's a patch.[^1]

[^1]: I know you can do this in some contexts, but there's a nuance (perhaps with auth) that I've not yet figured out.
[0001-refactor-add-commit_to_reports.patch.zip](https://github.com/dask-contrib/dask-awkward/files/15457447/0001-refactor-add-commit_to_reports.patch.zip)

src/dask_awkward/lib/core.py
agoose77: @jpivarski I wonder whether we should add a new `merge` function to `TypeTracerReport`. Due to the uniqueness of layer IDs, and the existing association of reports with layouts, we can replace the need to carry a set of reports with a single report via
```
merge(A, B) -> A
```

I haven't yet thought about the memory implication for effectively having "larger" reports, but this might be a nicer UX.

martindurant:
Only three failures:
- assigning attributes
- IO report
- root (I have a fix for uproot in a branch, but not one that also accounts for nanoevents remapping)

PR #505, opened 2024-05-15
feat: Implement dak.nan_to_num

### LINKED ISSUE #504: `dak.nan_to_num()` not Implemented
I am implementing some corrections that involves dividing the result from one look-up table and another. Sometimes it is the case that there is division by zero, and the `nan` values need to be converted back to 0. I would like to ask that this unsupported call be supported.

martindurant:
Fixes #504


martindurant:
We also seem to be missing nan_to_none, which could have been another way to solve the original issue. I don't know if we keep a list of API calls from upstream that should be implemented, but I am assuming this wasn't among the list when dak.* TODO functions were written out.

PR #514, opened 2024-06-19
WIP: Use threaded dask for optimize pass

martindurant:
As described in slack, this change appears to deliver ~30% faster optimization on my system, without breaking anything. I am posting here, so that others can test more easily.


martindurant:
I have tried this now with various queries out of https://github.com/CoffeaTeam/coffea-benchmarks/blob/master/coffea-adl-benchmarks.ipynb (some of which should be sufficiently complex) and I am no longer sure there is a benefit. Would be good to try with even bigger coffea workflows.

PR #520, opened 2024-06-26
Add __array_function__ to dispatch NumPy funcs to dask-awkward

supercoder-dev:
Add NumPy function dispatching to dask-awkward

This PR adds an __array_function__ method to the Array class, allowing NumPy functions to be automatically redirected to their dask-awkward counterparts. It covers 38 common functions, making it easier to use NumPy-style operations with dask-awkward arrays.

### DIFF
--- a/src/dask_awkward/lib/core.py
+++ b/src/dask_awkward/lib/core.py
@@ -1689,6 +1689,57 @@ class Array(DaskMethodsMixin, NDArrayOperatorsMixin):
             **kwargs,
         )
 
+    def __array_function__(self, func, types, args, kwargs):
+        import dask_awkward as dak
+
+        # List of supported functions
+        supported_funcs = {
+            "amin": dak.amin,
+            "nanmin": dak.nanmin,
+            "count_nonzero": dak.count_nonzero,
+            "mean": dak.mean,
+            "nanmean": dak.nanmean,
+            "concatenate": dak.concatenate,
+            "sum": dak.sum,
+            "nansum": dak.nansum,
+            "argsort": dak.argsort,
+            "copy": dak.copy,
+            "nan_to_num": dak.nan_to_num,
+            "ones_like": dak.ones_like,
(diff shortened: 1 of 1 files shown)


src/dask_awkward/lib/core.py
```diff
@@ -1689,6 +1689,57 @@ def __array_ufunc__(self, ufunc, method, *inputs, **kwargs):
...
+            "ravel": dak.ravel,
+            "argmin": dak.argmin,
+            "nanargmin": dak.nanargmin,
+        }
+
+        # Check if the function is supported
+        if func.__name__ in supported_funcs:
+            return supported_funcs[func.__name__](*args, **kwargs)
+        else:
+            return NotImplemented
```
martindurant, line 1741:
```suggestion
            return NotImplemented(str(func))
```

src/dask_awkward/lib/core.py
```diff
@@ -1689,6 +1689,57 @@ def __array_ufunc__(self, ufunc, method, *inputs, **kwargs):
             **kwargs,
         )
 
+    def __array_function__(self, func, types, args, kwargs):
+        import dask_awkward as dak
+
+        # List of supported functions
+        supported_funcs = {
```
martindurant, line 1696: Mildly prefer this as a global constant rather than in the function

martindurant (commented):
Could use tests.

However: isn't the purpose of array_function to return a numpy view on the data rather than (as with ufuncs) to apply the numpy idea to producing an array of our own type? Maybe I misunderstand.

lgray:
I'd be a bit careful with this kind of dispatch. The numpy functions can have subtly different behaviors from their awkward counterparts. This mapping isn't very well defined for the arbitrary numpy functions (but is for ufuncs).

It would almost be better to suggest converting to `dask.array` and error out since there the expected behavior is much more clear.

PR #529, opened 2024-07-29
fix: Scalar typetracer after reduction

### LINKED ISSUE #528: Left- vs right-multiplying array by Dask scalar
The following raises
```pytb
NotImplementedError: no overload for ak.to_layout with arguments (dask.awkward<from-awkward, npartitions=1>,)
```
when attempting to multiply the array by a scalar from the left:
```python
import awkward as ak
import dask_awkward as dak

arr = dak.from_awkward(ak.Array([1,3,4]), npartitions=1)
dak.sum(arr, axis=-1) * arr
```
Doing the same thing from the right works:
```python
arr * dak.sum(arr, axis=-1)
```

I'm using `awkward` 2.6.6 and `dask_awkward` 2024.6.0.

martindurant:
Fixes #528 

This appears to fix the linked issue, but I'm not really sure why a typetracer object is being passed around rather than a high-level Array.

### DIFF
--- a/src/dask_awkward/lib/core.py
+++ b/src/dask_awkward/lib/core.py
@@ -500,19 +500,17 @@ class Scalar(DaskMethodsMixin, DaskOperatorMethodMixin):
             deps = [self]
             plns = [self.name]
             if is_dask_collection(other):
-                task = (op, self.key, *other.__dask_keys__())
                 deps.append(other)
+                plns.append(other.name)
                 if inv:
-                    plns.insert(0, other.name)
+                    task = (op, *other.__dask_keys__(), self.key)
                 else:
-                    plns.append(other.name)
+                    task = (op, self.key, *other.__dask_keys__())
             else:
                 if inv:
                     task = (op, other, self.key)
                 else:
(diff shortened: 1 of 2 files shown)


src/dask_awkward/lib/core.py
```diff
@@ -532,6 +530,11 @@ def f(self, other):
                     meta = op(other, self._meta)
                 else:
                     meta = op(self._meta, other)
+            if meta.ndim:
```
martindurant, line 533: This applies to the scenario, which apparently we hadn't anticipated, that an operation on a scalar produces an array.

src/dask_awkward/lib/core.py
```diff
@@ -570,6 +573,15 @@ def f(*args):
         args = tuple(
             ak.Array(arg.content) if isinstance(arg, MaybeNone) else arg for arg in args
         )
+        args = tuple(
```
martindurant, line 576: Here we account for getting the low-level typetracer through during column optimisation. This is the questionable part.

PR #542, opened 2024-09-06
[feat] Explicitly raise `TracerConversionError` when a tracer is used in a context that requires a concrete value

### LINKED ISSUE #538: Conditional branching and __bool__ for arrays
In discussions with @jpivarski and @pfackeldey at pyhep.dev we identified a few aspects of behavior that I believe could be improved.

### Short version
- `bool()` on a `dask-awkward` is `True` at the moment, which can result in unintended behavior
- automatic evaluation of Dask expressions to determine conditional branching is unintuitive and I believe instead raising an exception with `dask-awkward` might be a better idea

### `bool()` on a `dask-awkward` array

```python
import dask_awkward as dak

arr = dak.from_lists([[0, 0]])
bool(arr)
```

evaluates as `True`. This seems independent of the content of the array. The example above should raise an exception. If it is already known that the `bool()` does not make sense, such in this case where the argument is not a scalar, I believe an exception such as what `numpy` provides makes sense:

```python
import dask.array as da
import numpy as np

arr = da.array(np.zeros(2))
bool(arr)
```

-> `ValueError: The truth value of a Array is ambiguous. Use a.any() or a.all().`

### Automatic computation for branching

As a consequence of the behavior shown above, where the array evaluates as `True`, branching based on the content of a dask-awkward array is not working correctly.

```python
import uproot
import dask
import awkward as ak

ttbar_file = "https://github.com/scikit-hep/scikit-hep-testdata/"\
    "raw/main/src/skhep_testdata/data/nanoAOD_2015_CMS_Open_Data_ttbar.root"

evts = uproot.dask({ttbar_file: "Events"})
evts["meta"] = ak.zeros_like(evts.run, dtype="float")

def f(evts):
    if ak.sum(evts["meta"]) > 100:
        res = {"num_electrons": ak.sum(ak.num(evts.Electron_pt, axis=1))}
    else:
        res = {"num_muons": ak.sum(ak.num(evts.Muon_pt, axis=1))}
    return res

dask.compute(f(evts))
```

This returns `({'num_electrons': 69},)`, as `ak.sum(evts["meta"]) > 100` is `True`, despite 

```python
bool((ak.sum(evts["meta"]) > 100).compute())
```

being `False` when actually computed. The behavior here departs from what Dask does. Consider the following variation with `dask.array`:

```python
import uproot
import dask
import awkward as ak
import numpy as np
import dask.array as da

ttbar_file = "https://github.com/scikit-hep/scikit-hep-testdata/"\
    "raw/main/src/skhep_testdata/data/nanoAOD_2015_CMS_Open_Data_ttbar.root"

evts = uproot.dask({ttbar_file: "Events"})
meta_arr = da.array(np.zeros(200))

def f(evts, meta_arr):
    if np.sum(meta_arr) > 100:
        res = {"num_electrons": ak.sum(ak.num(evts.Electron_pt, axis=1))}
    else:
        res = {"num_muons": ak.sum(ak.num(evts.Muon_pt, axis=1))}
    return res

dask.compute(f(evts, meta_arr))
```

In this case the last line evaluates as `({'num_muons': 41},)`, which is the correct answer. What happens as far as I understand is that a `.compute()` is triggered when building the graph and encountering `if np.sum(meta_arr) > 100`: [dask/array/core.py#L1861](https://github.com/dask/dask/blob/3003db5b84070b1afc197c7c70c76c5c4c2bc821/dask/array/core.py#L1861) (same for scalars [dask/array/core.py#L1869](https://github.com/dask/dask/blob/3003db5b84070b1afc197c7c70c76c5c4c2bc821/dask/array/core.py#L1869)).

The behavior of `dask.array` ultimately causes the expected result to be produced, however the way in which this happens might be surprising. A possibly expensive computation can happen during what otherwise would just be graph building, in a way that is not explicitly communicated to the user.

The automatic `.compute()` in certain cases is also the reason for why
```python
import dask.array as da
import numpy as np
bool(da.array(np.zeros(1))), bool(da.array(np.ones(1)))
```
evaluates to `(False, True)` without any `.compute()` visible.

### Thoughts on expected behavior

I find the automatic `.compute()` very unintuitive. Users coming from pure Dask might be used to it, so I can see arguments for following the same pattern. I would however propose to raise an exception regardless when a `bool()` evaluation on a dask-awkward needs to happen during tracing, informing the user that they might want to consider calling `.compute()` manually. In the first example, this would look like the following:

```python
if (ak.sum(evts["meta"]) > 100).compute():
```

instead of:

```python
if ak.sum(evts["meta"]) > 100:
```

In high energy physics applications, typical use cases I have in mind where branching is needed are different processing pipelines for different types of input files. In these cases I believe it is desirable to not evaluate all possible ways to trace through the function, which I think is what a `np.where` does in contrast (evaluate both and then combine afterwards). This helps prevent potentially expensive calculations. Whenever possible, such branching should probably be made with static or already evaluated metadata as opposed to in-file non-evaluated metadata to prevent complications from the behavior listed above.

This behavior might catch users off guard, especially if they port awkward code over to dask-awkward. I am not sure where to best communicate this as a possible "gotcha" but I think it could be useful to spread awareness.

pfackeldey:
This fixes #538 by adopting to dask.Array's behavior. 

However, I think - in the long run - we should aim for avoiding "silent" `.compute()` calls like these, and rather raise exceptions.

Best, Peter

### DIFF
--- a/src/dask_awkward/lib/core.py
+++ b/src/dask_awkward/lib/core.py
@@ -48,9 +48,9 @@ from dask.utils import funcname, is_arraylike, key_split
 from dask_awkward.layers import AwkwardBlockwiseLayer, AwkwardMaterializedLayer
 from dask_awkward.lib.optimize import all_optimizations
 from dask_awkward.utils import (
+    ConcretizationTypeError,
     DaskAwkwardNotImplemented,
     IncompatiblePartitions,
-    TracerConversionError,
     field_access_to_front,
     first,
     hyphenize,
@@ -923,19 +923,21 @@ class Array(DaskMethodsMixin, NDArrayOperatorsMixin):
     __dask_scheduler__ = staticmethod(threaded_get)
 
     def __bool__(self):
-        raise TracerConversionError(bool, self)
+        raise ConcretizationTypeError(f"The __bool__() method was called on {self!r}.")
 
     def __int__(self):
-        raise TracerConversionError(int, self)
+        raise ConcretizationTypeError(f"The __int__() method was called on {self!r}.")
 
     def __float__(self):
-        raise TracerConversionError(float, self)
+        raise ConcretizationTypeError(f"The __float__() method was called on {self!r}.")
 
     def __complex__(self):
-        raise TracerConversionError(complex, self)
+        raise ConcretizationTypeError(
+            f"The __complex__() method was called on {self!r}."
+        )
 
     def __index__(self):
-        raise TracerConversionError(operator.index, self)
+        raise ConcretizationTypeError(f"The __index__() method was called on {self!r}.")
 
     def __setitem__(self, where: Any, what: Any) -> None:
         if not (
--- a/src/dask_awkward/lib/core.py
+++ b/src/dask_awkward/lib/core.py
@@ -921,6 +921,36 @@ class Array(DaskMethodsMixin, NDArrayOperatorsMixin):
 
     __dask_scheduler__ = staticmethod(threaded_get)
 
+    # taken from https://github.com/dask/dask/blob/3003db5b84070b1afc197c7c70c76c5c4c2bc821/dask/array/core.py#L1854-L1883
+    def __bool__(self):
+        materialized = self.compute().to_numpy()
+        if materialized.size != 1:
+            raise ValueError(
+                f"The truth value of a {self.__class__.__name__} is ambiguous. "
+                "Use a.any() or a.all()."
+            )
+        else:
+            return bool(materialized)
+
+    def _scalarfunc(self, cast_type):
+        materialized = self.compute().to_numpy()
+        if materialized.size != 1:
+            raise TypeError("Only length-1 arrays can be converted to Python scalars")
+        else:
+            return cast_type(materialized.item())
+
+    def __int__(self):
+        return self._scalarfunc(int)
+
+    def __float__(self):
+        return self._scalarfunc(float)
+
+    def __complex__(self):
+        return self._scalarfunc(complex)
+
+    def __index__(self):
+        return self._scalarfunc(operator.index)
+
     def __setitem__(self, where: Any, what: Any) -> None:
         if not (
             isinstance(where, str)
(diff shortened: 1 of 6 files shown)


src/dask_awkward/lib/core.py
```diff
@@ -921,6 +921,36 @@ def __dask_postpersist__(self):
 
     __dask_scheduler__ = staticmethod(threaded_get)
 
+    # taken from https://github.com/dask/dask/blob/3003db5b84070b1afc197c7c70c76c5c4c2bc821/dask/array/core.py#L1854-L1883
+    def __bool__(self):
+        materialized = self.compute().to_numpy()
+        if materialized.size != 1:
```
martindurant, line 927: Can we know the `size` before compute? It seems awfully wasteful to run a graph only then to realise that we can't use the output.

martindurant: Hm. I am sorely tempted to say we should `raise NoteImplemented` for _all_ these methods ...

pfackeldey: I agree. I can add a `TracerConversionError` (similar to what JAX does) and raise it with a helpful message for all of these methods. I'll update the PR accordingly if you're fine with this procedure.

(Then dask-awkward and dask behave differently though, which might be confusing for people coming from dask. Do you think this is a problem?)

martindurant: I am not sure what the type of the exception should be, so long as the message it delivers is clear.

src/dask_awkward/utils.py
```diff
@@ -38,6 +38,49 @@ def divisions_msg(name: str, *args: Array) -> str:
         return msg
 
 
+class TracerConversionError(TypeError):
+    """
+    This error occurs when a tracer is used in a context that requires a concrete
```
martindurant, line 43: "tracer" is a strange name for it. It's always a dask collection, usually a dak.Array, right?

martindurant:
OK, I like this.

PR #547, opened 2024-10-07
docs: CONTRIBUTING.md

pfackeldey:


CONTRIBUTING.md
```diff
@@ -0,0 +1,27 @@
+# Contributing to dask-awkward
+
+Install the development environment (once) with
+
+```bash
+uv venv
```
martindurant, line 6:
```
python -m venv
```
rather than assume users have `uv` (or `conda` or whatever)

CONTRIBUTING.md
```diff
@@ -0,0 +1,27 @@
...
+uv pip install -e '.[io,docs,test]'
+pre-commit install
+```
+
+Source development environment with
+
+```bash
+source .venv/bin/activate
```
martindurant, line 14: Doesn't this need to come before pip install?

CONTRIBUTING.md
```diff
@@ -0,0 +1,27 @@
...
+```
+
+Run tests with
+
+```bash
+pytest .
```
martindurant, line 20:
```suggestion
pytest -v
```

CONTRIBUTING.md
```diff
@@ -0,0 +1,27 @@
...
+```
+
+Build documentation with
+
+```bash
+cd docs && make html
```
martindurant, line 26: Does `&&` work on windows?

Also worth saying where the docs end up.

martindurant (commented):
Have you seen https://github.com/dask-contrib/dask-awkward/blob/main/docs/dev/contributing.rst ?

pfackeldey:
Thank you very much for pointing to the contribution guide in the documentation - I didn't see it!
I updated this PR and linked it directly in the README.md; I hope this sounds reasonable to you.

CONTRIBUTING.md
pfackeldey: In general yes, but not with `uv`. If it finds a venv (even though it is not activated) it will install software into it.

martindurant: OK, so if we remove requiring `uv`, then this will need to be reordered. That's conceptually cleaner anyway, if you ask me.

PR #554, opened 2024-12-03
fix: allow renaming when rebuilding dask-awkward Arrays

pfackeldey:
This follows the implementation of `dask.array.Array`: https://github.com/dask/dask/blob/main/dask/array/core.py#L1442

With this PR it is possible to `dask.graph_manipulation.clone` graphs containing `dask_awkward.Arrays`.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -130,3 +130,6 @@ venv.bak/
 
 # mypy
 .mypy_cache/
+
+# pyright
+pyrightconfig.json
--- a/src/dask_awkward/lib/core.py
+++ b/src/dask_awkward/lib/core.py
@@ -381,7 +381,7 @@ class Scalar(DaskMethodsMixin, DaskOperatorMethodMixin):
     def _rebuild(self, dsk, *, rename=None):
         name = self._name
         if rename:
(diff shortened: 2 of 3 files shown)


martindurant:
I am attending pydata global right now, but I'll get to this. I notice that coverage reports that your code isn't touched.

martindurant:
It looks like this is ready? It maybe should have an explicit test for `rename`. The failures are on a nearly-equal that seems to pop up occasionally (based on random input). I don't see why anything here would particularly trigger them; but maybe it would be worth while making the assert-nearly-equal give us more information about what doesn't match, to see if there's something systematic we should fix.

pfackeldey:
Yes, it's ready :+1:
I'll add a test later, I'm just not sure what you mean with 'assert-nearly-equal'?

martindurant:
> what you mean with 'assert-nearly-equal'?

I meant calling `assert ak.almost_equal` within `assert_eq_arrays` .

...but actually I see that a_comp is [all zeros](https://github.com/dask-contrib/dask-awkward/actions/runs/12142271980/job/33856342244?pr=554#step:5:1064), so something else is going on?

PR #555, opened 2024-12-03
fix: tokenization of ArgsKwargsPackedFunction

### LINKED ISSUE #553: wrong `token` generation with `dak.map_partitions`
Currently the `token` generation for the HLG layer with `dak.map_partitions` seems to be broken, i.e. it changes everytime, e.g.:

```python
import awkward as ak
import dask_awkward as dak

dak_array = dak.from_awkward(ak.Array([1, 2, 3]), 1)

def fun(x):
  return x + 1

dak.map_partitions(fun, dak_array).dask
# HighLevelGraph with 2 layers.
# <dask.highlevelgraph.HighLevelGraph object at 0x10577cd30>
# 0. from-awkward-8fdcc693b11857efa42a556ced92dfbc
# 1. <dask-awkward.lib.core.ArgsKwargsPackedFunction ob-7ae414b04e6de108142215f755b2eb3e

dak.map_partitions(fun, dak_array).dask
# HighLevelGraph with 2 layers.
# <dask.highlevelgraph.HighLevelGraph object at 0x10595ae90>
# 0. from-awkward-8fdcc693b11857efa42a556ced92dfbc
# 1. <dask-awkward.lib.core.ArgsKwargsPackedFunction ob-51ec44d7d3f220ecde45b63a6fd614d0
 
dak.map_partitions(fun, dak_array).dask
# HighLevelGraph with 2 layers.
# <dask.highlevelgraph.HighLevelGraph object at 0x112fde680>
# 0. from-awkward-8fdcc693b11857efa42a556ced92dfbc
# 1. <dask-awkward.lib.core.ArgsKwargsPackedFunction ob-eb6d01930cd4e0e1e190e3dda45e2f17
```

Here, one can see that the token of the second layer always changes: `7ae414b04e6de108142215f755b2eb3e` -> `51ec44d7d3f220ecde45b63a6fd614d0` -> `eb6d01930cd4e0e1e190e3dda45e2f17`. This token does not change if one directly applies `fun`, i.e. `fun(dak_array).dask`.

The reason for different tokens is that each time `fun` is wrapped inside a new instance of `ArgsKwargsPackedFunction`, and this instance is passed to the tokenizer. 

This also results in never hitting the `dak_cache`, which is likely unwanted: https://github.com/dask-contrib/dask-awkward/blob/main/src/dask_awkward/lib/core.py#L2011-L2012.

Passing the original `fun` (instead of the wrapped one) to `dask.base.tokenize` fixes this issue, but leads to a lot of failing tests (probably because the caches are hit now and that doesn't work as expected?).

How should this token generation be treated correctly @martindurant ?

I'm happy to provide a fix (given that I understand this issue here correctly), because I think that this mechanism can be used to speed up graph building in general.

---
My end goal is to make it possible to give `token=` directly to `uproot.dask` such that we can diminish the amount of layer building:

```python
# `b.dask` can reuse the layers of `a.dask` from the `dak_cache` if the `token` match:
a = uproot.dask("DY.root", token="MC")
b = uproot.dask("ttbar.root", token="MC")
```

Here, the graph would still be built per .root file but the internal layer (`dak.from_map`) is just constructed once for `a` and could then be reused from cache for `b`. All consecutive steps (other `dak.map_partitions` or `dak.mapfilter` calls) could be cached then aswell in a similar way.
In combination with [dak.mapfilter](https://github.com/dask-contrib/dask-awkward/pull/551), which allows to skip typetracing entirely when providing the correct `needs` and `meta`, one could speed up graph building significantly.

(cc @lgray and @jpivarski because this goes a bit in the direction of duplicating graphs for datasets that should be treated in the same way, but rather through a caching mechanism that exists already in dask-awkward.)

pfackeldey:
Fixes #553.

Now, the token is generated deterministically and thus the `dak_cache` is hit correctly for multiple repetitive invocations of `dak.map_partitions` with the same arguments.

Unfortunately I couldn't use `__dask_tokenize__` because one needs to repack the `args` and `kwargs` correctly.

Not sure how much of a performance gain there is, would be nice to measure it somehow...

### DIFF
--- a/src/dask_awkward/lib/core.py
+++ b/src/dask_awkward/lib/core.py
@@ -1956,7 +1956,9 @@ class ArgsKwargsPackedFunction:
         self.arg_lens_for_repackers = arg_lens_for_repackers
 
     def __repr__(self):
-        return "repacked-" + repr(self.fn).replace("<", "").replace(">", "").replace("function ", "").replace("built-in ", "")
+        return "repacked-" + repr(self.fn).replace("<", "").replace(">", "").replace(
+            "function ", ""
+        ).replace("built-in ", "")
 
     def _repack(self, *args_deps_expanded):
         """This packing function receives a list of strictly
--- a/src/dask_awkward/lib/core.py
+++ b/src/dask_awkward/lib/core.py
@@ -1966,7 +1966,15 @@ class ArgsKwargsPackedFunction:
         self.arg_lens_for_repackers = arg_lens_for_repackers
 
     def __repr__(self):
-        return f"{self.fn.__qualname__}-repacked"
+        if hasattr(self.fn, "__qualname__")
+            return f"{self.fn.__qualname__}-repacked"
+        return (
+            repr(self.fn)
+            .replace("<", "")
+            .replace(">", "")
+            .replace("function ", "")
+            .replace("built-in ", "")
+        ) + "-repacked"
 
     def _repack(self, *args_deps_expanded):
         """This packing function receives a list of strictly
--- a/src/dask_awkward/lib/core.py
+++ b/src/dask_awkward/lib/core.py
@@ -1956,19 +1956,6 @@ class ArgsKwargsPackedFunction:
         self.arg_lens_for_repackers = arg_lens_for_repackers
 
     def _repack(self, *args_deps_expanded):
-        args = []
-        len_args = 0
-        for repacker, n_args in zip(self.arg_repackers, self.arg_lens_for_repackers):
-            args.append(
-                repacker(args_deps_expanded[len_args : len_args + n_args])[0]
-                if repacker is not None
-                else args_deps_expanded[len_args]
-            )
-            len_args += n_args
-        kwargs = self.kwarg_repacker(args_deps_expanded[len_args:])[0]
-        return args, kwargs
-
-    def __call__(self, *args_deps_expanded):
         """This packing function receives a list of strictly
         ordered arguments. The first range of arguments,
         [0:sum(self.arg_lens_for_repackers)], corresponding to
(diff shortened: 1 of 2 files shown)


pfackeldey:
@lgray do you want to add something else here? I would add a test tomorrow, and then the PR is ready from my side :+1:

pfackeldey:
This one is ready from my side @martindurant and @lgray :+1:

lgray:
@pfackeldey with #558 you'll need to `map_partitions` a function that has structured inputs to get it to pack it into ArgsKwargsPackedFunction. Just want to make sure your test stays pertinent!

martindurant:
+1

lgray:
i.e. you'll have to pass a `dict` or something like that which contains dask collections.

pfackeldey:
I hope this is now correct @lgray and @martindurant

src/dask_awkward/lib/core.py
```diff
@@ -1965,7 +1965,16 @@ def __init__(self, the_fn, arg_repackers, kwarg_repacker, arg_lens_for_repackers
...
         self.arg_lens_for_repackers = arg_lens_for_repackers
 
-    def __call__(self, *args_deps_expanded):
+    def __repr__(self):
+        return (
+            repr(self.fn)
+            .replace("<", "")
+            .replace(">", "")
+            .replace("function ", "")
+            .replace("built-in ", "")
```
martindurant, line 1974: Fortunately this doesn't get called a lot - or not at all on the compute path - but it's pretty annoying! 

Does it amount to `f"{self.fn.__module__}.{self.fn.__qualname__}-repacked"` ?

lgray: ah yeah it's this but without the `__module__` component. Thanks!

lgray:
Actually @martindurant I'm going to revert that change. Dealing with all the cases is worse than the string manipulation.

martindurant:
Sorry :|

lgray:
that one test is weirdly flakey...

pfackeldey:
@lgray is this PR ready from your side? I think we can merge it, right?

lgray:
Yes, please!

PR #556, opened 2024-12-04
fix: adapt to new Task spec in dask, now used in blockwise

lgray:
Adapts one bit of dask_awkward code that makes a graph to use a task object instead.

c.f.:
https://github.com/dask/dask/pull/11568

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -37,7 +37,8 @@ classifiers = [
 ]
 dependencies = [
   "awkward >=2.5.1",
-  "dask >=2023.04.0",
+  "dask >=2024.12.0;python_version>'3.9'"
+  "dask >=2023.04.0;python_version<'3.10'",
   "cachetools",
   "typing_extensions >=4.8.0",
 ]
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -37,7 +37,7 @@ classifiers = [
 ]
 dependencies = [
   "awkward >=2.5.1",
-  "dask >=2024.12.0;python_version>'3.9'"
+  "dask >=2024.12.0;python_version>'3.9'",
   "dask >=2023.04.0;python_version<'3.10'",
   "cachetools",
   "typing_extensions >=4.8.0",
--- a/src/dask_awkward/lib/core.py
+++ b/src/dask_awkward/lib/core.py
@@ -1928,7 +1928,11 @@ def partitionwise_layer(
             pairs.extend([arg.name, "i"])
             numblocks[arg.name] = (1,)
         elif isinstance(arg, Delayed):
-            pairs.extend([arg.key, None])
+            if _dask_uses_tasks:
+                from dask._task_spec import TaskRef
+                pairs.extend([TaskRef(arg.key), None])
+            else:
+                pairs.extend([arg.key, None])
         elif is_dask_collection(arg):
             raise DaskAwkwardNotImplemented(
                 "Use of Array with other Dask collections is currently unsupported."
--- a/src/dask_awkward/lib/core.py
+++ b/src/dask_awkward/lib/core.py
@@ -1930,6 +1930,7 @@ def partitionwise_layer(
         elif isinstance(arg, Delayed):
             if _dask_uses_tasks:
                 from dask._task_spec import TaskRef
+
                 pairs.extend([TaskRef(arg.key), None])
             else:
                 pairs.extend([arg.key, None])
--- a/src/dask_awkward/lib/core.py
+++ b/src/dask_awkward/lib/core.py
@@ -21,10 +21,20 @@ import dask.config
 import numpy as np
 from awkward._do import remove_structure as ak_do_remove_structure
 from awkward.highlevel import NDArrayOperatorsMixin, _dir_pattern
-from awkward.typetracer import (MaybeNone, OneOf, TypeTracerArray,
-                                create_unknown_scalar, is_unknown_scalar)
-from dask.base import (DaskMethodsMixin, dont_optimize, is_dask_collection,
-                       tokenize, unpack_collections)
+from awkward.typetracer import (
+    MaybeNone,
+    OneOf,
+    TypeTracerArray,
+    create_unknown_scalar,
+    is_unknown_scalar,
+)
+from dask.base import (
+    DaskMethodsMixin,
+    dont_optimize,
+    is_dask_collection,
+    tokenize,
+    unpack_collections,
+)
 from dask.blockwise import BlockwiseDep
 from dask.blockwise import blockwise as dask_blockwise
 from dask.context import globalmethod
@@ -37,10 +47,15 @@ from dask.utils import funcname, is_arraylike, key_split
 
 from dask_awkward.layers import AwkwardBlockwiseLayer, AwkwardMaterializedLayer
 from dask_awkward.lib.optimize import all_optimizations
-from dask_awkward.utils import (ConcretizationTypeError,
-                                DaskAwkwardNotImplemented,
-                                IncompatiblePartitions, field_access_to_front,
-                                first, hyphenize, is_empty_slice)
+from dask_awkward.utils import (
+    ConcretizationTypeError,
+    DaskAwkwardNotImplemented,
+    IncompatiblePartitions,
+    field_access_to_front,
+    first,
+    hyphenize,
+    is_empty_slice,
+)
 
 if TYPE_CHECKING:
     from awkward.contents.content import Content
@@ -981,8 +996,10 @@ class Array(DaskMethodsMixin, NDArrayOperatorsMixin):
           (npartitions=1) is a special case of this.
         """
         from dask_awkward.layers import AwkwardMaterializedLayer
-        from dask_awkward.lib.structure import (repartition_layer,
-                                                simple_repartition_layer)
+        from dask_awkward.lib.structure import (
+            repartition_layer,
+            simple_repartition_layer,
+        )
 
         if (
             sum(
--- a/src/dask_awkward/lib/core.py
+++ b/src/dask_awkward/lib/core.py
@@ -45,7 +45,11 @@ from dask.utils import IndexCallable
 from dask.utils import OperatorMethodMixin as DaskOperatorMethodMixin
 from dask.utils import funcname, is_arraylike, key_split
 
-from dask_awkward.layers import AwkwardBlockwiseLayer, AwkwardMaterializedLayer, _dask_uses_tasks
+from dask_awkward.layers import (
+    AwkwardBlockwiseLayer,
+    AwkwardMaterializedLayer,
+    _dask_uses_tasks,
+)
 from dask_awkward.lib.optimize import all_optimizations
 from dask_awkward.utils import (
     ConcretizationTypeError,
--- a/src/dask_awkward/lib/core.py
+++ b/src/dask_awkward/lib/core.py
@@ -45,7 +45,7 @@ from dask.utils import IndexCallable
 from dask.utils import OperatorMethodMixin as DaskOperatorMethodMixin
 from dask.utils import funcname, is_arraylike, key_split
 
-from dask_awkward.layers import AwkwardBlockwiseLayer, AwkwardMaterializedLayer
+from dask_awkward.layers import AwkwardBlockwiseLayer, AwkwardMaterializedLayer, _dask_uses_tasks
 from dask_awkward.lib.optimize import all_optimizations
 from dask_awkward.utils import (
     ConcretizationTypeError,
--- a/src/dask_awkward/lib/core.py
+++ b/src/dask_awkward/lib/core.py
@@ -21,20 +21,10 @@ import dask.config
 import numpy as np
 from awkward._do import remove_structure as ak_do_remove_structure
 from awkward.highlevel import NDArrayOperatorsMixin, _dir_pattern
-from awkward.typetracer import (
-    MaybeNone,
-    OneOf,
-    TypeTracerArray,
-    create_unknown_scalar,
-    is_unknown_scalar,
-)
-from dask.base import (
-    DaskMethodsMixin,
-    dont_optimize,
-    is_dask_collection,
-    tokenize,
-    unpack_collections,
-)
+from awkward.typetracer import (MaybeNone, OneOf, TypeTracerArray,
+                                create_unknown_scalar, is_unknown_scalar)
+from dask.base import (DaskMethodsMixin, dont_optimize, is_dask_collection,
+                       tokenize, unpack_collections)
 from dask.blockwise import BlockwiseDep
 from dask.blockwise import blockwise as dask_blockwise
 from dask.context import globalmethod
@@ -47,15 +37,10 @@ from dask.utils import funcname, is_arraylike, key_split
 
 from dask_awkward.layers import AwkwardBlockwiseLayer, AwkwardMaterializedLayer
 from dask_awkward.lib.optimize import all_optimizations
-from dask_awkward.utils import (
-    ConcretizationTypeError,
-    DaskAwkwardNotImplemented,
-    IncompatiblePartitions,
-    field_access_to_front,
-    first,
-    hyphenize,
-    is_empty_slice,
-)
+from dask_awkward.utils import (ConcretizationTypeError,
+                                DaskAwkwardNotImplemented,
+                                IncompatiblePartitions, field_access_to_front,
+                                first, hyphenize, is_empty_slice)
 
 if TYPE_CHECKING:
     from awkward.contents.content import Content
@@ -996,10 +981,8 @@ class Array(DaskMethodsMixin, NDArrayOperatorsMixin):
           (npartitions=1) is a special case of this.
         """
         from dask_awkward.layers import AwkwardMaterializedLayer
-        from dask_awkward.lib.structure import (
-            repartition_layer,
-            simple_repartition_layer,
-        )
+        from dask_awkward.lib.structure import (repartition_layer,
+                                                simple_repartition_layer)
 
         if (
             sum(
--- a/src/dask_awkward/lib/core.py
+++ b/src/dask_awkward/lib/core.py
@@ -61,6 +61,9 @@ from dask_awkward.utils import (
     is_empty_slice,
 )
 
+if _dask_uses_tasks:
+    from dask._task_spec import TaskRef
+
 if TYPE_CHECKING:
     from awkward.contents.content import Content
     from awkward.forms.form import Form
@@ -1933,8 +1936,6 @@ def partitionwise_layer(
             numblocks[arg.name] = (1,)
         elif isinstance(arg, Delayed):
             if _dask_uses_tasks:
-                from dask._task_spec import TaskRef
-
                 pairs.extend([TaskRef(arg.key), None])
             else:
                 pairs.extend([arg.key, None])
--- a/src/dask_awkward/lib/optimize.py
+++ b/src/dask_awkward/lib/optimize.py
@@ -238,14 +238,23 @@ def _touch_all_data(*args, **kwargs):
 
 def _mock_output(layer):
     """Update a layer to run the _touch_all_data."""
-    assert len(layer.dsk) == 1
+    if _dask_uses_tasks:
+        new_layer = copy.deepcopy(layer)
+        task = new_layer.task.copy()
+        # replace the original function with _touch_all_data
+        # and keep the rest of the task the same
+        task.func = _touch_all_data
+        new_layer.task = task
+        return new_layer
+    else:
+        assert len(layer.dsk) == 1
 
-    new_layer = copy.deepcopy(layer)
-    mp = new_layer.dsk.copy()
-    for k in iter(mp.keys()):
-        mp[k] = (_touch_all_data,) + mp[k][1:]
-    new_layer.dsk = mp
-    return new_layer
+        new_layer = copy.deepcopy(layer)
+        mp = new_layer.dsk.copy()
+        for k in iter(mp.keys()):
+            mp[k] = (_touch_all_data,) + mp[k][1:]
+        new_layer.dsk = mp
+        return new_layer
 
 
 @no_type_check
@@ -358,10 +367,13 @@ def rewrite_layer_chains(dsk: HighLevelGraph, keys: Sequence[Key]) -> HighLevelG
                 outlayer.io_deps[k] = layer.io_deps[k]
 
             if _dask_uses_tasks:
-                from dask._task_spec import Task
+                from dask._task_spec import GraphNode, Task
 
                 func = layer.task.func
-                args = layer.task.dependencies
+                args = [
+                    arg.key if isinstance(arg, GraphNode) else arg
+                    for arg in layer.task.args
+                ]
                 # how to do this with `.substitute(...)`?
                 args2 = _recursive_replace(args, layer, parent, indices)
                 all_tasks.append(Task(chain_member, func, *args2))
@@ -408,7 +420,13 @@ def _recursive_replace(args, layer, parent, indices):
             else:
                 # arg refers to things defined in io_deps
                 indices.append(layer.indices[ind])
-                args2.append(f"__dask_blockwise__{len(indices) - 1}")
+                arg2 = f"__dask_blockwise__{len(indices) - 1}"
+                if _dask_uses_tasks:
+                    from dask._task_spec import TaskRef
+
+                    args2.append(TaskRef(arg2))
+                else:
+                    args2.append(arg2)
         elif isinstance(arg, list):
             args2.append(_recursive_replace(arg, layer, parent, indices))
         elif isinstance(arg, tuple):
--- a/src/dask_awkward/lib/optimize.py
+++ b/src/dask_awkward/lib/optimize.py
@@ -13,7 +13,11 @@ from dask.core import flatten
 from dask.highlevelgraph import HighLevelGraph
 from dask.local import get_sync
 
-from dask_awkward.layers import AwkwardBlockwiseLayer, AwkwardInputLayer
+from dask_awkward.layers import (
+    AwkwardBlockwiseLayer,
+    AwkwardInputLayer,
+    _dask_uses_tasks,
+)
 from dask_awkward.lib.utils import typetracer_nochecks
 from dask_awkward.utils import first
 
@@ -340,7 +344,10 @@ def rewrite_layer_chains(dsk: HighLevelGraph, keys: Sequence[Key]) -> HighLevelG
         deps[outkey] = deps[chain[0]]
         [deps.pop(ch) for ch in chain[:-1]]
 
-        subgraph = layer0.dsk.copy()  # mypy: ignore
+        if _dask_uses_tasks:
+            all_tasks = [layer0.task]
+        else:
+            subgraph = layer0.dsk.copy()
         indices = list(layer0.indices)
         parent = chain[0]
 
@@ -349,14 +356,27 @@ def rewrite_layer_chains(dsk: HighLevelGraph, keys: Sequence[Key]) -> HighLevelG
             layer = dsk.layers[chain_member]
             for k in layer.io_deps:  # mypy: ignore
                 outlayer.io_deps[k] = layer.io_deps[k]
-            func, *args = layer.dsk[chain_member]  # mypy: ignore
-            args2 = _recursive_replace(args, layer, parent, indices)
-            subgraph[chain_member] = (func,) + tuple(args2)
+
+            if _dask_uses_tasks:
+                from dask._task_spec import Task
+
+                func = layer.task.func
+                args = layer.task.dependencies
+                # how to do this with `.substitute(...)`?
+                args2 = _recursive_replace(args, layer, parent, indices)
+                all_tasks.append(Task(chain_member, func, *args2))
+            else:
+                func, *args = layer.dsk[chain_member]  # mypy: ignore
+                args2 = _recursive_replace(args, layer, parent, indices)
+                subgraph[chain_member] = (func,) + tuple(args2)
             parent = chain_member
         outlayer.numblocks = {
             i[0]: (numblocks,) for i in indices if i[1] is not None
         }  # mypy: ignore
-        outlayer.dsk = subgraph  # mypy: ignore
+        if _dask_uses_tasks:
+            outlayer.task = Task.fuse(*all_tasks)
+        else:
+            outlayer.dsk = subgraph  # mypy: ignore
         if hasattr(outlayer, "_dims"):
             del outlayer._dims
         outlayer.indices = tuple(  # mypy: ignore
@@ -379,7 +399,12 @@ def _recursive_replace(args, layer, parent, indices):
                 args2.append(layer.indices[ind][0])
             elif layer.indices[ind][0] == parent:
                 # arg refers to output of previous layer
-                args2.append(parent)
+                if _dask_uses_tasks:
+                    from dask._task_spec import TaskRef
+
+                    args2.append(TaskRef(parent))
+                else:
+                    args2.append(parent)
             else:
                 # arg refers to things defined in io_deps
                 indices.append(layer.indices[ind])
--- a/src/dask_awkward/lib/optimize.py
+++ b/src/dask_awkward/lib/optimize.py
@@ -21,6 +21,9 @@ from dask_awkward.layers import (
 from dask_awkward.lib.utils import typetracer_nochecks
 from dask_awkward.utils import first
 
+if _dask_uses_tasks:
+    from dask._task_spec import GraphNode, Task, TaskRef
+
 if TYPE_CHECKING:
     from awkward._nplikes.typetracer import TypeTracerReport
     from dask.typing import Key
@@ -367,8 +370,6 @@ def rewrite_layer_chains(dsk: HighLevelGraph, keys: Sequence[Key]) -> HighLevelG
                 outlayer.io_deps[k] = layer.io_deps[k]
 
             if _dask_uses_tasks:
-                from dask._task_spec import GraphNode, Task
-
                 func = layer.task.func
                 args = [
                     arg.key if isinstance(arg, GraphNode) else arg
@@ -412,8 +413,6 @@ def _recursive_replace(args, layer, parent, indices):
             elif layer.indices[ind][0] == parent:
                 # arg refers to output of previous layer
                 if _dask_uses_tasks:
-                    from dask._task_spec import TaskRef
-
                     args2.append(TaskRef(parent))
                 else:
                     args2.append(parent)
@@ -422,8 +421,6 @@ def _recursive_replace(args, layer, parent, indices):
                 indices.append(layer.indices[ind])
                 arg2 = f"__dask_blockwise__{len(indices) - 1}"
                 if _dask_uses_tasks:
-                    from dask._task_spec import TaskRef
-
                     args2.append(TaskRef(arg2))
                 else:
                     args2.append(arg2)
--- a/.github/workflows/pypi-tests.yml
+++ b/.github/workflows/pypi-tests.yml
@@ -18,7 +18,7 @@ jobs:
       fail-fast: false
       matrix:
         platform: [ubuntu-latest, macos-latest, windows-latest]
-        python-version: ["3.8","3.9", "3.10", "3.11", "3.12"]
+        python-version: ["3.8", "3.9", "3.10", "3.11", "3.12"]
     runs-on: ${{matrix.platform}}
     steps:
     - name: Checkout
--- a/.github/workflows/pypi-tests.yml
+++ b/.github/workflows/pypi-tests.yml
@@ -18,7 +18,7 @@ jobs:
       fail-fast: false
       matrix:
         platform: [ubuntu-latest, macos-latest, windows-latest]
-        python-version: ["3.9", "3.10", "3.11", "3.12"]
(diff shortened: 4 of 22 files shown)


martindurant:
I asked on the dask PR whether there are specific migration instructions associated with their change.

lgray:
@martindurant `rewrite_layer_chains` needs to be adjusted, along with various checking that's happening during the tests. The latter is all localized in one function for the most part so hopefully a reasonable fix.

There must be some way to just pop out a good old dictionary from the Task class.

lgray:
We will likely need to put a `dask >= 2024.12.0` requirement on the next release, and a `dask < 2024.12.0` for python < 3.10.

pfackeldey:
If I'm understanding this correctly a `Task` allows to [`.substitute`](https://github.com/dask/dask/blob/main/dask/_task_spec.py#L754-L784) itself by another one? This could be useful for graph cloning and then replacing the IO layer - or am I misunderstanding this?

martindurant:
> Task allows to .substitute itself

Some docs on that method would be very useful.

martindurant:
> appease mypy's insatiable lust for perfect correctness

:)

martindurant:
> rewrite_layer_chains needs to be adjusted, along with various checking that's happening during the tests

Is this something I'll need to get on?

lgray:
@martindurant seems like it - but it also looks like it'll get significantly more easy to read using the new fuse/substitute interfaces in Task.

pfackeldey:
> > rewrite_layer_chains needs to be adjusted, along with various checking that's happening during the tests
> 
> Is this something I'll need to get on?

I'd be interested to get familiar with this new `Task` class - I think it would be beneficial to know what `substitute` allows to do. 
If you want I can try to look into updating `rewrite_layer_chains` aswell.

lgray:
probably a good first introduction to update `rewrite_layer_chains` +1 to that.

martindurant:
> I can try to look into updating rewrite_layer_chains

sure, go for it! If it's hard to understand, ask any time.

fjetter:
This is off-topic but I am confused about what `rewrite_layer_chains` is supposed to do. I haven't dug into the code but the doc string reads as if this was a reimplementation of `optimize_blockwise`. We're already fusing adjacent blockwise layers internally so I'm curious what this is doing that we are not already by default.

fjetter:
fwiw the one code change you'll need is

```diff
diff --git a/src/dask_awkward/layers/layers.py b/src/dask_awkward/layers/layers.py
index 9244144..e89401f 100644
--- a/src/dask_awkward/layers/layers.py
+++ b/src/dask_awkward/layers/layers.py
@@ -8,6 +8,7 @@ from dask.blockwise import Blockwise, BlockwiseDepDict, blockwise_token
 from dask.highlevelgraph import MaterializedLayer
 from dask.layers import DataFrameTreeReduction
 from typing_extensions import TypeAlias
+from dask._task_spec import Task, TaskRef

 from dask_awkward.utils import LazyInputsDict

@@ -163,7 +164,7 @@ class AwkwardInputLayer(AwkwardBlockwiseLayer):
         super().__init__(
             output=self.name,
             output_indices="i",
-            dsk={name: (self.io_func, blockwise_token(0))},
+            task=Task(name, self.io_func, TaskRef(blockwise_token(0))),
             indices=[(io_arg_map, "i")],
             numblocks={},
             annotations=None,
```

If I disable the `rewrite_layer_chains ` optimization, most of the test suite passes with this change. The test failures relate to specific assertions on the `dsk` attribute of the blockwise layer (which has been removed in favor of the Task class) so that needs a bit more thorough review.

With all of this, please note the underscore in the `_task_spec` module. It's not unlikely that we'll end up refactoring things around that class. The high level API, e.g. the above diff, should stay the same but if you intend to dig deeper this may become brittle down the line.

Regarding a migration guide, the challenging part of this PR will be to migrate the `rewrite_layer_chains` optimization. I don't think this will just be a syntactic rewrite as the above diff but requires some more thought. The code looks simpler than the dask internal blockwise optimize so I assume this shouldn't be too much trouble.

fjetter:
I assume you are referring to https://github.com/dask/dask/issues/9795? I believe there is a misunderstanding of some kind, either on your or on my part.

Dask has a couple of different optimizer steps (that we're actually currently overhauling partially)

- blockwise fusion / optimization
- Low level task fusion

The blockwise fusion takes consecutive Blockwise layers and is merging them into a single Blockwise layer if certain requirements are fulfilled. The graph underneath that is not materialized at this stage. 

After this step, the graph is materialized (i.e. then you'll have millions of tasks) and we are calling "low level task fusion" on that graph that is looking at chains of individual tasks. We actually removed the old low level fusion code with something simpler (and faster) recently.

It feels like `rewrite_layer_chains` and our blockwise fusion is doing the same. Or at least, it is supposed to do the same. 

(I know you pinged me repeatedly on https://github.com/dask/dask/issues/9795 but I haven't had the time to follow along back then so I'm catching up right now)

pfackeldey:
Some performance comparisons regarding `rewrite_layer_chains` and `optimize_blockwise`:

```python
# takes from https://github.com/dask/dask/issues/9795#issuecomment-1372537369

import sys
import dask.array as da

sys.setrecursionlimit(10000)

def make_graph(num_layers):
    array = da.zeros(1000, chunks=(100,))
    for i in range(num_layers):
        array = array + 1
    return array
    
arr = make_graph(1_000)


# -------------
# dask.__version__ == [2024.12.0](https://github.com/dask/dask/releases/tag/2024.12.0)
from dask.blockwise import optimize_blockwise

%timeit optimize_blockwise(arr.dask, arr.__dask_keys__())
# > 291 ms ± 2.16 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)


# -------------
# dask.__version__ == [2024.11.2](https://github.com/dask/dask/releases/tag/2024.11.2)
from dask.blockwise import optimize_blockwise

%timeit optimize_blockwise(arr.dask, arr.__dask_keys__())
# > 838 ms ± 16.3 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)


# -------------
# dask.__version__ == [2024.11.2](https://github.com/dask/dask/releases/tag/2024.11.2)
# dask-awkward.__version__ == '2024.9.1.dev9+g1d4d4e9'
from dask_awkward.lib.optimize import rewrite_layer_chains

%timeit rewrite_layer_chains(arr.dask, arr.__dask_keys__())
# > 388 μs ± 1.84 μs per loop (mean ± std. dev. of 7 runs, 1,000 loops each)
```

So it looks like there's a speedup already in `optimize_blockwise` by upgrading dask to `2024.12.0`, but `rewrite_layer_chains` is still significantly faster.


---
Oh wait, something is off with `rewrite_layer_chains`, I guess I'm using it wrongly? Looking at the output HLG no fusion happened in the case of `rewrite_layer_chains`?

```python
print(optimize_blockwise(arr.dask, arr.__dask_keys__()))
# HighLevelGraph with 1 layers.
# <dask.highlevelgraph.HighLevelGraph object at 0x10958f6a0>
# 0. add-cf3b8e445fcb2608bb3016be1ec4b299

print(rewrite_layer_chains(arr.dask, arr.__dask_keys__()))
# HighLevelGraph with 1002 layers.
# <dask.highlevelgraph.HighLevelGraph object at 0x1090bbd00>
# 0. zeros_like-f0d60d8a5800b8c89bf04b383f380015
# 1. add-41ea24f1508f8670fba2857f55292924
# 2. add-043382e00b1563572167b7cd00b6c4fa
# ...
# 999. add-93272e7661ed72d8cd3475c2834e3c84
# 1000. add-cf3b8e445fcb2608bb3016be1ec4b299
```
(Maybe you should ignore the benchmark for `rewrite_layer_chains` for now until I use it/make it work correctly...)

fjetter:
Ok, so TLDR is that if our blockwise optimizer was faster and less complex, you wouldn't have a need for your own (and possibly your own is a drop in replacement for ours). That's good enough for me now.

Regarding your migration, I suppose you can copy a couple of sections from https://github.com/dask/dask/pull/11568

I suspect your optimizer is migrated with something like

```diff
--- a/src/dask_awkward/lib/optimize.py
+++ b/src/dask_awkward/lib/optimize.py
@@ -5,7 +5,7 @@ import logging
 import warnings
 from collections.abc import Iterable, Mapping, Sequence
 from typing import TYPE_CHECKING, Any, cast, no_type_check
-
+from dask._task_spec import Task
 import dask.config
 from awkward.typetracer import touch_data
 from dask.blockwise import Blockwise, fuse_roots, optimize_blockwise
@@ -340,23 +340,19 @@ def rewrite_layer_chains(dsk: HighLevelGraph, keys: Sequence[Key]) -> HighLevelG
         deps[outkey] = deps[chain[0]]
         [deps.pop(ch) for ch in chain[:-1]]

-        subgraph = layer0.dsk.copy()  # mypy: ignore
+        all_tasks = [layer0.task]
         indices = list(layer0.indices)
-        parent = chain[0]

         outlayer.io_deps = layer0.io_deps  # mypy: ignore
         for chain_member in chain[1:]:
             layer = dsk.layers[chain_member]
             for k in layer.io_deps:  # mypy: ignore
                 outlayer.io_deps[k] = layer.io_deps[k]
-            func, *args = layer.dsk[chain_member]  # mypy: ignore
-            args2 = _recursive_replace(args, layer, parent, indices)
-            subgraph[chain_member] = (func,) + tuple(args2)
-            parent = chain_member
+            all_tasks.append(layer.task.substitute(...))
         outlayer.numblocks = {
             i[0]: (numblocks,) for i in indices if i[1] is not None
         }  # mypy: ignore
-        outlayer.dsk = subgraph  # mypy: ignore
+        outlayer.task = Task.fuse(*all_tasks)
```

with the appropriate input to `substitute`. `Task.fuse` will raise if the fusion is invalid (i.e. if the subgraph that's spanned by the task are not reducing to a single key)

Note: substitute is currently not very fast. For ~7k layers I hope this is fast enough but it's something I want to look into in the near to mid term future. Generally, I expect that the graph construction cost will go up with the `Task` class but some foo on the scheduler and during submission should go down and make up for the increased cost of construction. If you end up seeing something differently please let me know

martindurant:
@pfackeldey : rewrite chains only works on dask-awkward layers.

pfackeldey:
> @pfackeldey : rewrite chains only works on dask-awkward layers.

Yes, I realized this aswell, I've updated my comment with the benchmarks accordingly. :+1:

lgray:
@fjetter Given that there was some low hanging fruit in terms of improving computational complexity of the optimizer, there is probably more. Likely worth a look.

I also had some reasonable success using numba to optimize hotspot special cases.

lgray:
@fjetter while we can capture a small amount of your attention, and since it's somewhat pertinent as we are discussing graph optimization. We see the following behavior, calculating a single object of interest (in this case a histogram) tends to optimize very well (the bottom squares are some large constants, the middle node is processing, the last is filling histograms in a tree reduce):
image

(the boxes at the bottom are dask.delayed objects which are constant correction data and could be inlined but there's no out-of-the-box function for it for HLGs)

However, when we start plotting multiple histograms, in this case one histogram is of an input value to the quantity in the other, the optimizations seem to stop where data gets reused *:
image

In particular, here you can see "met.phi" gets sent to multiple nodes for various uses and that seems to more or less ~stop a great degree of optimization. Likewise the control region definition logic (the blobs of nodes right before the end) seem to not be collapsed as before. All in all, lose the graph layer fusion that really helps with creating tractable, scalable optimized task graphs. What would be a good strategy for making an optimizer that would be able to still contract the graph to the minimum number of nodes while still allowing data re-use?

\* here's the un-optimized task graph for your amusement:
[kelci_opt.pdf](https://github.com/user-attachments/files/18031117/kelci_opt.pdf)

fjetter:
The pictures are unfortunately not properly loading for me so I can't inspect them closely. However, it is a known problem that shared dependencies are hard to deal with. 

It's a common fallacy to think that the reuse is beneficial or even necessary. Sometimes breaking this sharing is actually beneficial for the computation because you can pipeline computations and release memory much more quickly. That's particularly true if you are computing in a memory constrained environment. If the shared dependency is very expensive to compute (and/or small), it should be cached and shared. If it is cheap to compute, just doing it many times is good. That is intuitive for most people. Data loading/production is the part that's tripping people. It is typically relatively expensive so people want to do it as little as possible. However, while it is not cheap, it is often also very large so caching is expensive (having a larger graph is also a type of cost, especially for your computation scale).

I don't know what's blocking your optimization. Shared dependencies sometimes inherently block optimization. Sometimes it's also just a bug. I don't know and can't investigate without a reproducer

lgray:
Ok, sorry that didn't render correctly. I'll see if I can get a concise reproducer with few dependencies.

lgray:
Also, thanks for your description of generally what we'd like out of optimization. It confirms my own understanding of the trade offs and goals.

pfackeldey:
I managed to update `rewrite_layer_chains` successfully as far as I could test it, however I wasn't able to make it work with `.substitute()` [here](https://github.com/dask-contrib/dask-awkward/pull/556/files#diff-1010f143e076892f5109c98cf4723866153b3837e6905ca1896467efeaef7a03R360-R367); maybe @fjetter could give me some hints/tipps how to update it correctly?

I'll look into the output mocking next, which needs to be updated aswell.

martindurant:
> I'll look into the output mocking 

This is an "optimization v1" feature. I wonder is v2 needs any work at all - but since we've not been pushing on that, maybe not worth checking right now.

lgray:
Nice, down to one failure :-)

pfackeldey:
I think now there's only one test left that needs to be fixed: `tests/test_core.py::test_partitionwise_op_with_delayed`.
The problem here is that the new dask treats the key of the delayed as a `DataNode` which is passed down to the actual awkward operation. This results in something like: `ak.Array([....]) * DataNode(...).key` where `DataNode(...).key` is something like `a_delayed_array-da5d2abd-ca30-4227-9243-dd7b7d9ba5df`.
If someone has an idea how to solve this, I'd appreciate some tipps as I'm not really understanding how this happens...

douglasdavis:
Seems like it may be a hidden performance issue now turned bug when mixing `dak.Array` with `dask.delayed.Delayed`. If I start inspecting the arguments to that operator.mul call via a higher level function, we see in the typetracer only graph computation that we have a materialized array representing the thing that is delayed. 

```python
def mult_func(a, b):
    breakpoint()
    return operator.mul(a, b)

def test_partitionwise_op_with_delayed():
    array = ak.Array([[1, 2, 3], [4], [5, 6, 7], [8]])
    dak_array = dak.from_awkward(array, npartitions=2)
    result = map_partitions(
        mult_func,
        dak_array,
        a_delayed_array(),
        meta=dak_array._meta,
        output_divisions=1,
    )

    ...
```

`dask==2024.10.0`:
```
> /Users/ddavis/software/repos/dask-awkward/tests/test_core.py(929)mult_func()
-> breakpoint()
(Pdb) a
a = <Array-typetracer [...] type='## * var * int64'>
b = <Array [2, 4] type='2 * int64'>      # <--- the delayed thing!
```
`dask==2024.12.0`:
```
> /Users/ddavis/software/repos/dask-awkward/tests/test_core.py(929)mult_func()
-> breakpoint()
(Pdb) a
a = <Array-typetracer [...] type='## * var * int64'>
b = 'a_delayed_array-69f75d8d-9f27-454b-83f7-cbda699c7f12'
```

The fact that the older version of dask ends up yielding a concrete awkward array in the typetracer only graph is a red flag. I suspect that the compute that occurs in the optimization step is doing a "real" compute. If that delayed were expensive then this would be a bad optimization scenario. Seems like the issue may be more serious than whatever evolution has happened in Dask, and perhaps new Dask has exposed the problem. In the past mocks were introduced for some special cases. Perhaps this is another place where a mock is needed to represent the Delayed layer.

lgray:
@douglasdavis - thanks for taking a look, super useful. 

My worry about this is when the delayed object is a function, then it gets rather complicated.

I would argue, however, a concrete awkward array in this case is expected since Delayed doesn't know anything about typetracers and the computation of the graph for data-touching is a "real compute". So this really all checks out.

I think we even had a conversation about this since keeping the optimization time down is critical for coffea.

So I guess we're not understanding how to embed things in the graph correctly? Or something fundamental changed in delayed?

lgray:
Ah! if I turn all optimization off it still yields the task key in the function rather than the value (which is should eventually).

So I would say something else is up.

@douglasdavis @pfackeldey

pfackeldey:
I'm still in the process of understanding _when_ it should materialize, because in principle we need it already during the typetracing pass to trace the function with arrays. But then: what's the point of having a delayed that yields a concrete Array anyway if it has to evaluate already in the typetracing pass?
At that point one can directly pass the materialized array to `map_partitions` aswell?

douglasdavis:
Lindsey Gray ***@***.***> writes:

> @douglasdavis - thanks for taking a look, super useful.
>
> My worry about this is when the delayed object is a function, then it gets rather complicated.

Yes of course. The idea is certainly to be able to do "anything"
represented by a Delayed object

> I would argue, however, a concrete awkward array in this case is
> expected since Delayed doesn't know anything about typetracers and the
> computation of the graph for data-touching is a "real compute". So
> this really all checks out.

I agree a concrete array makes sense for the test of interest here. But
like you said above, we want to have maximal flexibility with Delayed--
and I don't think we want to be computing that "anything" if it's some
expensive function.

> I think we even had a conversation about this since keeping the
> optimization time down is critical for coffea, but it's been quite a
> long time.
>
> So I guess we're not understanding how to embed things in the graph
> correctly? Or something fundamental changed in delayed?

Possibly a two part problem? The simpler one being that something
fundamental changed in dask.delayed (without digging into this more..
I'm assuming Dask's evolution to using dedicated Tasks rather than
tuples maybe led to some internal changes there?). I think it's possible
that change has uncovered an issue in dask-awkward w.r.t. performing
dask-awkward's optimizations for graphs that combine Delayed with
dak.Array. It appears the optimization works when the result of the
delayed object is a concrete awkward Array (as proven by the originally
passing test). The optimization graph is actually getting that
concrete array to materialize during its execution, and the _typetracer_
array (from the dak.Array) that exists during the compute of
optimization-time-only graph is perfectly fine doing binary operations
with that concrete array (e.g. `a * b` is totally fine when `a` has
typetracer backend while `b` has numpy backend).

Unfortunately its impossible to know exactly what kind of object will
result from an operation between an arbitrary Delayed and a dak.Array,
and the typetracer pass is completely dependent on being able to know
what the result would be. It seems like we somehow need to have metadata
attached to a Delayed

lgray:
delayed < - > dask-awkward interactions really only happen in HEP when someone is passing a big correction into the task graph where we use delayed to wrap the function.

So I'd argue it *should* be materialized because in such a case you'd need to pass the typetracer to the function in any case.

I think this is just an essential limitation of "going across the fence" between dak and specifically delayed since you don't know what you're getting when you materialize it. Otherwise it significantly complicates the delayed interface of other packages.

Though maybe it is a wrong choice to do so?

lgray:
Let's try to first get the status quo re-established since that was at least reasonable behavior. If we can do that we'll have more understanding of what's going wrong, as well.

douglasdavis:
Lindsey Gray ***@***.***> writes:

> delayed < - > dask-awkward interactions really only happen in HEP when
> someone is passing a big correction into the task graph where we use
> delayed to wrap the function.
>
> So I'd argue it *should* be materialized because in such a case you'd
> need to pass the typetracer to the function in any case.

Yeah the more I think about it the more it makes sense to have whatever
is represented by the Delayed go ahead and "exist" when combined with
the typetracer during dak's optimization pass. The (potentially
terrible) side effect here is that the Delayed thing gets computed twice
(once for optimization pass, and again for the real compute).

> I think this is just an essential limitation of "going across the
> fence" between dak and specifically delayed since you don't know what
> you're getting when you materialize it. Otherwise it significantly
> complicates the delayed interface of other packages.
>
> Though maybe it is a wrong choice to do so?

Maybe "questionable" instead of "wrong" :P. certainly useful feature,
but a complex to get it right for all cases. When I first starting
working on Delayed <--> dak stuff, I took inspiration from dask.array
and dask.dataframe; for going between the different collections one
needed to go ahead and at least run the optimizations for the "original"
collection before shoving the (now optimized graph) into the "new"
collection (all the to_delayed functions have an optimize_graph argument
that defaults to true). So optimizations have always been something to
explicitly worry about in this area. Conversions are one thing (I'd say
simpler), but combined operations is another. I wonder if perhaps the
new expression stuff solves that complexity. All that said I agree
getting the status quo working again is absolutely highest priority.

douglasdavis:
Started to poke around the new Task stuff and found

https://github.com/dask/dask/blob/02185bedf2e6aba70ac2d3987c4f739998da835e/dask/blockwise.py#L388-L404

Looks like

https://github.com/dask-contrib/dask-awkward/blob/1d4d4e9afd3c7c68debc762297b2ff75b6103828/src/dask_awkward/lib/core.py#L1930-L1931

needs to be wrapped in `dask._task_spec.Alias`. I blindly tried it:

```diff
diff --git a/src/dask_awkward/lib/core.py b/src/dask_awkward/lib/core.py
index d7c1a4e..4c8cd45 100644
--- a/src/dask_awkward/lib/core.py
+++ b/src/dask_awkward/lib/core.py
@@ -1910,6 +1910,7 @@ def partitionwise_layer(
         The Dask HighLevelGraph Blockwise layer.
 
     """
+    from dask._task_spec import Alias
     pairs: list[Any] = []
     numblocks: dict[str, tuple[int, ...]] = {}
     for arg in args:
@@ -1928,7 +1929,7 @@ def partitionwise_layer(
             pairs.extend([arg.name, "i"])
             numblocks[arg.name] = (1,)
         elif isinstance(arg, Delayed):
-            pairs.extend([arg.key, None])
+            pairs.extend([Alias(arg.key), None])
         elif is_dask_collection(arg):
             raise DaskAwkwardNotImplemented(
                 "Use of Array with other Dask collections is currently unsupported."
```

and got a different error with default compute (e.g. `a.compute()`), but now the no-optimize case works (`a.compute(optimize_graph=False)`), so that's some positive progress.

Need to look more into how the Alias impacts the typetracer graph pass

```python
In [1]: import dask_awkward as dak
   ...: from dask.delayed import delayed
   ...: import awkward as ak
   ...:
   ...:

In [2]: x = ak.Array([[1,2,3], [4]])

In [3]: a = dak.from_awkward(x, npartitions=2)

In [4]: @delayed
   ...: def one():
   ...:     return 1
   ...:

In [5]: import operator

In [6]: b = a.map_partitions(operator.mul, one(), meta=a._meta)

In [7]: b.compute(optimize_graph=False)
Out[7]: <Array [[1, 2, 3], [4]] type='2 * var * int64'>

In [8]: b.compute()
---------------------------------------------------------------------------
ValueError                                Traceback (most recent call last)
Cell In[8], line 1
----> 1 b.compute()

File ~/software/repos/dask-awkward/.venv/lib/python3.13/site-packages/dask/base.py:372, in DaskMethodsMixin.compute(self, **kwargs)
    348 def compute(self, **kwargs):
    349     """Compute this dask collection
    350
    351     This turns a lazy Dask collection into its in-memory equivalent.
   (...)
    370     dask.compute
    371     """
--> 372     (result,) = compute(self, traverse=False, **kwargs)
    373     return result

File ~/software/repos/dask-awkward/.venv/lib/python3.13/site-packages/dask/base.py:660, in compute(traverse, optimize_graph, scheduler, get, *args, **kwargs)
    657     postcomputes.append(x.__dask_postcompute__())
    659 with shorten_traceback():
--> 660     results = schedule(dsk, keys, **kwargs)
    662 return repack([f(r, *a) for r, (f, a) in zip(results, postcomputes)])

File ~/software/repos/dask-awkward/.venv/lib/python3.13/site-packages/dask/local.py:455, in get_async(submit, num_workers, dsk, result, cache, get_id, rerun_exceptions_locally, pack_exception, raise_exception, callbacks, dumps, loads, chunksize, **kwargs)
    452     rerun_exceptions_locally = config.get("rerun_exceptions_locally", False)
    454 if state["waiting"] and not state["ready"]:
--> 455     raise ValueError("Found no accessible jobs in dask")
    457 def fire_tasks(chunksize):
    458     """Fire off a task to the thread pool"""

ValueError: Found no accessible jobs in dask
```

lgray:
@pfackeldey @douglasdavis @fjetter 
After some further poking around it is specifically running the `optimize_blockwise` pass that causes the graph to throw `ValueError: Found no accessible jobs in dask`.

If I merely comment out https://github.com/dask-contrib/dask-awkward/blob/main/src/dask_awkward/lib/optimize.py#L52 (i.e. optimize_blockwise!!)

HLG with optimize.py#52 commented out:
```
un-optimized: HighLevelGraph with 3 layers.
<dask.highlevelgraph.HighLevelGraph object at 0x115839310>
 0. a_delayed_array-07e0822d-9a8c-4164-a4b5-26570955e35e
 1. from-awkward-85ae75dd547aa970cf45ae978772e6e4
 2. <dask-awkward.lib.core.ArgsKwargsPackedFunction ob-92230ff6d385deae6aca0cffcabdad9f
 
optimized:    HighLevelGraph with 1 layers.
<dask.highlevelgraph.HighLevelGraph object at 0x117b7d430>
 0. <dask-awkward.lib.core.ArgsKwargsPackedFunction ob-92230ff6d385deae6aca0cffcabdad9f
```

with L52 restored:
```
un-optimized: HighLevelGraph with 3 layers.
<dask.highlevelgraph.HighLevelGraph object at 0x1089cc980>
 0. a_delayed_array-0f23670a-dae0-49de-b285-3732450abc70
 1. from-awkward-db6b93afdbe5c3bf69e753378b7fd590
 2. <dask-awkward.lib.core.ArgsKwargsPackedFunction ob-25fddb97bdfc2ebcd68404195ebb0836
 
optimized:    HighLevelGraph with 1 layers.
<dask.highlevelgraph.HighLevelGraph object at 0x108fe31d0>
 0. <dask-awkward.lib.core.ArgsKwargsPackedFunction ob-25fddb97bdfc2ebcd68404195ebb0836
```

The test passes and at the highest level appears to be producing the same HLG structure, I will dig into it a bit more to try to understand what's packed up differently.

lgray:
With `optimize_blockwise` commented out the resulting materialized layer contains this:

```
[(('<dask-awkward.lib.core.ArgsKwargsPackedFunction ob-6c779eab894c25ed7b3f751615b94fe3',
   0),
  <Task ('<dask-awkward.lib.core.ArgsKwargsPackedFunction ob-6c779eab894c25ed7b3f751615b94fe3', 0) <dask_awkward.lib.core.ArgsKwargsPackedFunction ob(...)>),
 (('<dask-awkward.lib.core.ArgsKwargsPackedFunction ob-6c779eab894c25ed7b3f751615b94fe3',
   1),
  <Task ('<dask-awkward.lib.core.ArgsKwargsPackedFunction ob-6c779eab894c25ed7b3f751615b94fe3', 1) <dask_awkward.lib.core.ArgsKwargsPackedFunction ob(...)>),
 (('from-awkward-d54829cedb4c32f715e92986dfca221f', 0),
  <Task ('from-awkward-d54829cedb4c32f715e92986dfca221f', 0) <dask_awkward.lib.io.io.FromAwkwardFn object at 0x(...)>),
 ('a_delayed_array-1abece2e-5d98-44eb-abcb-8ea21563a7ac',
  (<function __main__.a_delayed_array()>,)),
 (('from-awkward-d54829cedb4c32f715e92986dfca221f', 1),
  <Task ('from-awkward-d54829cedb4c32f715e92986dfca221f', 1) <dask_awkward.lib.io.io.FromAwkwardFn object at 0x(...)>)]
```

where with `optimize_blockwise` the resulting layer contains:
```
[(('<dask-awkward.lib.core.ArgsKwargsPackedFunction ob-1beb66991b79fa2f49330ef901027c85',
   0),
  <Task ('<dask-awkward.lib.core.ArgsKwargsPackedFunction ob-1beb66991b79fa2f49330ef901027c85', 0) _execute_subgraph(...)>),
 (('<dask-awkward.lib.core.ArgsKwargsPackedFunction ob-1beb66991b79fa2f49330ef901027c85',
   1),
  <Task ('<dask-awkward.lib.core.ArgsKwargsPackedFunction ob-1beb66991b79fa2f49330ef901027c85', 1) _execute_subgraph(...)>)]
```

if I then dig into the `Task` and look at its dependencies I get the set of keys:
```
frozenset({'a_delayed_array-ffab6114-4add-4b7c-9004-fd3770ba9771'})
```

and `a_delayed_array` is definitely not in the graph!

So, how to we keep the delayed node in the graph?

lgray:
With dask 2024.11.x the results are as follows:

no `optimize_blockwise`:
```
HighLevelGraph with 1 layers.
<dask.highlevelgraph.HighLevelGraph object at 0x10af66ba0>
 0. repacked-instrumented-mul at 0x1093cc540-535da6829bbcce92a42b85b9be259a05
```
and that one layer contains:
```
[(('repacked-instrumented-mul at 0x1093cc540-535da6829bbcce92a42b85b9be259a05',
   1),
  ('from-awkward-repacked-instrumented-mul at 0x1093cc540-535da6829bbcce92a42b85b9be259a05',
   1)),
 (('repacked-instrumented-mul at 0x1093cc540-535da6829bbcce92a42b85b9be259a05',
   0),
  ('from-awkward-repacked-instrumented-mul at 0x1093cc540-535da6829bbcce92a42b85b9be259a05',
   0)),
 (('from-awkward-repacked-instrumented-mul at 0x1093cc540-535da6829bbcce92a42b85b9be259a05',
   1),
  (subgraph_callable-fc246caf28f6aec6243873f1810cbd77,
   (subgraph_callable-5f18606dd7d6041a38bfe013d4e97716, (2, 4)),
   'a_delayed_array-cc51865c-c9b1-4111-8e9f-98f655e1ced9')),
 (('from-awkward-repacked-instrumented-mul at 0x1093cc540-535da6829bbcce92a42b85b9be259a05',
   0),
  (subgraph_callable-fc246caf28f6aec6243873f1810cbd77,
   (subgraph_callable-5f18606dd7d6041a38bfe013d4e97716, (0, 2)),
   'a_delayed_array-cc51865c-c9b1-4111-8e9f-98f655e1ced9')),
 ('a_delayed_array-cc51865c-c9b1-4111-8e9f-98f655e1ced9',
  (<function __main__.a_delayed_array()>,))]
```

It executes as expected.

*with* `optimize_blockwise` turned on the resulting graph is:
```
HighLevelGraph with 2 layers.
<dask.highlevelgraph.HighLevelGraph object at 0x1053fc4d0>
 0. a_delayed_array-aa8cfacb-8634-4dc5-a65d-4a19fd4bd6e2
 1. repacked-instrumented-mul at 0x10412e520-7ed96731e732a0de63e1eff62ca2e023
```
and the last layer contains:
```
[(subgraph_callable-19a20e3065e8a2b3047767df7a8e80f3,
  'a_delayed_array-aa8cfacb-8634-4dc5-a65d-4a19fd4bd6e2',
  (0, 2)),
 (subgraph_callable-19a20e3065e8a2b3047767df7a8e80f3,
  'a_delayed_array-aa8cfacb-8634-4dc5-a65d-4a19fd4bd6e2',
  (2, 4))]
```

So the old `optimize_blockwise` is de-inlining the delayed array into its own node and the new one is completely removing it? Though I think the errant behavior may be in `rewrite_layer_chains` which shouldn't be touching the Delayed layer in the first place?

@martindurant thoughts?

lgray:
Ah, my bad - it's after the cull step:

```
after fuse_roots HighLevelGraph with 2 layers.
<dask.highlevelgraph.HighLevelGraph object at 0x10ee260f0>
 0. a_delayed_array-099cec65-ea25-4b10-8183-48aa676a54b3
 1. repacked-instrumented-mul at 0x10cb6d3a0-070fec221c3aceaa77bd0b44557dd84c

after cull HighLevelGraph with 1 layers.
<dask.highlevelgraph.HighLevelGraph object at 0x10cb098e0>
 0. repacked-instrumented-mul at 0x10cb6d3a0-070fec221c3aceaa77bd0b44557dd84c
```

So somehow we're not retaining the information in the graph that marks the delayed layer as necessary.

lgray:
Indeed, if I skip the cull step everything computes as expected.

martindurant:
> if I skip the cull step everything computes as expected

So the graph *is* right, the dependency must be there, else it wouldn't compute - but cull is making some other assumption that we don't meet.

lgray:
Indeed, something has definitely changed:
image

Even though the `blockwise_optimize`d layers claim the `Delayed` as a dependency (!!) cull drops the `Delayed`.

lgray:
I have found the source of the difference. 

After https://github.com/dask/dask/pull/11568 the delayed array no longer shows up as a constant dependency in the task graph coming from this loop:
https://github.com/dask/dask/blob/main/dask/blockwise.py#L674-L684

I checked it's not another kind of dep, even in old dask the delayed-wrapped array is only a constant.

If I add code to correctly deal with `Alias`es to https://github.com/dask/dask/blob/main/dask/blockwise.py#L674-L684, i.e.:
```python3
...
            if isinstance(arg, Alias):
                arg = arg.target.key
...
```

Then everything works as expected again.

lgray:
@fjetter is this skipping of Aliases as constant dependencies of blockwise graphs desired or expected?

lgray:
I also find that if I use `TaskRef` instead of `Alias` for our `Delayed` object everything executes correctly without issue.

Understanding of what is correct would be appreciated. For the time being I will change the dask_awkward code to use a `TaskRef` as opposed to an `Alias` to see what else breaks in our tests.

lgray:
OK - only failing tests are over in uproot where we'll have to patch things up to deal with Tasks there as well!

@jpivarski

lgray:
However, we should settle these correct usage issues and get bugfixes in the right places if they are necessary.

We then wait on further input and guidance from @fjetter as to correct/expected usage.

lgray:
It's also a bit weird that the `GraphNode` base class's `.ref()` function returns an `Alias`. Hmm.

lgray:
@fjetter when you have time, we would appreciate your commentary so that we can resolve this.

lgray:
@fjetter just a ping

martindurant:
Thanks for all your effort here, @lgray . I hope @fjetter can OK the code now.

pyproject.toml
```diff
@@ -37,7 +37,8 @@ classifiers = [
 ]
 dependencies = [
   "awkward >=2.5.1",
-  "dask >=2023.04.0",
+  "dask >=2024.12.0;python_version>'3.9'",
+  "dask >=2023.04.0;python_version<'3.10'",
```
martindurant, line 41: In live discussions, we were tending towards dropping backward compatibility here, which means dropping py3.9 support (which dask and numpy already have). Users of py3.9 will not have hit the original problem, since the new dask was not released for them.

This would also save about half the lOC in this PR.

lgray:
@fjetter are you available to discuss this? Thanks!

src/dask_awkward/lib/core.py
```diff
@@ -1928,7 +1935,10 @@ def partitionwise_layer(
             pairs.extend([arg.name, "i"])
             numblocks[arg.name] = (1,)
         elif isinstance(arg, Delayed):
-            pairs.extend([arg.key, None])
+            if _dask_uses_tasks:
+                pairs.extend([TaskRef(arg.key), None])
```
fjetter, line 1939: yes, that's correct 👍

src/dask_awkward/lib/optimize.py
```diff
@@ -234,14 +241,23 @@ def _touch_all_data(*args, **kwargs):
 
 def _mock_output(layer):
     """Update a layer to run the _touch_all_data."""
-    assert len(layer.dsk) == 1
+    if _dask_uses_tasks:
+        new_layer = copy.deepcopy(layer)
+        task = new_layer.task.copy()
```
fjetter, line 246: My guess is that the task specific copy is not required after the deepcopy. I was already contemplating whether we should get rid of `copy` (because it is difficult to maintain / would require subclasses to overwrite it and we might want to make use of subclassing)

src/dask_awkward/lib/optimize.py
```diff
@@ -349,14 +368,28 @@ def rewrite_layer_chains(dsk: HighLevelGraph, keys: Sequence[Key]) -> HighLevelG
...
-            args2 = _recursive_replace(args, layer, parent, indices)
-            subgraph[chain_member] = (func,) + tuple(args2)
+
+            if _dask_uses_tasks:
+                func = layer.task.func
+                args = [
+                    arg.key if isinstance(arg, GraphNode) else arg
+                    for arg in layer.task.args
+                ]
+                # how to do this with `.substitute(...)`?
```
fjetter, line 378: is this still an open question?

fjetter (approved):
Looks good

src/dask_awkward/lib/optimize.py
pfackeldey: Yes, I was unsure how do implement this with `.substitute()`. I used our internal function instead, but it would be nice to use `.substitute` if that does the same thing.

It's not a show stopper right now though.

lgray:
@fjetter Just for reference, the source of confusion about `Alias` vs. `TaskRef` w.r.t. constant delayed objects comes from here: https://github.com/dask/dask/blob/main/dask/blockwise.py#L388-L404

I realize it's a sharp edge but since you said TaskRef is correct here, it would be good to correct the docs for posterity.

lgray:
@martindurant @pfackeldey I would say we merge as-is, and deal with the backwards compat stuff later. We should get uproot and coffea passing again with some priority...

lgray:
uproot was an easy fix https://github.com/scikit-hep/uproot5/pull/1352

PR #558, opened 2024-12-09
perf: go direct to _map_partitions in functions that receive already flat deps

lgray:
Justification: if the dependencies are already flat there's no need to unpack them.

### DIFF
--- a/src/dask_awkward/lib/core.py
+++ b/src/dask_awkward/lib/core.py
@@ -1302,7 +1302,7 @@ class Array(DaskMethodsMixin, NDArrayOperatorsMixin):
             else:
                 m = to_meta([where])[0]
                 meta = self._meta[m]
-        return _map_partitions(
+        return (
             operator.getitem,
             self,
             where,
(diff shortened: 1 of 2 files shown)


martindurant:
I suppose we need a dask version pin or wait for the Task stuff?

lgray:
I thought of something even better for this.

I think I just want to change `map_partitions` to check for nested collections (compare `traverse=True` to `traverse=False` deps) and if they're the same just go directly to `_map_partitions`. It'll save a lot of wrapping.

lgray:
cool old tests pass. I'll devise a test to make it it is doing what I expect.

lgray:
so this is ready to go except for the fact it uncovered a bug in fuse linear chains. it does not respect kwargs for the older dask versions.

martindurant:
> except for the fact it uncovered a bug in fuse linear chains

Make that a separate issue and merge this?

lgray:
I'd have to back out the change that allows non-collection kwargs to pass through if I didn't fix that issue.

lgray:
But I do agree it's probably better to factor.

lgray:
@pfackeldey we should also merge this one.

PR #562, opened 2024-12-17
fix: deal with kwargs when using _task_spec

### LINKED ISSUE #561: rewrite_layer_chains does not respect kwargs
This is only exposed in #558 since otherwise everything gets wrapped in a ArgsKwargsPackedFunction which only ever has `args`, once we stop doing that (because it is expensive when called many times) unless it is really needed then it becomes apparent that only `args` are forwarded inside `rewrite_layer_chains`.

lgray:
Fixes #561 

Only for dask >= 2024.12.0 now.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -71,7 +71,7 @@ test = [
   "distributed",
   "hist",
   "pandas",
-  "pytest >=6.0,<8",
+  "pytest >=6.0",
   "pytest-cov >=3.0.0",
   "requests",
   "uproot >=5.1.0",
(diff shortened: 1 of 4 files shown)


martindurant:
OK, this looks right, but there should be an explicit test

lgray:
Looks like #558 also requires this one since this bug was found there. I'll write a test soon.

lgray:
@martindurant So I've introduced a test but this test will not actually function until #561 is rebased on top of this PR.

lgray:
It's otherwise not really easy to test what this is doing because in this version of `_map_partitions` cannot receive kwargs in the first place!

lgray:
Test failures seem unrelated...

lgray:
I'm going to merge this and then make the test specific to kwargs over in #558

PR #563, opened 2024-12-17
fix: relax dask version requirement and use a better check for Task

lgray:
@martindurant we got pretty... erm direct... feedback that backwards compatibility would be appreciated.

Also the current check was not working with very recent, but older, dask that has `Task` but it isn't used in blockwise.

### DIFF
--- a/src/dask_awkward/layers/layers.py
+++ b/src/dask_awkward/layers/layers.py
@@ -6,14 +6,14 @@ from typing import TYPE_CHECKING, Any, Literal, Protocol, TypeVar, Union, cast
 
 import dask
 
-_dask_uses_tasks = hasattr(dask, "_task_spec")
-
 from dask.blockwise import Blockwise, BlockwiseDepDict, blockwise_token
 from dask.highlevelgraph import MaterializedLayer
 from dask.layers import DataFrameTreeReduction
 from typing_extensions import TypeAlias
 
 from dask_awkward.utils import LazyInputsDict
+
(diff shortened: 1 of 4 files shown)


src/dask_awkward/layers/layers.py
```diff
@@ -5,16 +5,15 @@
...
-_dask_uses_tasks = hasattr(dask, "_task_spec")
-
 from dask.blockwise import Blockwise, BlockwiseDepDict, blockwise_token
 from dask.highlevelgraph import MaterializedLayer
 from dask.layers import DataFrameTreeReduction
 from typing_extensions import TypeAlias
 
 from dask_awkward.utils import LazyInputsDict
 
+_dask_uses_tasks = hasattr(dask, "_task_spec") and hasattr(dask.blockwise, "Task")
```
martindurant, line 15: The second check here is probably enough? It's the change in blockwise specifically that broke things.

martindurant:
We are accepting more future work by allowing for earlier dask. That must be recognised.

lgray:
It's less future work w.r.t. people managing clusters and images. There are all sorts of weird regressions from dask version to dask version, their pinning times are much longer than ours.

lgray:
@martindurant changing things so that we avoid touching the private module `_task_spec` and grab Task/TaskRef/GraphNode from blockwise. The latter is far less likely to disappear or move, and the naming of the classes is probably pretty stable.

lgray:
any further commentary on this one?

martindurant (approved):

PR #565, opened 2025-01-14
feat: add possibility to manually perform the column projection

### LINKED ISSUE #559: Expose column optimization more explicitly?
Hi,

it would be nice if there's a way to use the column optimization without having to "buy into" building (large) compute graphs. One thing that could be useful is to expose an API for the column optimization specifically, enabled through a new namespace in dask-awkward called `manual` (because one is doing the column optimization now _manually_ instead of _automatically_ as part of the usual dask graph optimization). 

Consider the following example:

```python
import coffea
import awkward as ak
import dask_awkward as dak

# create a lazy dask-awkward array
events = coffea.NanoEventsFactory.from_root(...).events()

# define my analysis with eager awkward-array
def my_analysis(events: ak.Array) -> ak.Array:
  ...
  
# explicitly typetrace `my_analysis` to infer the required columns,
# this is not requiring `my_analysis` to be part of a dask graph!
tracer = dak.manual.to_tracer(events)
out_tracer, touched_columns = dak.manual.typetrace(my_analysis, tracer)

# use `touched_columns` to update the IO layer to only load these columns,
# this function looks for the `AwkwardInputLayer` and projects its `io_func`
events = dak.manual.project_columns(events, touched_columns)
```

Why is this useful? - One can now continue with a _non_-dask compute model but still benefit from dask-awkward's partitioning and column optimization, e.g.:

```python
# run analysis per partition only with the needed columns.
# this loads only `touched_columns` for `chunk` because 
# of the previous explicit `project_columns` call
for i in range(events.npartitions):
  chunk = events.partition[i].compute()
  analysed_events = my_analysis(chunk)
```

This pattern allows to run this loop locally e.g. with multiprocessing, on batch systems as classic batch jobs, or you can fall back to continue working with `dask-awkward` on `events` (the normal use-case) with e.g. `map_partitions`/`mapfilter`/`dak`-operations. 

Some additional benefits would be that tracing failures might be easier to debug, because there's a simple way to rerun only the tracing step: `dak.manual.typetrace(my_analysis, tracer)`. In addition, one would be able to reuse `touched_columns` of one tracing step for other input datasets (where appropriate of course), inspect the `touched_columns` (I suppose there's already `dak.report_necessary_columns` for this), or even manipulate `touched_columns` as needed.

While there's still the standard way to do things, i.e., use `dask-awkward` operations and it's optimization procedures, the `manual` namespace would allow people to benefit from `dask-awkward`'s IO features (column optimization + partitioning) without having to build dask graphs (except for the one AwkwardInputLayer) for the computation of `my_analysis`. 

What do you think about this?

pfackeldey:
This PR adds a new function (`dak.manual.optimize_columns`) that allows to perform the column projection by hand given a set of columns (or keys) in the same form that `dak.inspect.report_necessary_columns` returns. (Closes #559)

### Why is this useful?
- This PR allows to run `dak.inspect.report_necessary_columns` once for one dataset and then reuse its output to manually project these columns for other datasets. 
- This PR allows the user to manipulate the set of columns, see the test `tests/test_manual.py` as an example. 

### Note
Currently, `dak.manual.optimize_columns` will disable the possibility to project theAwkwardInputLayers again. This is because otherwise there might be unexpected results when running the manual optimization and then the automatic one that's enabled by default. 

Looking forward to your feedback!

PS: If this gets accepted, I'll open another PR in uproot that allows to perform the manual column optimization aswell.

### DIFF
--- a/src/dask_awkward/layers/layers.py
+++ b/src/dask_awkward/layers/layers.py
@@ -79,6 +79,12 @@ class ImplementsProjection(ImplementsMocking, Protocol[T]):
 
     def project(self, report: TypeTracerReport, state: T) -> ImplementsIOFunction: ...
 
+    # `project_manually` is typically an alias to `project_columns`. Some IO functions
+    # might implement this method with a different name, because their respective file format
+    # uses different conventions. An example is ROOT, where the columns are called "keys".
+    # In this case, the method should be aliased to `project_keys`.
+    def project_manually(self, columns: frozenset[str]) -> ImplementsIOFunction: ...
+
 
 class ImplementsNecessaryColumns(ImplementsProjection[T], Protocol):
     def necessary_columns(
@@ -153,6 +159,7 @@ class AwkwardInputLayer(AwkwardBlockwiseLayer):
         produces_tasks: bool = False,
         creation_info: dict | None = None,
         annotations: Mapping[str, Any] | None = None,
+        is_projectable: bool | None = None,
     ) -> None:
         self.name = name
         self.inputs = inputs
@@ -161,6 +168,7 @@ class AwkwardInputLayer(AwkwardBlockwiseLayer):
         self.produces_tasks = produces_tasks
         self.annotations = annotations
         self.creation_info = creation_info
+        self._is_projectable = is_projectable
 
         io_arg_map = BlockwiseDepDict(
             mapping=LazyInputsDict(self.inputs),  # type: ignore
@@ -188,9 +196,18 @@ class AwkwardInputLayer(AwkwardBlockwiseLayer):
     @property
     def is_projectable(self) -> bool:
         # isinstance(self.io_func, ImplementsProjection)
-        return (
-            io_func_implements_projection(self.io_func) and not self.has_been_unpickled
-        )
+        if self._is_projectable is None:
+            return (
+                io_func_implements_projection(self.io_func)
+                and not self.has_been_unpickled
+            )
+        assert isinstance(self._is_projectable, bool)
+        return self._is_projectable
+
+    @is_projectable.setter
+    def is_projectable(self, value: bool) -> None:
+        assert isinstance(value, bool)
+        self._is_projectable = value
 
     @property
     def is_mockable(self) -> bool:
@@ -297,6 +314,20 @@ class AwkwardInputLayer(AwkwardBlockwiseLayer):
         assert self.is_columnar
         return cast(ImplementsNecessaryColumns, self.io_func).necessary_columns(
             report=report, state=state
+        )
+
+    def project_manually(self, columns: frozenset[str]) -> AwkwardInputLayer:
+        """Project the necessary _columns_ to the AwkwardInputLayer."""
+        assert self.is_projectable
+        io_func = cast(ImplementsProjection, self.io_func).project_manually(columns)
+        return AwkwardInputLayer(
+            name=self.name,
+            inputs=self.inputs,
+            io_func=io_func,
+            label=self.label,
+            produces_tasks=self.produces_tasks,
+            creation_info=self.creation_info,
+            annotations=self.annotations,
         )
 
 
--- a/src/dask_awkward/layers/layers.py
+++ b/src/dask_awkward/layers/layers.py
@@ -201,12 +201,10 @@ class AwkwardInputLayer(AwkwardBlockwiseLayer):
                 io_func_implements_projection(self.io_func)
                 and not self.has_been_unpickled
             )
-        assert isinstance(self._is_projectable, bool)
         return self._is_projectable
 
     @is_projectable.setter
     def is_projectable(self, value: bool) -> None:
-        assert isinstance(value, bool)
         self._is_projectable = value
 
     @property
--- a/src/dask_awkward/lib/io/columnar.py
+++ b/src/dask_awkward/lib/io/columnar.py
@@ -44,6 +44,8 @@ class ImplementsColumnProjectionMixin(ImplementsNecessaryColumns, Protocol):
 
     def project_columns(self: T, columns: frozenset[str]) -> T: ...
 
+    def project_manually(self: T, columns: frozenset[str]) -> ImplementsIOFunction: ...
+
     def __call__(self, *args, **kwargs): ...
 
 
@@ -176,3 +178,6 @@ class ColumnProjectionMixin(ImplementsNecessaryColumns[FormStructure]):
             return self
 
         return self.project_columns(self.necessary_columns(report, state))
+
+    def project_manually(self: S, columns: frozenset[str]) -> ImplementsIOFunction:
+        return self.project_columns(columns)
--- a/src/dask_awkward/manual/column_optimization.py
+++ b/src/dask_awkward/manual/column_optimization.py
@@ -1,0 +1,58 @@
+from __future__ import annotations
+
+from typing import cast
+
+from dask.highlevelgraph import HighLevelGraph
+
+from dask_awkward.layers.layers import AwkwardInputLayer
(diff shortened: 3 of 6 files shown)


pfackeldey:
The Conda tests seems to fail unrelated to this PR right now.

martindurant:
> The Conda tests seems to fail unrelated to this PR right now.

mambaforge is deprecated, it should be reverted back to miniconda in the GHA yaml definition

src/dask_awkward/layers/layers.py
```diff
@@ -188,9 +196,18 @@ def __repr__(self) -> str:
...
         # isinstance(self.io_func, ImplementsProjection)
-        return (
-            io_func_implements_projection(self.io_func) and not self.has_been_unpickled
-        )
+        if self._is_projectable is None:
+            return (
+                io_func_implements_projection(self.io_func)
+                and not self.has_been_unpickled
+            )
+        assert isinstance(self._is_projectable, bool)
```
martindurant, line 204: These ind of type-checks should really not be necessary. In fact, assertions should not normally appear in non-test code at all.

src/dask_awkward/layers/layers.py
```diff
@@ -299,6 +316,20 @@ def necessary_columns(self, report: TypeTracerReport, state: T) -> frozenset[str
             report=report, state=state
         )
 
+    def project_manually(self, columns: frozenset[str]) -> AwkwardInputLayer:
+        """Project the necessary _columns_ to the AwkwardInputLayer."""
+        assert self.is_projectable
+        io_func = cast(ImplementsProjection, self.io_func).project_manually(columns)
```
martindurant, line 322: Personally, I am not a fan of all the cruft is takes to make mypy happy - I'd rather do with simpler or missing types.

src/dask_awkward/lib/io/columnar.py
```diff
@@ -44,6 +44,8 @@ def attrs(self) -> dict | None: ...
 
     def project_columns(self: T, columns: frozenset[str]) -> T: ...
 
+    def project_manually(self: T, columns: frozenset[str]) -> ImplementsIOFunction: ...
```
martindurant, line 47: This is fully specified in the implementation of ColumnProjectionMixin, right? So does the protocol definition here serve any purpose?

src/dask_awkward/manual/column_optimization.py
```diff
@@ -0,0 +1,58 @@
...
+
+from typing import cast
+
+from dask.highlevelgraph import HighLevelGraph
+
+from dask_awkward.layers.layers import AwkwardInputLayer
+from dask_awkward.lib.core import Array
+
+
+def optimize_columns(array: Array, columns: dict[str, frozenset[str]]) -> Array:
```
martindurant, line 11: We need to think of way to ensure that the resulting output is not optimised again at compute time.

tests/test_manual.py
```diff
@@ -0,0 +1,53 @@
...
+    key, cols = needs.popitem()
+    cols |= {"u8"}
+
+    needs = {key: cols}
+
+    u4_and_u8_array = dak.manual.optimize_columns(array, needs)
+
+    assert u4_and_u8_array.fields == ["u4", "u8"]
+
+    materialized_u4_and_u8_array = u4_and_u8_array.compute()
```
martindurant, line 45: With or without optimise?

martindurant (commented):
On a first quick scan, it is indeed going to be sueful and not too complex to extract the column selection/optimisation as you have envisaged.

I only have a couple of comments so far, but one bigger question: if you are familiar with the one-pass PR, is it obvious how to port what is here to that branch? It should not be hard, since necessary_columns still exists, and you're not actually calling any of the opt machinery here, but I think it's worth checking - in the long run, I still think that one-pass is what we should aim for.

src/dask_awkward/layers/layers.py
pfackeldey: yes :+1: I'll remove them

src/dask_awkward/layers/layers.py
pfackeldey: I was following the same implementations as we have here for `project` and `necessary_columns`.

src/dask_awkward/lib/io/columnar.py
pfackeldey: I may be able to remove it without mypy complaining, I'll check. Apart from that, I was adding it here as other ones were defined for this protocol aswell, e.g. `project_columns`.

src/dask_awkward/manual/column_optimization.py
pfackeldey: That's why I set `.is_projectable` on the layers that are optimized "by hand" to `False`. However, I'm not sure if that's the most elegant solution.

tests/test_manual.py
pfackeldey: This would result in the same output, because currently `dak.manual.optimize_columns` sets the `is_projectable` property of AwkwardInputLayers to `False`, which disables any other column optimization afterwards.

src/dask_awkward/manual/column_optimization.py
martindurant: Hah, OK, I can see that works. We should see what it does with multiple input layers.

pfackeldey:
> On a first quick scan, it is indeed going to be sueful and not too complex to extract the column selection/optimisation as you have envisaged.
> 
> I only have a couple of comments so far, but one bigger question: if you are familiar with the one-pass PR, is it obvious how to port what is here to that branch? It should not be hard, since necessary_columns still exists, and you're not actually calling any of the opt machinery here, but I think it's worth checking - in the long run, I still think that one-pass is what we should aim for.

I think so aswell: It should be straight forward to port this PR to the one-pass PR because it just uses the underlying optimization implementation and doesn't care how it's implemented. I haven't tried it yet though. Given that it's conceptually independent, I went forward to open the PR already now.

src/dask_awkward/manual/column_optimization.py
pfackeldey: :+1: It should work as follows when calling `.compute(optimize_graph=True)`: any layer that has been manually projected won't be optimized again (they are basically treated as a any other non-column-optimizable layer), whereas any other AwkwardInputLayer goes through the usual automatic optimization process.

src/dask_awkward/layers/layers.py
pfackeldey: I tried to remove it, but unfortunately it is not that easy to appease mypy otherwise (except for putting `# type: ignore` at multiple places - but what's the point of type checking then?). 

The reason for the current implementation is that we support a protocol that all projectable IO layers need to adhere to (i.e. `ImplementsProjection`). By making sure that `self.is_projectable` is true, we know that our IO layer has the methods of the `ImplementsProjection` protocol available. Unfortunately, mypy seems to not recognize this correctly, because `self.io_func` may be a more general implementation that only needs to implement the `ImplementsIOFunction` protocol. `ImplementsProjection` is a specification of that, and we can safely "cast" it here after making sure that `self.is_projectable` is True. 

It's not something that I invented/added new here. I'm following the existing protocols for the IO functions.

(The same argument goes for the comment above)

martindurant: > It's not something that I invented/added new here. I'm following the existing protocols for the IO functions.


Agreed. It is mostly removed in the one-pass PR, and I would say nothing useful got lost :)

martindurant:
What would you like to do here? As far as I can tell, everything is ready (except some lint), ans we can go ahead. It would be nice to add some example(s) to the docs too, but that can follow.

pfackeldey:
Yes, sorry, it is done from my side. I'll fix the lint soon. I was just caught up with other work...

PR #569, opened 2025-01-31
chore: drop python 3.8

pfackeldey:
I hope I haven't forgotten any places?

### DIFF
--- a/.github/workflows/conda-tests.yml
+++ b/.github/workflows/conda-tests.yml
@@ -28,7 +28,7 @@ jobs:
     - name: Setup Conda Environment
       uses: conda-incubator/setup-miniconda@v3
       with:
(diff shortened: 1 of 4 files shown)


martindurant:
Should add 3.13 at the same time

martindurant:
(need the same at dask-histogram too, but all CI is red right now)

lgray:
conda tests will fail until dask histogram is updated in conda.

lgray:
ok - looks like we're good to go. Need to check something over in uproot though.

lgray:
Nope - right - uproot is OK because the tree reduction for the reports is here. Very good.

pfackeldey:
any objections about merging this PR @martindurant and @lgray ?

Issue #577, opened 2025-04-11
Cannot use `dak.num(events, axis=0)` after having computed a slice of the `events` when `open_files=False` during reading


ikrommyd:
To reproduce
```py
import uproot
import dask_awkward as dak

events = uproot.dask({"../coffea/tests/samples/nano_dy.root": "Events"}, open_files=False)
events[:10].compute()
dak.num(events, axis=0)
```
will raise
```
TypeError: unsupported dtype: dtype('O'). Must be one of

    bool, int8, uint8, int16, uint16, int32, uint32, int64, uint64, float32, float64, complex64, complex128, datetime64, timedelta64, float16

or a datetime64/timedelta64 with units (e.g. 'datetime64[15us]')

This error occurred while calling

    ak.to_layout(
        TypeTracerArray-instance
        allow_record = False
        regulararray = False
        primitive_policy = 'error'
    )
```
Similarly in coffea:
```py
from coffea.nanoevents import NanoEventsFactory
import dask_awkward as dak

events = NanoEventsFactory.from_root({"../coffea/tests/samples/nano_dy.root": "Events"}, mode="dask").events()
events[:10].compute()
dak.num(events, axis=0)
```

PR #578, opened 2025-04-11
fix: dtype inference in new_known_scalar

pfackeldey:
closes https://github.com/dask-contrib/dask-awkward/issues/577

if a 0-dim array is passed to `new_known_scalar` the dtype was inferred to be `np.dtype(type(array(...)))` which results in object dtype. This PR directly uses the dtype of the input arg if it has a dtype attribute, which is true for array types.

### DIFF
--- a/src/dask_awkward/lib/core.py
+++ b/src/dask_awkward/lib/core.py
@@ -697,6 +697,8 @@ def new_known_scalar(
             dtype = np.dtype(int)
         elif isinstance(s, (float, np.floating)):
             dtype = np.dtype(float)
+        elif hasattr(s, "dtype"):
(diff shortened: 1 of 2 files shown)


pfackeldey:
Needs https://github.com/dask-contrib/dask-awkward/pull/579 to fix the awkward-main ci test

pfackeldey:
ugh... there seem to be the expectation of dask-awkward that a 0-dim typetracer array is _always_ promoted to a length-1 layout. This is something that does not happen with every other array type. I explicitly fixed this in awkward recently: https://github.com/scikit-hep/awkward/pull/3469#issuecomment-2802584138, but dask-awkward seems to rely on it...

PR #579, opened 2025-04-11
fix: remove `index_nplike`

pfackeldey:
Awkward does not support `index_nplike` anymore with https://github.com/scikit-hep/awkward/pull/3446 (we need to wait for a new release here). This PR replaces `index_nplike` with `nplike`.


martindurant:
Isn't it the case that index_nplike went away, but nplike was already there - so it doesn't need the latest awkward at all? Or maybe I misunderstood.

pfackeldey:
Oh you're completely right! Somehow I haven't thought about this being backwards compatible, but it is 😅

martindurant:
OK, now let's do #578

PR #580, opened 2025-04-17
fix: explicit primitive promotion for scalar typetracer arrays

pfackeldey:
This fixes: https://github.com/dask-contrib/dask-awkward/pull/578#issuecomment-2813033447


pfackeldey:
This is just restoring the original behavior, there shouldn't be any performance drawbacks. `ak.Array` would internally use `ak.to_layout` to promote a 0-dim typetracer array before. Now we do that explicitly and `ak.Array` skips this step then.

martindurant:
OK +1

PR #582, opened 2025-04-24
fix: avoid dask v2025.4.0 for CI tests

pfackeldey:
Our CI is broken since: https://github.com/dask/dask/pull/11859#issuecomment-2828709361

Need to get a corresponding fix in dask.

Even though the actual code seems to run judging from playing around, I'd be in favor of avoiding this dask release as we can't be entirely sure given that our tests won't run...

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -37,7 +37,7 @@ classifiers = [
 ]
 dependencies = [
   "awkward >=2.5.1",
-  "dask >=2023.04.0,!=2025.4.0",
+  "dask >=2023.04.0,<2025.4.0",
   "cachetools",
   "typing_extensions >=4.8.0",
 ]
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -37,7 +37,7 @@ classifiers = [
 ]
 dependencies = [
   "awkward >=2.5.1",
-  "dask >=2023.04.0",
+  "dask >=2023.04.0,!=2025.4.0",
   "cachetools",
   "typing_extensions >=4.8.0",
 ]


pyproject.toml
```diff
@@ -37,7 +37,7 @@ classifiers = [
 ]
 dependencies = [
   "awkward >=2.5.1",
-  "dask >=2023.04.0",
+  "dask >=2023.04.0,!=2025.4.0",
```
martindurant, line 40: Do we want a < in case another release comes out meanwhile?

pfackeldey: I can do that :+1:

martindurant:
We really need to `flaky` on that distance test or increase the tolerance

GaetanLepage:
Do you have an estimate about when `dask>=2025.4.0` compatibility will be fixed? This is blocking us to update `dask` and `awkward` in [nixpkgs](https://github.com/NixOS/nixpkgs/pull/429815).

GaetanLepage:
> @GaetanLepage : I'm afraid we really don't. You probably haven't been following the drama in dask, but things are very broken from our point of view, and we are actually recommending people use the 2024.12 versions of dask/distributed.
> 
> There is still momentum and good will behind dask, so hopefully we will be able to resolve this.

No, I'm not aware of the drama. I only have a "package maintainer" relationship with this ecosystem.
The nix python package set basically enforces that no more than one version of each package is present in the repository at the same time.
Currently, we have `dask==2025.7.0`, `dask-awkward==2025.5.0` (broken as incompatible with `dask>=2025.5.0`) and `awkward==2.8.5`.
Ideally, we would like to bump all these packages to their most recent version.
I'm not sure which combination you would recommend. Holding all the versions back prevents us from updating other packages and is not idea.

Thanks in advance for your opinion on the matter.

GaetanLepage:
> I don't believe anyone is working on this right now, and dask itself is in "maintenance" mode and unlikely to offer effort to help.

Ok good to know. Thanks for the clatrification.

PR #586, opened 2025-04-28
fix: forgot this promotion in #580

pfackeldey:


martindurant:
Yep, jusyt like all the others, +1

PR #591, opened 2025-08-18
fix: Add including_unknown kwarg to values_astype

rpsimeon34:
Update the available keywords in `dak.values_astype`'s signature to match those of `ak.values_astype`


rpsimeon34:
Hi Martin, this comes up for me in a `coffea` + ML inference workflow, where my ML tool (`pytorch` wrapped in a `coffea` template class) expects an array with known, numpy-like types. An `sklearn` tool further upstream converts my array to have `Unknown` types, so I needed to use `including_unknown` to make things compatible.

PR #592, opened 2025-08-23
fix: remove known scalar optimization from `dak.num`

ikrommyd:
This optimization is causing a headache to some users. The reason is that the users typically use `dak.num` on axis=0 to keep track of the number of events they processed. The problem is that if you know the divisions and yet you fail to read a partition due to some file access error (pretty common), you will have failed to read a chunk of events yet `dak.num` is going to report a fixed number `array.defined_divisions[-1]`. This causes users to over estimate the number of events they process.

I claim that this optimization is causing more problems than it solves, therefore I propose to disable it.
If a user wants the number without doing any computation, they can just use `array.defined_divisions[-1]`.
We also can't add an argument to `dak.num` because the whole point is that the function should have the same. arguments as `ak.num` (+ automatic dispatching).

### DIFF
--- a/src/dask_awkward/lib/structure.py
+++ b/src/dask_awkward/lib/structure.py
@@ -20,7 +20,6 @@ from dask_awkward.lib.core import (
     PartitionCompatibility,
     _map_partitions,
     map_partitions,
-    new_known_scalar,
     new_scalar_object,
     partition_compatibility,
 )


martindurant:
This is a heavy change! But it's not just `num` that wrong, but the partitions in general - is this information not used anywhere, not even by end-user introspection?

In chat, we talked about explicitly removing the knowledge about the partitions in the case that a uses suspects failures and is using the reporting mechanism.

ikrommyd:
For the "would be length", we also have `__len__`: https://github.com/dask-contrib/dask-awkward/blob/f006fb9a2be0833b9c44b027dc7c5a9b763be736/src/dask_awkward/lib/core.py#L1079-L1087

martindurant:
It seems to me still, that it would make more logical sense to drop knowledge of the partition edges along with the request for a failure report. What we are saying, is that we expect some partitions to be missing, so the partition edge values are wrong in that case - and this will result in num not having the shortcut available. In this case, num retains the same semantics as other ak operations (as opposed to len, which must evaluate, and could be costly in some cases).

ikrommyd:
But you can compute the divisions with `eager_compute_divisions` right? My point was that given the bad file access in HEP, we cannot trust a "previously computed length" as the actual length when we want it because a file become unavailable between the two calls. I will not fight this because I think the uproot fix covers us and I don't think users will use `eager_compute_divisions`, I just think this known scalar optimizations (even though faster) can potentially cause more harm than what it actually solves (performance)

ikrommyd:
> Ah so we could also teach our HEP users to manually do `eager_compute_divisions` @ikrommyd ?

when partitions are unknown, you can do `array.eager_compute_divisions()` which does involve opening up the files and reading shapes (I think).

My point was that even if you do `array.eager_compute_divisions()`, and read all the files now, maybe 1 file errors 10 seconds later when you actually do the actual `.compute()` but you have trusted the `dak.num` result that came from the `eager_compute_divisions` call. Anyways, I hope users don't do this.

ikrommyd:
Closing this then :)

PR #595, opened 2025-08-27
feat: handle virtual array in pickling

ikrommyd:
Dask-awkward registers a global pickler which will be used if someone tries to pickle an `ak.Array` if dask-awkward just exists in the environment (not even imported). Therefore it should handle virtual buffers too by simply materializing them. Users may run into `TypeError: a bytes-like object is required, not 'VirtualNDArray'` otherwise coming from `PickleBuffer`.

This was the easiest way for me to not break backwards compatibility with earlier versions of awkward and also not doing try except statements with `ImportError` to check if awkward is new-enough to implement virtual arrays. Let me know if you think there is a nicer way.

### DIFF
--- a/src/dask_awkward/pickle.py
+++ b/src/dask_awkward/pickle.py
@@ -15,6 +15,8 @@ def _maybe_make_pickle_buffer(buffer: Any) -> PlaceholderArray | PickleBuffer:
     if isinstance(buffer, PlaceholderArray):
         return buffer
     else:
+        if hasattr(buffer, "materialize") and callable(buffer.materialize):
+            buffer = buffer.materialize()
         return PickleBuffer(buffer)
 
 


ikrommyd:
I'm not sure how one would pickle a virtual array. It contains lambda functions (its generators) that have access to an open root file. My plan was to enforce materialization on pickling (i.e you can only pickle the full materialized array which contains `np.ndarray` buffers).

martindurant:
> It contains lambda functions (its generators) that have access to an open root file. Is it possible to pickle that? 

You can do anything with some effort :). In this case, you would have to come up with a description of the function and possible columns, and reopen the file when needed (which would only happen if you actually need to load more).

> this is only meant to happen if someone wants to to return some_arrays in a coffea processor

but we pickle ak arrays all the time in dak, and now they will all be virtual by default, no? If you have a first node: load-parquet-file, and second node select-one-column, the column optimisation will result in a virtual array in the first node with one real column and a bunch of vitual ones. Should the two nodes happen on different workers, you would materialise everything.

ikrommyd:
No no....dask-awkward is not affected by this. It's the dask-awkward affects pickling if it's just installed in the environment. Nothing is virtual unless people want to and that does not happen in dask-awkward workflows. See here:
https://github.com/scikit-hep/awkward/blob/fd5fa80d94ed47b3030e0d44a805f9e19fd5c7d8/src/awkward/highlevel.py#L1668-L1671
It's just that dak registers an entry point for its pickler that will be used in awkward if you just have dak in the same env, that's all. Therefore I claim that it needs to handle ALL of awkward since it enforces itself.

martindurant:
> its pickler that will be used in awkward if you just have dak in the same env, that's all. Therefore I claim that it needs to handle ALL of awkward since it enforces itself.

Yes, totally agree with this.

But this *is* the pickler that dak itself uses, no? Can you run distributed.protocol.serialize.dask_dumps on a virtual array (with your change) versus the materialised version? The former should be considerably smaller, and in fact I would say this is a good test we should include.

martindurant:
Yes yes - we want both to be able to use pickling via both ak's dispatch and distributed's. I'm still not totally convinced that even your users would never want to pickle the virtual form, though.

pfackeldey:
I think we should make it possible to pickle virtual arrays without materializing. Then we could get rid of placeholder arrays in `uproot.dask` and replace them with virtual arrays. That's how dask-awkward workflows would encounter virtual arrays. The benefit of replacing placeholder arrays is that if we under-touched or had a tracing issue, we could still run the computation and just load the columns as needed. 

Apart from that I don't see how or where virtual arrays would ever enter in a dask-awkward workflow.

Isn't dask(-distributed?) using `cloudpickle` anyways, which does support pickling lambdas?

Maybe the correct way is that in awkward's highlevel pickle protocol we do a `if layout.is_not_all_materialized: raise ...` with a reasonable error that says "please call ak.materialize(...)" before using plain pickle. But when using cloud pickle we just serialize the python object without materialization.

I'm not a pickle expert, so I'm not sure if this would work what I'm proposing. Do we need to add pickle protocols to the `VirtualNDArray`?

In any case, this all requires `uproot.dask` to replace placeholder arrays first with virtual arrays, otherwise I don't see how virtual arrays ever end up in dask-awkward workflows in the first place. Until then, I think it's good if dask-awkward's pickling plugin deals with all arrays types _somehow_, maybe for now just materialize them?

PS: what happens if we treat virtual arrays here as placeholder arrays, i.e. just:
```python
if isinstance(buffer, VirtualNDArray):
  return buffer
```
and let cloudpickle do the job then?

ikrommyd:
We could also just close this and attempt pickling virtual arrays now.

pfackeldey:
> My take is that if we don't wanna deal with this right now, we could put this PR in so that coffea users are not affected by dask-awkward existing and we can improve pickling a bit later.

I think I'd agree with that. As I said: the proper pickling is only needed after changing uproot.dask (which hasn't happen yet).

martindurant:
If we can guarantee that no virtual array will be the result of a dask task, then we are fine. But we'll have to come back to it pretty soon! The plan was to have virtual arrays be THE solution to under-touching in column optimisation, right?

martindurant:
OK, then.
@pfackeldey , happy?

pfackeldey:
I'd prefer an error with a message that this (virtual arrays in dask-awkward) is in development. Better error than an unexpected memory explosion I'd argue. I'm pretty sure no-one will hit this for now anyway.

So, :+1: from my side as well!

We should follow up on uproot with this soon, if we can get `uproot.dask` to replace PlaceholderArrays with VirtualNDArrays we're halfway there!

ikrommyd:
Well there is this outrageous `dak.from_awkward` in coffea: https://github.com/scikit-hep/coffea/blob/ab56206112a6e89eb748cc7948951395dfcc9589/src/coffea/jetmet_tools/CorrectedJetsFactory.py#L228-L232 which I'm currently removing.

ikrommyd:
> OK, so let's make from_awkward error or materialise with a warning for virtuals

Should we do this here or in a separate PR?

pfackeldey:
@ikrommyd and @martindurant turns out I have a working solution already for VirtualNDArrays in uproot.dask. I think we should then aim for pickling VirtualNDArrays properly. Do you have tips on how to do that correctly @martindurant ?

martindurant:
You typically want to rerun the same function that created it in the first place

martindurant:
OK go for it.
but it's not over! :)

PR #629, opened 2026-07-21
fix: make dask_property picklable

ikrommyd:
:robot: _AI text below_ :robot:

A class using `@dask_property` cannot be pickled by value:

```python
import cloudpickle
import dask_awkward as dak

class Thing:
    @dak.dask_property
    def foo(self):
        return 1

    @foo.dask
    def foo(self, array):
        return 2

cloudpickle.dumps(Thing)
# TypeError: cannot pickle '_DaskProperty' object
```

`property` keeps `fget`/`fset`/`fdel` in C-level slots and provides no reduction of its own, so CPython refuses outright — `pickle.dumps(property(f))` raises the same error. `_DaskProperty` inherits that refusal even though it does have an instance `__dict__`.

This matters because cloudpickle only pickles a class by reference when it is reachable as `getattr(module, cls.__qualname__)`. Anything else — a behavior mixin written in `__main__` or in a notebook cell, or one whose generated `*Record`/`*Array` classes have been shadowed in their module — gets pickled by value, which walks the class dict and hits the property. The result is that such behaviors cannot be sent to a distributed worker at all.

`_DaskProperty` now defines `__reduce__`, round-tripping `fget`/`fset`/`fdel`/`__doc__`/`_dask_get`.

`_DaskMethod` needs no change: it is a plain Python class whose whole state lives in the instance `__dict__`, so the default reduction already handles it and cloudpickle serializes the contained function and closure by value. It gets a test to keep it that way.

Found while debugging the coffea side of this, where the same copy of `_DaskProperty` broke `FCCSchema` under a distributed client: scikit-hep/coffea#1603.

### DIFF
--- a/src/dask_awkward/lib/core.py
+++ b/src/dask_awkward/lib/core.py
@@ -131,6 +131,22 @@ F = TypeVar("F", bound=Callable)
 G = TypeVar("G", bound=Callable)
 
 
+def _rebuild_dask_property(
+    fget: Callable | None,
+    fset: Callable | None,
+    fdel: Callable | None,
+    doc: str | None,
+    dask_get: Callable | None,
+) -> _DaskProperty:
+    """Reconstruct a `_DaskProperty` when unpickling."""
+    prop = _DaskProperty(fget, fset, fdel)
+    # a doc passed to property() lands in a C slot which, before 3.13, the
+    # subclass' own docstring shadows -- so assign it to the instance instead
+    prop.__doc__ = doc
+    prop._dask_get = dask_get
+    return prop
+
+
 class _DaskProperty(property):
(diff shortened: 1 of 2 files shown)


martindurant:
OK. Looks fine. Wonder how you came onto this? Wonder whether the classes that have the property should be changed to have their own reduce() - but there's no downside to this either.

ikrommyd:
> OK. Looks fine. Wonder how you came onto this? Wonder whether the classes that have the property should be changed to have their own reduce() - but there's no downside to this either.

Claude found it while working on https://github.com/scikit-hep/coffea/pull/1603. In coffea to make dask-awkward optional dependency we just vendored dask method and dask property. So I thought I'd just upstream the patch here too.

PR #632, opened 2026-08-07
chore: use pre v0.16 ruff rules and move isort and pyupgrade to ruff

ikrommyd:
ruff 0.16 introduced a million new rules by default making pre-commit fail very badly right now so we use the pre v0.16 rules as specified in https://astral.sh/blog/ruff-v0.16.0 + we also move isort and pyupgrade to ruff.


martindurant:
Well that was annoying!

Btw: logging statements should *not* use f-strings, so that formatting can be avoided if the logger is not active...

